PR #18, opened 2015-03-30
Taxonomy cleanup

phobologic:


phobologic:
Tested that this doesn't result in any changes to the stacks, this is all just internal changes.  The taxonomy may change again, but this is super helpful already.

Issue #15, opened 2015-03-18
If a stack already exists, and a parameter isn't provided on an update, pull from existing parameters


phobologic:
CloudFormation's webui already does this for you, and it's convenient and helps prevent accidentally changing a parameter that you don't want to update.

Maybe make it so you have to provide an additional flag to actually change a parameter on an update?  That may be a bit extreme and annoying.

Issue #11, opened 2015-03-12
Check all required Parameters


phobologic:
Stacker should be able to verify that all necessary Parameters are provided before attempting to launch any stacks.

PR #19, opened 2015-03-31
Use existing params

phobologic:
When updating a stack, if a parameter is not given but is set on the existing stack, use the existing stack's parameter.  Also, this pull allows us to catch missing parameters earlier.

Fixes GH-11
Fixes GH-15


phobologic:
Hrm - this is a diff from the taxonomy_cleanup branch, which is on deck.  I'm going to do some more testing of that branch, then I'll reopen this once I merge taxonomy_cleanup.

PR #20, opened 2015-03-31
Use existing params

phobologic:
Ok, now that the taxonomy pull has been merged, this is cleaner.

This pulls existing parameters from existing stacks if a given parameter isn't provided by the user on the cli or config.

It also checks for missing, required (without default) parameters earlier.


phobologic:
Fixes GH-11
Fixes GH-15

PR #25, opened 2015-04-23
Remove automatic subnet building

phobologic:
Before this stacker would automatically determine availability zones and
both public & private subnets, based on subnet math. It put too much of
the burden on python. Now, instead, the AZs are determined
automatically, but in CloudFormation (Fn::GetAZs) and you only provide a
count of AZs.  As well, you need to provide a set of Public & Private
subnets to use.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 ## pending
 
+- remove auto-subnet splitting in vpc stack
+
 ## 0.2.2 (2015-03-31)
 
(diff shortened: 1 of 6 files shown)


ejholmes:
Might be nice to add some approval based tests to this that generate the cloudformation JSON just for sanity checking. Otherwise, lgtm :+1:

phobologic:
I'm going to do a pull request with a lot of testing added after merging this one.  Still working out a bit of it.

PR #26, opened 2015-05-05
Remove dns requirement, add hook/plugin support

phobologic:
- build hooks
- remove domain as a core concept, replace with namespace
- bucket used for CF templates prepended with 'stacker-'
- asg blueprint optionally sets up DNS if BaseDomain provided
- route53 hook - requires BaseDomain parameter

### DIFF
--- a/stacker/builder.py
+++ b/stacker/builder.py
@@ -7,8 +7,7 @@ from aws_helper.connection import ConnectionManager
 
 from boto.exception import S3ResponseError, BotoServerError
 
-from .util import (create_route53_zone, load_object_from_string,
-                   get_bucket_location)
+from .util import load_object_from_string, get_bucket_location
 
 from .plan import (Plan, INPROGRESS_STATUSES, STATUS_SUBMITTED,
                    COMPLETE_STATUSES)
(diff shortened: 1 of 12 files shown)


stacker/builder.py
```diff
@@ -25,17 +24,17 @@ def __str__(self):
...
 
 def stack_template_key_name(blueprint):
     return "%s-%s.json" % (blueprint.name, blueprint.version)
 
 
-def stack_template_url(cfn_domain, blueprint):
+def stack_template_url(bucket_name, blueprint):
     key_name = stack_template_key_name(blueprint)
-    return "https://s3.amazonaws.com/%s/%s" % (cfn_domain, key_name)
+    return "https://s3.amazonaws.com/%s/%s" % (bucket_name, key_name)
```
ejholmes: I never noticed this before. Does the cloudformation api require that you upload the JSON to an S3 bucket first?

ejholmes:
Cool. Lgtm :+1:

stacker/builder.py
phobologic: You don't have to, you can pass the body directly, but if you do you are limited to 51k.  If you use a templateURL you are allowed to have a policy as large as 460k.  In general it's preferred to store them in S3.  We version the templates, so another benefit is that you can see every version you've ever uploaded of a stack in that bucket.

http://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_CreateStack.html (See TemplateBody/TemplateURL)

PR #27, opened 2015-05-11
Environment feature

phobologic:
These are simple key/value yaml files that can be used as a map against the config as if the config were a python string.Template template.

This allows for substituting values based on the environment, which seems like it might be a pretty common use case.


ejholmes:
:+1:

PR #29, opened 2015-05-11
Add optional Internal Zone setup to VPC

phobologic:
Requires that a new release of troposphere be released for this checkin:

https://github.com/cloudtools/troposphere/commit/cdd1b271030d26312c51fdc53ab04bc3aa6fc8b7


ejholmes:
:+1:

PR #30, opened 2015-05-13
Postgres internal alias

phobologic:
If you provide InternalZoneId, InternalZoneName, and InternalHostname parameters a CNAME will be created for the database.


ejholmes:
:+1:

Issue #22, opened 2015-03-31
Allow specifying stack name for parameters on the command line


phobologic:
Make it so --parameter <stack_name>::<parameter_name> applies that parameter only to that stack.  Right now they are global.

PR #32, opened 2015-05-19
Stack specific parameters on CLI

phobologic:
Fixes GH-22. Also adds tests to verify parameter precidence is what we
say it is.


ejholmes:
:+1:

PR #34, opened 2015-05-20
Random fixes and cleanup

mhahn:


stacker/blueprints/vpc.py
```diff
@@ -109,11 +119,13 @@ def create_dhcp_options(self):
         dhcp_options = t.add_resource(ec2.DHCPOptions(
             'DHCPOptions',
             DomainName=domain_name,
-            DomainNameServers=['AmazonProvidedDNS', ]))
+            DomainNameServers=['AmazonProvidedDNS', ],
+            Condition="CreateInternalDomain"))
```
phobologic: We need to remove this condition - no matter what, we need the DHCPOptions.  The domain name will just hold the external zone if no internal zone is provided already, since the default for InternalZone is "".

stacker/blueprints/vpc.py
```diff
@@ -109,11 +119,13 @@ def create_dhcp_options(self):
...
         t.add_resource(ec2.VPCDHCPOptionsAssociation(
             'DHCPAssociation',
             VpcId=VPC_ID,
-            DhcpOptionsId=Ref(dhcp_options)))
+            DhcpOptionsId=Ref(dhcp_options),
+            Condition="CreateInternalDomain"))
```
phobologic: Same here, we have to remove this Condition.

stacker/blueprints/vpc.py
mhahn: i see, this was failing for me when i was running it.

stacker/config.py
```diff
@@ -1,8 +1,7 @@
+import yaml
```
phobologic: So I tend to break up imports in three sections:

- python native modules
- third party modules
- project modules

That's why this was separate.  I think there's some benefit to doing that - even if you alphabetize inside them.

stacker/tests/test_builder.py
```diff
@@ -84,3 +90,21 @@ def test_stack_params_dont_override_given_params(self):
...
         result = handle_missing_parameters(def_params, required, stack)
         self.assertEqual(result, def_params.items())
+
+
+class TestBuilder(unittest.TestCase):
+
+    def setUp(self):
+        self.builder = Builder('us-east-1', 'namespace')
+
+    def test_resolve_parameters_referencing_non_existant_output(self):
```
phobologic: Is there maybe any refactoring we could do to make testing resolve_parameters better?  I had planned to dive back into that sometime soon - this is fine for now, but if you want to take a stab at that that'd be awesome.

phobologic:
Looks good!  Thanks for this!

stacker/tests/test_builder.py
mhahn: As a part of adding tear down i'll take a look

stacker/blueprints/vpc.py
```diff
@@ -65,7 +65,10 @@ class VPC(Blueprint):
 
     def create_conditions(self):
         self.template.add_condition(
-            "CreateInternalDomain",
+            "NoHostedZones",
+            Equals(Ref("InternalDomain"), ""))
+        self.template.add_condition(
+            "HasHostedZones",
```
phobologic: This only tests for InternalDomain - should also test for BaseDomain I think.

stacker/builder.py
```diff
@@ -142,15 +146,20 @@ def conn(self):
         return self._conn
 
     @property
+    def s3_conn(self):
+        if not getattr(self, '_s3_conn', None):
```
phobologic: Should probably set self._s3_conn to None in the init

phobologic:
:+1: Looks good! Let me know about the latest comments, then I'll merge.  Thanks!

stacker/builder.py
mhahn: switched this to "hasattr", the point is that this is a private variable and should only be accessed by the property to cache the connection.

stacker/blueprints/vpc.py
mhahn: done

mhahn:
Ok adjusted based on those two comments

phobologic:
Awesome!  Thanks!

PR #35, opened 2015-06-08
Support for subcommands within stacker

mhahn:
This small PR does the following:

- adds support for subcommands
- adds support for stacker "actions"
- moves original stacker script logic to "build" action
- adds "destroy" action
- adds "destroy" subcommand
- supports printing a basic outline of what stacker is going to for "destroy" and "build"

I still have some #TODO sections to call out docstrings that need to be verified/updated/added.


stacker/actions/build.py
```diff
@@ -0,0 +1,201 @@
+import logging
+
+from . import base
```
phobologic: nit: Since we only use one thing out of base, might as well:

`from .base import BaseAction`

stacker/actions/build.py
```diff
@@ -0,0 +1,201 @@
...
+from .. import exceptions, util
+from ..plan import Plan
+
+logger = logging.getLogger(__name__)
+
+
+class Action(base.BaseAction):
```
phobologic: Do you think it'd be worth naming the Actions a bit more descriptively?  ie: BuildAction, DestroyAction?

stacker/actions/destroy.py
```diff
@@ -0,0 +1,64 @@
...
+    def _get_dependencies(self, stacks_dict):
+        dependencies = {}
+        for stack_name, stack in stacks_dict.iteritems():
+            provider_stack = self.provider.get_stack(stack.fqn)
+            required_stacks = []
+            if provider_stack:
+                required_stacks = self.provider.get_required_stacks(provider_stack)
+
+            if not required_stacks:
+                if stack_name not in dependencies:
```
phobologic: If the stack is already in dependencies, is ignoring it the right thing?  Might want some debug log statements in here.

stacker/providers/aws.py
```diff
@@ -0,0 +1,98 @@
...
+
+import boto
+from boto import cloudformation
+
+from .base import BaseProvider
+
+logger = logging.getLogger(__name__)
+
+
+class Provider(BaseProvider):
```
phobologic: Same comment as with Actions - should this be called AWSProvider to make it more explicit/easier to read?

stacker/actions/build.py
mhahn: I prefer to have the filename set the context for the class it contains.

this isn't a very strong preference so I can rename if you want.

stacker/actions/destroy.py
mhahn: this is preventing us from overriding any dependencies that may already be listed for the stack with an empty list.

stacker/providers/aws.py
mhahn: yea, again, letting the filename specify the context. i can go more explicit if you'd like!

phobologic:
Ok!  It looks like we're finally there.  Thanks for sticking with this @mhahn - and next time hopefully the PRs will be smaller ;)

PR #36, opened 2015-06-19
Fix unchanged rollback stack issue

phobologic:
When a stack update fails & rolls back to an old state, it leaves the
stack in ROLLBACK_COMPLETE.  Unfortunately since this isn't recognized
as a 'COMPLETE' status (it's hard to know if it's complete or not) this
left the entire meta-stack in an un-updateable state.

This handles the case where we 'skip' a stack because nothing has
changed a little bit better, and seems to actually speed things up since
we deal with skipped stacks earlier.


bmarini:
:+1:

PR #40, opened 2015-07-16
Internal hostname output conditional now

phobologic:
We made the internal hostname conditional, but not the output of the
internal hostname.  This fixes that.


benjamin-guillet:
LGTM :+1:

PR #43, opened 2015-07-23
Example Production Empire Stacks

phobologic:
This most definitely isn't perfect, but it fits with the way that we
(Remind) have been doing Empire production stacks, and I think it's a
good example for everyone else to start playing with.

Making this Release 0.4.1


stacker/blueprints/empire/empire_controller.py
```diff
@@ -0,0 +1,305 @@
...
+        'AvailabilityZones': {'type': 'CommaDelimitedList',
+                              'description': 'Availability Zones to deploy '
+                                             'instances in.'},
+        'InstanceType': {'type': 'String',
+                         'description': 'Empire AWS Instance Type',
+                         'default': 'm3.medium'},
+        "MinHosts": {"type": "Number",
+                     "description": "Minimum # of empire minion instances.",
+                     "default": "2"},
+        "MaxHosts": {"type": "Number",
```
ejholmes: Super minor, but this formatting is a little hard to read. Is there some pep recommended format? I think it would be easier to read as (NL after key):

```python
"MaxHosts": {
  "type": "Number",
  "description": "",
}
```

Also, mixed double and single quotes.

ejholmes:
What do you think about having some kind of approvals based test suite? Basically, just generate the CF JSON for the stacks and compare them. If anything, it would be nice as a sanity check when you make a change that includes python logic.

ejholmes:
Cool! This looks really nice to me. It'll be nice if we can rebase our private stack on the open source.

phobologic:
re: Dumping json & comparing: Yeah, that might be hard in this version - that said there's a pull request in process by @mhahn which will make writing tests like that a lot easier.

Mostly it'd just catch the 'oh, we thought we made a non-changing change but were wrong' stuff.  It's hard to test this since it's so heavily tied into third party stuff.

Re: basing our internal stacks off of this - our stacks are sort of based off this (though it's the other way around) and we let ansible do more of the work, with less variables being managed in cloudformation, which is kind of nice.  @bmarini was talking about that too - I'll see if there's a nice middle ground we can reach.

stacker/blueprints/empire/empire_base.py
```diff
@@ -0,0 +1,55 @@
...
+    def create_load_balancer(self):
+        logger.debug("No load_balancer to setup for %s", self.name)
+
+    def create_iam_profile(self):
+        logger.debug("No iam_profile to setup for %s", self.name)
+
+    def create_autoscaling_group(self):
+        logger.debug("No autoscaling_group to setup for %s", self.name)
+
+    def create_instances(self):
```
mhahn: no longer used

stacker/blueprints/empire/empire_base.py
```diff
@@ -0,0 +1,55 @@
...
+        return stanza
+
+    def create_template(self):
+        self.create_conditions()
+        self.create_security_groups()
+        self.create_ecs_cluster()
+        self.create_load_balancer()
+        self.create_iam_profile()
+        self.create_autoscaling_group()
+        self.create_instances()
```
mhahn: no longer used

mhahn:
I just tried bringing this up from scratch and am running into the following issues after everything builds successfully:

on both minions (i only brought up 2 instead of 3):
```
$ docker logs `docker ps -l -q`
2015-07-24T07:49:47Z [INFO] Starting Agent: Amazon ECS Agent - v1.2.1 (5da1555)
2015-07-24T07:49:47Z [INFO] Loading configuration
2015-07-24T07:49:48Z [INFO] Checkpointing is enabled. Attempting to load state
2015-07-24T07:49:48Z [CRITICAL] Error creating state manager: stat '/data/': no such file or directory
```

If I remove the `ECS_DATADIR` and `ECS_CHECKPOINT` from the ecs_agent.env, I get the following:
```
# docker logs `docker ps -l -q`
2015-07-24T07:51:32Z [INFO] Starting Agent: Amazon ECS Agent - v1.2.1 (5da1555)
2015-07-24T07:51:32Z [INFO] Loading configuration
2015-07-24T07:51:33Z [INFO] Checkpointing not enabled; a new container instance will be created each time the agent is run
2015-07-24T07:51:33Z [INFO] Registering Instance with ECS
2015-07-24T07:51:33Z [ERROR] Could not register module="api client" err="Cluster not found."
2015-07-24T07:51:33Z [ERROR] Error registering: Cluster not found.
```

the `ECS_CLUSTER` variable is referencing an existing cluster.

on the controllers i get this:
```
$ docker logs `docker ps -l -q`
2015/07/24 07:47:12 [Driver '' not found.]
```

the only variables i didn't populate within `example.env` were:
```
empire_controller_cert_name: "''"

empire_controller_github_client_id:
empire_controller_github_client_secret:
empire_controller_github_organization:
```

but i don't think those would be causing these issues.

phobologic:
I think the new AMI should fix the issues that @mhahn had.

phobologic:
Ok - verified that the new AMI fixed things.  @mhahn was able to get an environment up and running, and it sounds like it went smoothly once he tried the new AMI.  I'm going to go ahead and merge this.

PR #46, opened 2015-07-29
Subcommands documentation

mhahn:
Depends on the `subcommands` PR.

Added/updated docstrings to be inline with: http://sphinxcontrib-napoleon.readthedocs.org/en/latest/example_google.html


phobologic:
Cool - thx.

PR #57, opened 2015-08-06
Support encrypting the RDS instance

mhahn:


brianz:
:+1:

phobologic:
Have you tested whether this will cause a stack update if you are coming from a stack that was built before this feature?  That'd be bad (since it'd likely replace the database).

This brings up the versioning of blueprints like I was talking about - otherwise there's going to be a lot of conditionals just to avoid this sort of thing going forward.

If you haven't tested this, I'll try to get to it.

mhahn:
Yea thats a really good point. I have not tested that since I was using this to write stacks from scratch.

Do you have another thread where we can talk about the versioning?

phobologic:
#58 just created

phobologic:
So I just got around to testing this - and it does result in a rebuild of the database - which sucks, because it basically means that the existing database gets nuked.  I think a conditional and two DBInstance blocks would avoid this, though I'm not 100% certain.

phobologic:
IT's really annoying - since the databases it is building are exactly the same:

https://www.dropbox.com/s/w3uvu6ut7fwoi22/Screenshot%202015-08-10%2011.48.30.png?dl=0

mhahn:
thats so annoying.

the versioning seems like the more longterm solution to this.

phobologic:
Agreed - though that's a pretty significant change, and I still haven't gotten my head around how we'd make it work.

mhahn:
yea, i might just change this to have a `get_rds_parameters` or something that can be overriden by a subclass so i can have encrypted dbs that are in my private stacks.

phobologic:
Yeah - that seems like a decent compromise for now.

phobologic:
No longer necessary w/ new RDS v2 blueprints.

PR #59, opened 2015-08-06
Add destroy action

mhahn:
The destroy action no longer relies on the "required_stacks" tag within
CloudFormation since we can't rely on updating those values. Destroy
will determine the order in which resources should be destroyed based on
any build dependencies determined from the configuration.

### DIFF
--- a/stacker/commands/base.py
+++ b/stacker/commands/base.py
@@ -34,6 +34,7 @@ class BaseCommand(object):
                 )
                 subcommand.add_arguments(subparser)
                 subparser.set_defaults(run=subcommand.run)
+                subparser.set_defaults(get_context_kwargs=subcommand.get_context_kwargs)
 
     @property
     def logger(self):
@@ -66,3 +67,17 @@ class BaseCommand(object):
 
     def configure(self, options, **kwargs):
         pass
+
+    def get_context_kwargs(self, options, **kwargs):
+        """Return a dictionary of kwargs that will be passed when setting up the Context.
+
+        This allows commands to pass in any specific arguments they define to the context.
+
+        Args:
+            options (argparse.Namespace): arguments that have been passed via the command line
+
+        Returns:
+            dict: Dictionary that will be passed to Context initializer as kwargs.
+
+        """
+        return {}
--- a/stacker/commands/stacker/destroy.py
+++ b/stacker/commands/stacker/destroy.py
@@ -23,7 +23,7 @@ class Destroy(StackerCommand):
     def run(self, options, **kwargs):
         super(Destroy, self).run(options, **kwargs)
         action = destroy.Action(options.context, provider=options.provider)
-        stack_names = map(lambda x: '  - %s' % (x.fqn,), options.context.get_stacks())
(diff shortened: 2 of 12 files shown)


stacker/commands/base.py
```diff
@@ -66,3 +67,17 @@ def run(self, options, **kwargs):
 
     def configure(self, options, **kwargs):
         pass
+
+    def get_context_kwargs(self, options, **kwargs):
```
phobologic: Maybe name this `additional_context`?  No strong opinion here, but it took me a moment to understand what you were doing.

stacker/stack.py
```diff
@@ -57,7 +57,7 @@ def __repr__(self):
 
     @property
     def requires(self):
-        requires = set(self.definition.get('requires', []))
+        requires = set(map(self.context.get_fqn, self.definition.get('requires', [])))
```
phobologic: So, a style thing, but I think a list comprehension would be easier to read here:

```
requires = [self.context.get_fqn(r) for r in self.definition.get('requires', [])]
```

stacker/commands/stacker/destroy.py
```diff
@@ -0,0 +1,36 @@
...
+    def add_arguments(self, parser):
+        super(Destroy, self).add_arguments(parser)
+        parser.add_argument('-f', '--force', action='store_true',
+                            help="Whehter or not you want to go through "
+                                 " with destroying the stacks")
+
+    def run(self, options, **kwargs):
+        super(Destroy, self).run(options, **kwargs)
+        action = destroy.Action(options.context, provider=options.provider)
+        stack_names = map(lambda x: '  - %s' % (x.fqn,), options.context.get_stacks())
```
phobologic: again, I think list comprehensions would be better here:

```
["  - %s" % x.fqn for x in options.context.get_stacks()]
```

stacker/commands/base.py
mhahn: well it literally is getting kwargs to pass to the context. adding "additional" implies there are some other kwargs they're being merged with

stacker/commands/stacker/destroy.py
```diff
@@ -0,0 +1,36 @@
...
+        message = '\nAre you sure you want to destroy the following stacks:\n%s\n\n(yes/no): ' % (
```
phobologic: Just a convention, but maybe capitalize `No` to indicate that it's the default if you just hit enter.

stacker/context.py
```diff
@@ -64,5 +64,13 @@ def get_stacks_dict(self):
...
 
     def get_fqn(self, name=None):
-        """Return the fully qualified name of an object within this context."""
+        """Return the fully qualified name of an object within this context.
+
+        If the name passed already appears to be a fully qualified name, it
+        will be returned with no further processing.
+
+        """
+        if name and name.startswith(self._base_fqn + '-'):
```
phobologic: awesome

phobologic:
:+1: Awesome, thanks.

PR #61, opened 2015-08-06
Local parameters for python

phobologic:
This allows the definition of local parameters that are meant to be used
by python, not by cloudformation.  The parameters are still passed into
the blueprint the same way (from the command line, or the config) but
they can be used without having to share them with the template.


mhahn:
:ship:

PR #62, opened 2015-08-07
Add ability to lock stacks

phobologic:


stacker/actions/build.py
```diff
@@ -175,17 +183,19 @@ def _handle_missing_parameters(self, params, required_params,
 
         return params.items()
 
-    def _generate_plan(self):
+    def _generate_plan(self, force_stacks=[]):
```
mhahn: shouldn't have mutable values as kwargs

brianz:
:+1:

mhahn:
The concept of "locking" stacks seems important enough that every action should be able to have some support for handling locked stacks if necessary. For example, the destroy action should definitely not destroy something if its locked.

With that in mind, I would suggest something like:

- add `force_stacks` as a parameter here: https://github.com/remind101/stacker/blob/master/stacker/commands/stacker/base.py#L57
- add it as an optional key here: https://github.com/remind101/stacker/blob/master/stacker/context.py#L13
- configure the stack to have `force_update` based on the stacks specified in `force_stacks` here: https://github.com/remind101/stacker/blob/master/stacker/context.py#L52
- pass `force_stacks` into the context here: https://github.com/remind101/stacker/blob/master/stacker/commands/stacker/__init__.py#L21

that way any action can just look at the `stack` and check `stack.locked and stack.force_update`

right now it seems like we have logic for "locked" in the Stack object but then the logic for determining whether or not something is being forced is just within the build action.

what do you think?

phobologic:
I actually thought about that when I was writing this, and it didn't seem like a viable use case.  Destroy, by it's very nature, is destructive.  When you say 'stacker destroy' you mean to destroy things.

When you do a build though, you don't always know what/when you are going to update.  That was what I thought this feature was more about.  I think if we required 'force' when doing a destroy it'd actually be kind of a PITA.

mhahn:
Yea thats a good point, I already have a confirmation prompt within destroy.

i guess i just feel like since the `locked` parameter is now a part of the stack definition, the `force` parameters should also be available on the stack.

ie. `stack.locked and stack.force_update` seem cleaner to me than passing everything through build.

phobologic:
Yeah, let me dig into that suggestion - I focused too much on the first part of your comment (because I already had some reasoning for why I did it this way) and less on the second part (because I actually have to look at code :)).  Let me dig into it.  From your comment it does sound like a cleaner interface.

mhahn:
it puts more logic in the context, but thats mainly because it has the `get_stacks` function, which maybe should go somewhere else

phobologic:
@mhahn take a look now, I think this matches closer to what you had in mind.  Not sure about should_update being in the build action - though it seems like it's pretty specific to that action, so I left it there for now.

stacker/context.py
```diff
@@ -50,6 +54,8 @@ def get_stacks(self):
                     context=self,
                     parameters=self.parameters,
                     mappings=self.mappings,
+                    force=stack_def['name'] in self.force_stacks,
+                    locked=stack_def.get('locked', False),
```
phobologic: There could be some confusion here if someone puts `locked: FALse` or something similar, since that will get evaluated to the string `FALse` rather than the python False, and so it will be locked.  That said, I think a worse case would be someone including `TRue` and not getting True.  Not sure the best way to handle this.

stacker/context.py
```diff
@@ -10,13 +10,17 @@ class Context(object):
 
     """
 
-    _optional_keys = ('environment', 'stack_names', 'parameters', 'mappings',
-                      'config')
-
-    def __init__(self, namespace, **kwargs):
+    def __init__(self, namespace, environment=None, stack_names=None,
```
phobologic: `_optional_keys` wasn't providing us with much, so I went ahead with just definining all the arguments and giving them reasonable defaults.  landscape.io will complain about the number of arguments, but not sure we want an options object for this.

phobologic: Actually - I should probably get rid of kwargs now, since we don't do anything with it.  This will help with testing.

phobologic:
Tested with the empire.yml stack, locking the db: ```
# stacker build -e remind/repos/stacker/conf/empire-mike.env -r us-east-1 mike-classchirp-com remind/repos/stacker/conf/empire/empire.yaml
[2015-08-11T18:29:50] Executing pre_build hooks: stacker.hooks.route53.create_domain
[2015-08-11T18:29:50] Launching stacks: mike-classchirp-com-vpc, mike-classchirp-com-empireDB, mike-classchirp-com-empireMinion, mike-classchirp-com-empireController, mike-classchirp-com-bastion
[2015-08-11T18:29:51] Launching stack mike-classchirp-com-vpc now.
[2015-08-11T18:29:52] Attempting to update stack mike-classchirp-com-vpc.
[2015-08-11T18:29:52] Stack mike-classchirp-com-vpc did not change, not updating.
[2015-08-11T18:29:52] Launching stack mike-classchirp-com-empireDB now.
[2015-08-11T18:29:53] Stack empireDB locked and not in --force list. Refusing to update.
[2015-08-11T18:29:53] Launching stack mike-classchirp-com-empireMinion now.
[2015-08-11T18:29:53] Attempting to update stack mike-classchirp-com-empireMinion.
[2015-08-11T18:29:53] Stack mike-classchirp-com-empireMinion did not change, not updating.
[2015-08-11T18:29:53] Launching stack mike-classchirp-com-empireController now.
[2015-08-11T18:29:54] Attempting to update stack mike-classchirp-com-empireController.
[2015-08-11T18:29:54] Stack mike-classchirp-com-empireController did not change, not updating.
[2015-08-11T18:29:55] Launching stack mike-classchirp-com-bastion now.
[2015-08-11T18:29:55] Attempting to update stack mike-classchirp-com-bastion.
[2015-08-11T18:29:55] Stack mike-classchirp-com-bastion did not change, not updating.
[2015-08-11T18:29:55] Plan Status:
[2015-08-11T18:29:55]   - step "mike-classchirp-com-vpc": skipped
[2015-08-11T18:29:55]   - step "mike-classchirp-com-empireDB": skipped
[2015-08-11T18:29:55]   - step "mike-classchirp-com-empireMinion": skipped
[2015-08-11T18:29:55]   - step "mike-classchirp-com-empireController": skipped
[2015-08-11T18:29:55]   - step "mike-classchirp-com-bastion": skipped
```

ejholmes:
Sweet. lgtm :+1:

PR #71, opened 2015-08-13
Create a BaseCommandMixin

brianz:
The way BaseCommand was constructed was a bit confusing, having two base.py files near each other.
It was also being used as a mixin, rather than a base class.

This doesn't change any functionality but moves the higher level BaseCommand to a mixin class which
makes thing easier to read and is more inline with how the class was being used in the first place.


phobologic:
The commands/base.py & commands/stacker/base.py thing was pretty confusing for me.  Not sure we need to call it a Mixin, since it's just a parent class at this point (also, I'm allergic to mixins - multiple parent inheritance is my nightmare).  I almost wonder if we should just move everything in commands/stacker up into commands. @mhahn ?

mhahn:
yea this was pretty annoying.

my thought process was:

`stacker/commands/base.py`

is the base command for all future commands, includes logic for adding subcommands, logging, running etc.

`stacker/commands/stacker/base.py`

was my way around argparse not having great support for global parameters that get passed to subcommands.

what I wanted to ensure was that:

1. You could call a subcommand like this `stacker build -p UsingGlobalParameter=Foo` instead of having to do this `stacker -p UsingGlobalParameter=Foo build`.
2. I wanted to ensure that if you ran `stacker build -h` you saw the help from the global parameters.

I couldn't find a way to do this using argparse by default, so I created `stacker/commands/stacker/base.py` which was meant to be the base command for all the `stacker` subcommands since. By overriding `add_arguments` any subcommand that called super would inherit the global parameters as first class arguments, solving the above 2 issues.

maybe renaming `stacker/commands/stacker/base.py` to `StackerGlobalParametersMixin` makes more sense though?

brianz:
Like @phobologic mentioned, I did this b/c it was confusing have two `base.py` files and it took me a while to figure out what that was about.  I think this change keeps the intended behavior....at least from my limited testing @mhahn ...can you confirm?  I'm still not sure I understand how `stacker/commands/base.py` would ever be used outside of a stacker command.

Calling it a Mixin or not, having that class live in `stacker/base.py` is a lot more clear for me.

mhahn:
i think this simplifies things and doesn't add Mixin or anything else:

https://github.com/remind101/stacker/pull/74/files

mhahn:
my initial thinking was that we might have more than just the "stacker" command published by the library, but that doesn't seem relevant in the short term

phobologic:
Does #74 obviate the need for this?  Should I go ahead and close this?

mhahn:
Yea

PR #72, opened 2015-08-13
Make environment a required parameter.

mhahn:
Environment must at least contain a "namespace" value.

### DIFF
--- a/conf/empire/example.env
+++ b/conf/empire/example.env
@@ -8,6 +8,12 @@
 # Note: The minimum allowed azcount is 2.
 azcount: 2
 
+# namespace is a unique name that the stacks will be built under. This value
+# will be used to prefix the CloudFormation stack names as well as the s3
+# bucket that contains revisions of the stacker templates. This is the only
+# required environment variable.
+namespace:
+
 # An external domain where empire.<domain> CNAME will be created
 external_domain:
 ssh_key_name:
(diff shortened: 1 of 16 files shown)


phobologic:
I like this - one thing: Can you update the README to match the syntax here?

Also, going to share it on the slack channel again - make sure noone has any other opinion about this.

brianz:
:+1: I like this change as well.

mhahn:
@phobologic just pushed an update to the README, let me know if that doesn't work.

conf/empire/example.env
```diff
@@ -1,3 +1,4 @@
+namespace: empire.example
```
phobologic: Lets remove the default here - the reason is that since the namespace is used for the s3 bucket name, this'll likely cause confusion.  Give a good comment to to explain what people should set this too.

conf/stage.env
```diff
@@ -1,4 +1,7 @@
-# This is just an example of the format for environment files
+# This is an example of the format for environment files
 # You can use a flag dictionary in yaml format. Values can be strings or
-# numbers, but cannot be more complex data types (dictionaries, lists, etc)
+# numbers, but cannot be more complex data types (dictionaries, lists, etc).
+#
+# NOTE: "namespace" is the only required environment variable
+namespace: stacker.example
```
phobologic: Same as above.

phobologic:
:+1:

PR #73, opened 2015-08-13
Add action for getting information on stacks

mhahn:
Currently just logs the outputs.

### DIFF
--- a/stacker/actions/info.py
+++ b/stacker/actions/info.py
@@ -1,0 +1,25 @@
+import logging
+
+from .base import BaseAction
+
+logger = logging.getLogger(__name__)
+
+
+class Action(BaseAction):
+    """Get information on CloudFormation stacks.
+
+    Displays the outputs for the set of CloudFormation stacks.
+
+    """
+
+    def run(self, *args, **kwargs):
+        logger.info('Outputs for stacks: %s', self.context.get_fqn())
+        for stack in self.context.get_stacks():
+            provider_stack = self.provider.get_stack(stack.fqn)
+            if not provider_stack:
+                logger.info('Stack "%s" does not exist.' % (stack.fqn,))
+                continue
+
+            logger.info('%s:', stack.fqn)
+            for output in provider_stack.outputs:
+                logger.info('\t%s: %s', output.key, output.value)
--- a/stacker/actions/info.py
+++ b/stacker/actions/info.py
@@ -1,6 +1,7 @@
 import logging
 
 from .base import BaseAction
+from .. import exceptions
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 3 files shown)


stacker/actions/info.py
```diff
@@ -0,0 +1,25 @@
...
+
+class Action(BaseAction):
+    """Get information on CloudFormation stacks.
+
+    Displays the outputs for the set of CloudFormation stacks.
+
+    """
+
+    def run(self, *args, **kwargs):
+        logger.info('Outputs for stacks: %s', self.context.get_fqn())
```
phobologic: Does this actually work?  Doesn't get_fqn need a stack name?

stacker/actions/info.py
```diff
@@ -0,0 +1,25 @@
...
+        for stack in self.context.get_stacks():
+            provider_stack = self.provider.get_stack(stack.fqn)
```
phobologic: This will raise a StackDoesNotExist exception.

stacker/commands/stacker/info.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+class Info(StackerCommand):
+
+    name = 'info'
+    description = __doc__
+
+    def run(self, options, **kwargs):
+        super(Info, self).run(options, **kwargs)
+        action = info.Action(options.context, provider=options.provider)
```
phobologic: Probably worth adding an argument to get the output from a single stack if specified.

phobologic: Also, might be cool to have a --output or --format argument.  Maybe not necessary here, but I could see someone wanting this in a machine readable format (like csv).

stacker/commands/stacker/info.py
```diff
@@ -0,0 +1,15 @@
+"""Gets information on the CloudFormation stacks based on the given config."""
+
+from .base import StackerCommand
+from ...actions import info
+
+
+class Info(StackerCommand):
```
phobologic: Does Info sound right to you?  Maybe just Outputs for now or something like that?

mhahn: Yea, I just wasn't sure if we wanted to put more information in their like
current status or something

stacker/actions/info.py
mhahn: this defaults to simply the namespace

stacker/commands/stacker/info.py
mhahn: maybe including like:

- created date
- last changed date

would be useful and fit more with info?

haven't looked into whether or not we have access to that info

phobologic: Yeah, that's fine.  If you think there's more info we can provide, lets leave it as Info - doesn't necessarily need to happen in this PR.

stacker/commands/stacker/info.py
mhahn: lets look into the --output or --format in the next revision

PR #74, opened 2015-08-13
Simplify the BaseCommand classes

mhahn:


brianz:
:+1: as long as you update `command/stacker/__init__.py`...see tests.

mhahn:
ah, left over `pyc` files on my end were causing tests to pass locally

PR #78, opened 2015-08-16
Add more API docs

phobologic:


phobologic:
btw, you can view the rendered API docs here: http://stacker.readthedocs.org/en/api_docs/api/modules.html

PR #86, opened 2015-09-01
Initial attempt at new RDS blueprint framework

phobologic:
Still a WIP - mostly opening for discussion about the new
framework.


stacker/blueprints/rds/base.py
```diff
@@ -0,0 +1,336 @@
...
+class MasterInstance(Blueprint):
+    """ Blueprint for a generic Master RDS Database Instance.
+
+    Subclasses should be created for each RDS engine for better validation of
+    things like engine version.
+    """
+
+    ENGINE = None
+
+    def _get_engine_versions(self):
```
mhahn: if something should be overridden, i usually don't put an underscore since it implies some sort of public api

stacker/blueprints/rds/base.py
```diff
@@ -0,0 +1,336 @@
...
+        self.create_security_group()
+        self.create_rds()
+        self.create_dns_records()
+        self.create_db_outputs()
+
+
+class ReadReplica(MasterInstance):
+    """ Blueprint for a Read replica RDS Database Instance. """
+    def _get_parameters(self):
+        parameters = common_parameters()
```
mhahn: why not just do:

```
parameters = super(ReadReplica, self).get_parameters()
```
and have `common_parameters` returned

stacker/blueprints/rds/base.py
```diff
@@ -0,0 +1,336 @@
...
+        endpoint = GetAtt(self.get_db_instance(), "Endpoint.Address")
+        return endpoint
+
+    def create_rds(self):
+        t = self.template
+        # Non-provisioned iops database
+        t.add_resource(
+            DBInstance(
+                DBINSTANCE_NO_IOPS,
+                Condition="NoProvisionedIOPS",
```
mhahn: i would pull out all of these common parameters so i don't have to duplicate specifying them in each

stacker/blueprints/rds/base.py
```diff
@@ -0,0 +1,336 @@
...
+                           "replica of."}
+        return parameters
+
+    def create_rds(self):
+        t = self.template
+        # Non-provisioned iops database
+        t.add_resource(
+            DBInstance(
+                DBINSTANCE_NO_IOPS,
+                Condition="NoProvisionedIOPS",
```
mhahn: same here re: duplication of db parameters

stacker/blueprints/rds/base.py
phobologic: There are parameters that come from the MasterInstance that shouldn't be used in ReadReplicas. A better design might be to have a BaseInstance type or something, and have MasterInstance & ReadReplica both subclass off of it, rather than ReadReplica being based on MasterInstance.

stacker/blueprints/rds/base.py
```diff
@@ -0,0 +1,336 @@
...
+
+        Should only be overridden if the class variable ENGINE is defined on
+        the class.
+
+        Return:
+            list: A list of valid engine versions for the given engine.
+        """
+        return []
+
+    def _get_parameters(self):
```
mhahn: where is this value used? is this from `Blueprint`?

stacker/blueprints/rds/base.py
phobologic: Hard/ugly to do?  How would you do it?  Just *args/**kwargs?

stacker/blueprints/rds/base.py
mhahn: i see yea, not a big deal, you're just calling `common_parameters` and overriding it in both places.

stacker/blueprints/rds/base.py
phobologic: Yes - there was a change a little while back for this, to make it easy/possible to override Parameter definitions in subclasses.

stacker/blueprints/rds/base.py
mhahn: i was just thinking like:

```python
common = {
    'AllocatedStorage': Ref("AllocatedStorage"),
    'AllowMajorVersionUpgrade': Ref("AllowMajorVersionUpgrade"),
    'AutoMinorVersionUpgrade': Ref("AutoMinorVersionUpgrade"),
    'BackupRetentionPeriod': Ref("BackupRetentionPeriod"),
    'DBName': Ref("DatabaseName"),
    'DBInstanceClass': Ref("InstanceType"),
    'DBSubnetGroupName': Ref(SUBNET_GROUP),
    'Engine': self.ENGINE or Ref("Engine"),
    'EngineVersion': Ref("EngineVersion"),
    'MasterUsername': Ref("MasterUser"),
    'MasterUserPassword': Ref("MasterUserPassword"),
    'MultiAZ': Ref("MultiAZ"),
    'PreferredBackupWindow': Ref("PreferredBackupWindow"),
    'VPCSecurityGroups': [Ref(SECURITY_GROUP), ],
    'StorageEncrypted': Ref("StorageEncrypted"),
}
t.add_resource(
    DBInstance(
        DBINSTANCE_NO_IOPS,
        Condition="NoProvisionedIOPS",
        **common
     )
)
t.add_resource(
    DBInstance(
        DBINSTANCE_WITH_IOPS,
        Condition="ProvisionedIOPS",
        StorageEncrypted=Ref("StorageEncrypted"),
        StorageType="io1",
        Iops=Ref("IOPS"),
        **common
    )
)
```

which also has the added benefit of letting me see whats different about the two.

not sure if thats 100% accurate just a rough copy paste

stacker/blueprints/rds/base.py
```diff
@@ -0,0 +1,336 @@
...
+
+
+def common_parameters():
+    """ Return common parameters for all RDS stacks.
+
+    Returns:
+        dict: A dictionary of parameter definitions.
+    """
+
+    parameters = {
```
mhahn: should we add the snapshot parameter you were talking about too?

stacker/blueprints/rds/base.py
mhahn: yea, i would just make it `get_parameters`, to me underscore implies a private method within this class that no one should mess with.

mhahn:
+1 i like it!

stacker/blueprints/rds/base.py
phobologic: I haven't yet figured out if that's the right thing to do here or not.  There's a lot of 'if this parameter is specified, these parameters shouldn't be' which I handled in Read Replicas by just removing the unused/disallowed ones.  Not sure what the right case should be here.  If you have any ideas, let me know.

stacker/blueprints/rds/base.py
phobologic: So really the definition of private method that I've always followed is any method that won't be called directly by anything but the class itself, not whether it might get overriden in subclasses.  I don't plan on fixing this in this update :)

mhahn: well the subclass overrides it to conform to a specific interface for interacting with the parent class. the underscore implies "i can do whatever i want with this because i told you it could change by having an underscore". if i edited the base blueprint and removed/changed `_get_parameters` all subclasses would break.

mhahn: also, the bikeshed is BLUE!

stacker/blueprints/rds/base.py
mhahn: we could just leave that up to private stacks but make it so we can pass those parameters to the db instance troposphere instance

stacker/blueprints/rds/base.py
phobologic: Hahah - I think that's honestly just a personal interpretation of `_methods`.  In general, I've always read them as 'you probably shouldn't be calling this, but we're not going to make it impossible'.  `__methods` are a whole other ballgame, and basically a nightmare.

In general I don't think anyone would ever want to call `_get_parameters`, but yeah - if it becomes a problem later, it's an easy enough fix.

mhahn:
+1 again :)

PR #87, opened 2015-09-08
Add Dockerfile.

ejholmes:
This adds a method for running stacker to provision stacks using Docker, which can be an easier alternative for the pythonically challenged, and produces a more common environment between developers.

See https://github.com/remind101/private_stacks/pull/42 for an example on how this might be used.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,1 @@
+Dockerfile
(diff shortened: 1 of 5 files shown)


phobologic:
:+1:

PR #89, opened 2015-09-08
Redis cache cluster

mhahn:


stacker/blueprints/redis.py
```diff
@@ -0,0 +1,101 @@
...
+
+
+class RedisCluster(Blueprint):
+    PARAMETERS = {
+        'VpcId': {'type': 'AWS::EC2::VPC::Id', 'description': 'Vpc Id'},
+        'CacheClusterId': {
+            'type': 'String',
+            'description': 'The node group identifier',
+        },
+        'PrivateSubnets': {
```
phobologic: I would change this to Subnets - this lets you choose which subnets to put it in, rather than implicitly assuming private (which may be the safe case, but still).  This is a change I've been making in RDS as well.

stacker/blueprints/redis.py
```diff
@@ -0,0 +1,101 @@
...
+            'description': (
+                'A (minimum 60 minute) window in '
+                'ddd:hh24:mi-ddd:hh24:mi format in UTC for '
+                'maintence. Default: Sunday 2am-3am'
+            ),
+            'default': 'sun:02:00-sun:03:00',
+        },
+        'AutomaticFailoverEnabled': {
+            'type': 'String',
+            'description': (
```
phobologic: I'd set the 'allowed_values' here to true & false

stacker/blueprints/redis.py
```diff
@@ -0,0 +1,101 @@
...
+        t.add_resource(subnet_group)
+
+    def create_security_group(self):
+        t = self.template
+        security_group = ec2.SecurityGroup(
+            SECURITY_GROUP,
+            GroupDescription='%s ElastiCache security group' % (SECURITY_GROUP,),
+            VpcId=Ref('VpcId'),
+        )
+        resource = t.add_resource(security_group)
```
phobologic: Curious why you broke this step up?  resource == security_group in this case.

stacker/blueprints/redis.py
```diff
@@ -0,0 +1,101 @@
...
+
+    def create_replication_group(self):
+        t = self.template
+        replication_group = ReplicationGroup(
+            REPLICATION_GROUP,
+            AutomaticFailoverEnabled=Ref('AutomaticFailoverEnabled'),
+            CacheNodeType=Ref('CacheNodeType'),
+            CacheSubnetGroupName=Ref(SUBNET_GROUP),
+            Engine='redis',
+            EngineVersion='2.8.19',
```
phobologic: Worth making this configurable?  Same with Engine?

stacker/blueprints/redis.py
```diff
@@ -0,0 +1,101 @@
...
+            NumCacheClusters=Ref('NumCacheClusters'),
+            ReplicationGroupDescription='%s replication group' % (REPLICATION_GROUP,),
+            PreferredMaintenanceWindow=Ref('PreferredMaintenanceWindow'),
+            SecurityGroupIds=[GetAtt(SECURITY_GROUP, 'GroupId')],
```
phobologic: Does Ref() not work here?

stacker/blueprints/redis.py
mhahn: re: engine: 

http://docs.aws.amazon.com/AmazonElastiCache/latest/APIReference/API_CreateReplicationGroup.html

replication group is only valid for "redis", its probably just a parameter to leave it extensible in the future.

i'll make the EngineVersion configurable though

stacker/blueprints/redis.py
mhahn: I remember trying this and it didn't work, I'll try again though, its been a couple of months.

stacker/blueprints/redis.py
mhahn: hmm, can't remember. i'll try just `security_group`

stacker/blueprints/redis.py
mhahn: ah, i didn't know you could do that

stacker/blueprints/redis.py
```diff
@@ -0,0 +1,107 @@
...
+        t.add_resource(subnet_group)
+
+    def create_security_group(self):
+        t = self.template
+        security_group = ec2.SecurityGroup(
+            SECURITY_GROUP,
+            GroupDescription='%s ElastiCache security group' % (SECURITY_GROUP,),
+            VpcId=Ref('VpcId'),
+        )
+        t.add_resource(security_group)
```
phobologic: Sorry - I think there was miscommunication here.  Instead of what you have here, why not:

```
    def create_security_group(self):
        t = self.template
        t.add_resource(ec2.SecurityGroup(
            SECURITY_GROUP,
            GroupDescription='%s ElastiCache security group' % (SECURITY_GROUP,),
            VpcId=Ref('VpcId'),
        ))
```

stacker/blueprints/redis.py
```diff
@@ -0,0 +1,107 @@
...
+            REPLICATION_GROUP,
+            AutomaticFailoverEnabled=Ref('AutomaticFailoverEnabled'),
+            CacheNodeType=Ref('CacheNodeType'),
+            CacheSubnetGroupName=Ref(SUBNET_GROUP),
+            Engine='redis',
+            EngineVersion=Ref('EngineVersion'),
+            NumCacheClusters=Ref('NumCacheClusters'),
+            ReplicationGroupDescription='%s replication group' % (REPLICATION_GROUP,),
+            PreferredMaintenanceWindow=Ref('PreferredMaintenanceWindow'),
+            SecurityGroupIds=[GetAtt(SECURITY_GROUP, 'GroupId')],
```
phobologic: Ref didn't work?  This is a VPC security group, right? http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-security-group.html (See the part on Ref/GetAtt)

stacker/blueprints/redis.py
mhahn: this was just a style thing

mhahn:
@phobologic i think this is good to go as well

stacker/blueprints/redis.py
```diff
@@ -0,0 +1,107 @@
+from troposphere import Ref, ec2, Output, GetAtt
+from troposphere.elasticache import ReplicationGroup, SubnetGroup
+
+from .base import Blueprint
+
+SUBNET_GROUP = 'RedisSubnetGroup'
```
phobologic: I've been ditching prefixes like this lately - instead of `RedisSubnetGroup` I just use `SubnetGroup`.  Not sure there is much point to prefixing them.

mhahn:
closing in favor of: https://github.com/remind101/stacker/pull/124/files?diff=unified

PR #91, opened 2015-09-09
Add --version to base command

mhahn:


mhahn:
output:
```
$ stacker --version
stacker 0.5.0
```

stacker/commands/stacker/base.py
```diff
@@ -132,9 +133,6 @@ def get_context_kwargs(self, options, **kwargs):
         return {}
 
     def add_arguments(self, parser):
-        if not self.subcommands:
-            parser.set_defaults(run=self.run)
```
phobologic: This doesn't break the subcommands?

phobologic:
`0.5.0`!? We already have `0.5.1` out!  What're you doing!? ;)

Looks good.

stacker/commands/stacker/base.py
mhahn: no, we had stopped calling `add_arguments` on the base class, so this code was never getting called.

PR #94, opened 2015-09-10
Environment file should be a simple key: value pair file.

mhahn:
We no longer want to double yaml parse the environment file. This treats
the environment as a group of dumb key: value pairs, where the raw
string value will be passed to the yaml config.

### DIFF
--- a/stacker/commands/stacker/base.py
+++ b/stacker/commands/stacker/base.py
@@ -151,7 +151,7 @@ class BaseCommand(object):
                                  "up to twice.")
         parser.add_argument('environment', type=environment_file,
                             default={},
-                            help="Path to a simple key/value pair environment "
+                            help="Path to a simple `key: value` pair environment "
                                  "file. The values in the environment file can "
                                  "be used in the stack config as if it were a "
                                  "string.Template type: "
--- a/stacker/commands/stacker/base.py
+++ b/stacker/commands/stacker/base.py
@@ -2,7 +2,7 @@ import argparse
 from collections import Mapping
 import logging
 
-import yaml
+from ...environment import parse_environment
 
 DEBUG_FORMAT = ('[%(asctime)s] %(levelname)s %(name)s:%(lineno)d'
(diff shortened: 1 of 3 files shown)


stacker/commands/stacker/base.py
```diff
@@ -149,11 +149,12 @@ def add_arguments(self, parser):
         parser.add_argument("-v", "--verbose", action="count", default=0,
                             help="Increase output verbosity. May be specified "
                                  "up to twice.")
-        parser.add_argument('environment', type=yaml_file_type,
+        parser.add_argument('environment', type=environment_file,
                             default={},
-                            help="Path to a yaml environment file. The values in "
-                                 "the environment file can be used in the stack "
-                                 "config as if it were a string.Template type: "
+                            help="Path to a simple key/value pair environment "
```
phobologic: `key: value` maybe better?

stacker/environment.py
```diff
@@ -0,0 +1,11 @@
+
+
+def parse_environment(raw_environment):
+    environment = {}
+    for line in raw_environment.split('\n'):
```
phobologic: Should probably handle `#comments` as well

stacker/environment.py
```diff
@@ -0,0 +1,11 @@
...
+        if ':' not in line:
```
phobologic: Hmm - this is a little too loose.  You're trying to catch blank lines, right?

I'd maybe go with:

```
if not line:
    continue

try:
    key, value = line.split(":", 1)
except ValueError:
    raise Exception("Environment must be in key: value format")
```

stacker/tests/test_environment.py
```diff
@@ -0,0 +1,22 @@
+import unittest
+
+from stacker.environment import parse_environment
+
+test_env = """key1: value1
```
phobologic: Per above - add blank lines & comments just to make sure we don't miss anything.

mhahn:
@phobologic comments addressed

PR #95, opened 2015-09-11
Support custom constructors

mhahn:
Tried to keep this as simple as possible now. I only have a single use case and we can extend as we go.


stacker/config/constructors.py
```diff
@@ -0,0 +1,39 @@
...
+        $ vault write secret/hello value=world
+
+        # In stacker we would reference "world" with the following
+        conf_key: !vault secret/hello@value
+
+        # The above would resolve to
+        conf_key: world
+
+    """
+    if '@' not in value:
```
phobologic: I'd just catch the valueError that will come if you try to split a field and it doesn't have an @ in it.

stacker/config/constructors.py
```diff
@@ -0,0 +1,39 @@
...
+        raise TypeError('Vaulted vaules must include a "@<key in stored data>"')
+
+    path, key = value.split('@')
```
phobologic: I'd use the maxsplit argument to split here, just in case.

stacker/config/constructors.py
```diff
@@ -0,0 +1,39 @@
...
+    response = subprocess.check_output(['vault', 'read', '-format=json', path])
```
phobologic: not a fan of using subprocess in a library... do they not have an API?

mhahn: they do have an HTTP API, but then I have to know more about the vault endpoint that is set as well as auth. if i was going to require having that setup knowledge, i would just want to think about how we let other people register these and then be able to use subprocess.

what if we had a "translators" parameter?

```
stacker build -r us-east-1 conf/staging.env conf/example.yaml --translators my_private_repo/translators.py
```

stacker/config/translators.py
```diff
@@ -0,0 +1,43 @@
...
+    """
+    try:
+        path, key = value.split('@', 1)
+    except ValueError:
+        raise TypeError(
+            'Vaulted vaules must be of the format'
+            ' "<path in vault>@<key in stored data>" (got %s)' % (value,)
+        )
+
+    response = subprocess.check_output(['vault', 'read', '-format=json', path])
```
phobologic: If this fails, what happens here?

mhahn: throws a CallProcessError. I was thinking about catching it, but there are a lot of reasons it could have failed.

stacker/config/__init__.py
```diff
@@ -3,7 +3,10 @@
 
 import yaml
 
-from . import exceptions
+from .. import exceptions
+
+# register translators (yaml constructors)
+from .translators import *  # NOQA
```
phobologic: Why not import exactly what you want here?  Not a fan of * imports.

mhahn: yea, me either. this triggers the registration of the yaml constructors though. otherwise we would be importing something we didn't need. i could do something like within translators we append the constructors we want registered to a list, import that list and then register them with yaml within the config.

PR #98, opened 2015-09-11
Ensure cert exists

mhahn:
example:

```
pre_build:
  - path: stacker.hooks.iam.ensure_server_cert_exists
    required: true
    args:
      cert_name: ${vault_elb_cert_name}
  - path: stacker.hooks.iam.ensure_server_cert_exists
    required: true
    args:
      cert_name: ${empire_controller_cert_name}
```

makes it easy to use private certs for different environments

### DIFF
--- a/stacker/hooks/iam.py
+++ b/stacker/hooks/iam.py
@@ -78,7 +78,7 @@ def ensure_server_cert_exists(region, namespace, mappings, parameters, **kwargs)
                 continue
 
             path = raw_input('Path to %s (skip): ' % (key,))
-            if path == 'skip':
+            if path == 'skip' or not path.strip():
                 continue
 
             full_path = utils.full_path(path)
--- a/stacker/hooks/iam.py
+++ b/stacker/hooks/iam.py
@@ -68,12 +68,15 @@ def ensure_server_cert_exists(region, namespace, mappings, parameters, **kwargs)
             return False
 
         paths = {
-            'certificate': None,
-            'private_key': None,
-            'chain': None,
+            'certificate': kwargs.get('path_to_certificate'),
+            'private_key': kwargs.get('path_to_private_key'),
+            'chain': kwargs.get('path_to_chain'),
         }
 
-        for key in paths.keys():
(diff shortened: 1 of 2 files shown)


stacker/hooks/iam.py
```diff
@@ -30,3 +34,82 @@ def create_ecs_service_role(region, namespace, mappings, parameters,
...
+
+        paths = {
+            'certificate': None,
+            'private_key': None,
+            'chain': None,
+        }
+
+        for key in paths.keys():
+            path = raw_input('Path to %s (skip): ' % (key,))
+            if path == 'skip':
```
phobologic: This makes it look like skip is the default, but that doesn't seem to be the case, or am I confused?

stacker/hooks/iam.py
```diff
@@ -30,3 +34,82 @@ def create_ecs_service_role(region, namespace, mappings, parameters,
...
+    except BotoServerError:
+        upload = raw_input(
+            'Certificate "%s" wasn\'t found. Upload it now? (yes/no) ' % (
+                cert_name,
+            )
+        )
+        if upload != 'yes':
+            return False
+
+        paths = {
```
phobologic: worth maybe making this something you can have in the hook definition?  So you can define the paths to the certs on the local FS?

stacker/hooks/iam.py
mhahn: skip isn't the default, if you explicitly write "skip" we'll skip. if you don't enter anything, you'll get a path doesn't exist error

stacker/hooks/iam.py
mhahn: good idea

mhahn:
@phobologic comments addressed

stacker/hooks/iam.py
phobologic: `(skip)` says to me `this is the default` - maybe be a bit more explicit here?

mhahn: ok, skip is the default now

PR #100, opened 2015-09-11
User data parameter asg

mhahn:
I think we should come up with a more extensible ASG blueprint.

I needed the following changes to push an ASG for an ECS cluster and was in a hurry so didn't think too much about the structure. 

Given that ASG are such a core piece of infrastructure, I thought I would open this up so we could discuss refactoring them to be more extensible.

From working with it, i think we should support:

- Customiznig the LaunchConfiguartion parameters
- Customizing the AutoScalingGroup parameters


stacker/blueprints/asg.py
```diff
@@ -165,9 +165,18 @@ def create_load_balancer(self):
                     GetAtt(elb_name, 'DNSName')],
                 Condition="SetupELBDNS"))
 
+    def get_launch_configuration_parameters(self):
```
phobologic: I almost feel like this should return all of the parameters - and if people want to override any of them, they in theory could?

mhahn: like ImageId etc?

mhahn: I feel like we should come up with a standard for doing this. Anytime I have a slight divergence from one of the base blueprints, I have to either create a pull request to support a parameter that wasn't initially included or create my own subclass and copy paste.

phobologic: With the way parameters are handled now you shouldn't have to copy/paste.  Subclass, then modify _get_parameters.  If you have an idea for a standard that works for everything, I'm all for it.

mhahn:
@phobologic addressed the above

phobologic:
Awesome.

PR #109, opened 2015-10-13
Remove required_stacks tag

berndtj:
The required_stacks tag could get longer than the AWS limit of 255
chars.  This would cause a stacker build to fail for a largish
deployment.  The required_stacks tag was unused, and all operations
appear to still work.

Testing Done:
- ran unittests
- deployed stack
- destroyed stack

### DIFF
--- a/stacker/actions/build.py
+++ b/stacker/actions/build.py
@@ -87,11 +87,8 @@ class Action(BaseAction):
         logger.debug("Stack %s required stacks: %s",
                      stack.name, requires)
         tags = {
-            'template_url': template_url,
             'stacker_namespace': self.context.namespace,
         }
-        if requires:
-            tags['required_stacks'] = ':'.join(requires)
         return tags
 
     def _launch_stack(self, stack, **kwargs):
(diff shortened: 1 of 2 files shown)


stacker/actions/build.py
```diff
@@ -90,8 +90,6 @@ def _build_stack_tags(self, stack, template_url):
             'template_url': template_url,
```
phobologic: Lets go ahead and remove template_url as well, I don't believe it's used anywhere, and it ends up being misleading.

berndtj:
Removed the template_url as well from tags.  Not sure how I messed up the commits, but the diff is right.

stacker/actions/build.py
```diff
@@ -87,11 +87,8 @@ def _build_stack_tags(self, stack, template_url):
         logger.debug("Stack %s required stacks: %s",
```
phobologic: It looks like this, and the line above is no longer used?

berndtj:
Redoing this ugly commit... closing this request.  Will create new one.

PR #110, opened 2015-10-13
Remove required_stacks tag

berndtj:
The required_stacks tag could get longer than the AWS limit of 255
chars.  This would cause a stacker build to fail for a largish
deployment.  The required_stacks tag was unused, and all operations
appear to still work.

Testing Done:
- ran unittests
- deployed stack
- destroyed stack


phobologic:
Awesome.  Thanks!

Issue #111, opened 2015-10-19
Excessive Memory Consumption


gcallaghan:
As my stacker config grows, the memory consumption has grown significantly. I have a vagrant development environment that is running out of memory when attempting to deploy a stack. I've had to increase my vagrant box memory to 4gb in order to handle a stacker deployment with 17 stacks.

phobologic:
A little more info:

- Does the memory consumption happen all at once, or as stacker runs over time?
- Can I get a little more info about the stacks you're building?  Are you using the blueprints that come with stacker, or are they custom blueprints?  How big are the resulting templates?
- How much memory is in use when you run into these issues?
- What arguments are you running stacker with?
- Does this happen both on first deploy and future updates?

dramdass:
* Does the memory consumption happen all at once, or as stacker runs over time? 
 * All at once after it says "Deploying! Executing Stacker..."
* Can I get a little more info about the stacks you're building? Are you using the blueprints that come with stacker, or are they custom blueprints? How big are the resulting templates?
  * We're doing some custom ones. They are 17 stacks ranging from about 2KB to a couple at 17KB
* How much memory is in use when you run into these issues?
 * PID     USER    PR  NI    VIRT    RES      SHR   S  %CPU %MEM     TIME+ COMMAND
 *  10097 vagrant   20   0 1344212 1.240g   3676 R  99.8     63.5            1:18.89 stacker
* What arguments are you running stacker with?
 * stacker build -r us-west-2 <env path> <yaml path>
* Does this happen both on first deploy and future updates?
 * In my case, future updates
* What version of stacker are you using?
 * 0.5.2

phobologic:
BTW, I've confirmed this, and I've found the issue.  I'm going to be out of town starting tomorrow (Mexico) so won't be able to get back to this till Wednesday, but I've done a bunch of research and attempted a few fixes (that successfully fixed the memory usage, but broke the system in other ways).  Anyway, I'm going to put some notes up here to help me when I can get back to this, or to help anyone who might want to look at it.

So - here's some relevant lines of code:

- Here we create each stack from the Context - passing the context in:
https://github.com/remind101/stacker/blob/master/stacker/context.py#L72

- Here is the Stack object, and you can see the fun comment above the line of code:
https://github.com/remind101/stacker/blob/master/stacker/stack.py#L68

[...]

PR #114, opened 2015-11-03
Remove storage of stacks on context object

phobologic:
Fixes GH-111

This will be, in theory, less efficient since every time you call
get_stacks it will have to rebuild the Stack objets, but in the end it
saves the nasty memory issue we have, and it's simple.

If it ever becomes a problem we could store the stack objects in a
STACKS global or something similar, but that seems gross, and the type
of pre-optimization that'll bite you in the ass (like this did).


mhahn:
+1

PR #115, opened 2015-11-04
Adt add enabled flag v2

synfinatic:
Would be nice to use a single yaml file for multiple envs and pick certain stacks to deploy on a per-env basis.

### DIFF
--- a/stacker/actions/build.py
+++ b/stacker/actions/build.py
@@ -28,6 +28,22 @@ def should_update(stack):
     return True
 
 
+def should_submit(stack):
+    """Tests whether a stack should be submitted to CF for update/create
+
+    Args:
+        stack (stacker.stack.Stack): The stack object to check.
+
+    Returns:
+        bool: If the stack should be updated, return True.
+    """
+    if stack.enabled:
+        return True
+
+    logger.info("Stack %s is not enabled.  Skipping.", stack.name)
+    return False
+
+
 class Action(BaseAction):
     """Responsible for building & coordinating CloudFormation stacks.
 
@@ -122,8 +138,7 @@ class Action(BaseAction):
                                                      required_params,
                                                      provider_stack)
 
-        if not stack.enabled:
-            logger.info("Stack %s is not enabled.  Skipping.", stack.name)
+        if not should_submit(stack):
             return SKIPPED
 
         try:
--- a/stacker/actions/build.py
+++ b/stacker/actions/build.py
@@ -12,7 +12,7 @@ def should_update(stack):
     """Tests whether a stack should be submitted for updates to CF.
 
     Args:
-        stack (stacker.stack.Stack): The stack object to check.
+        stack (:class:`stacker.stack.Stack`): The stack object to check.
 
     Returns:
         bool: If the stack should be updated, return True.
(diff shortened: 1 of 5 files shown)


stacker/actions/build.py
```diff
@@ -122,6 +138,9 @@ def _launch_stack(self, stack, **kwargs):
                                                      required_params,
                                                      provider_stack)
 
+        if not should_submit(stack):
```
phobologic: This seems a little late to test for this - any reason not to do it right at the beginning, since you already have the stack?

phobologic:
Cool!  This is a feature we've been talking about here as well.  I'm actually curious how this works with other stack actions - like destroy, etc?  Will it lead to issues with those?

Also - could you add the enabled flag (probably set to true) with a comment about what it means to https://github.com/remind101/stacker/blob/master/conf/example.yaml - we've used that as a source of documentation for the most part, till we have actual documentation.  I'd just set it to true for each stack, and mention that that is the default.

Thanks!

mhahn:
+1 i've been using this locally for a couple months now, i forgot to submit a PR for it. I called it `ignore` but its functionally the same thing. 

re: destroy, in my case if its ignored, it wasn't built, so shouldn't be destroyed. i think that works the same with this.

phobologic:
Ok - so if `destroy` gets a stack that doesn't exist, it just ignores it currently?

Should there be a CLI arg to ignore this flag?  Probably not - not sure there's any reason you'd want to do that, this seems like something you'd mostly use in a dev stack.

synfinatic:
Yep, if you call destroy on a stack which isn't there, it skips it and moves on.  I'd argue against a CLI flag as that is more knobs to confuse the user.

stacker/actions/build.py
```diff
@@ -28,6 +28,22 @@ def should_update(stack):
...
 
 
+def should_submit(stack):
+    """Tests whether a stack should be submitted to CF for update/create
+
+    Args:
+        stack (stacker.stack.Stack): The stack object to check.
+
+    Returns:
+        bool: If the stack should be updated, return True.
```
phobologic: nit: updated -> submitted

stacker/actions/build.py
```diff
@@ -28,6 +28,22 @@ def should_update(stack):
     return True
 
 
+def should_submit(stack):
+    """Tests whether a stack should be submitted to CF for update/create
+
+    Args:
+        stack (stacker.stack.Stack): The stack object to check.
```
phobologic: Can you use the syntax found here for the class - it looks better in the generated docs: https://github.com/remind101/stacker/blob/master/stacker/stack.py#L50

stacker/stack.py
```diff
@@ -54,10 +54,11 @@ class Stack(object):
             blueprint.
         locked (Optional(bool)): Whether or not the stack is locked.
         force (Optional(bool)): Whether to force updates on this stack.
+        enabled (Optional(bool)): Wether this stack is enabled
```
phobologic: Wether - Whether

synfinatic:
Anything else?

phobologic:
Looks good - thanks!

PR #117, opened 2015-11-16
Adds support to Stacker templates for AWS Cloudformation List<> types.

synfinatic:
Adds support to Stacker templates for the AWS Cloudformation List<> types.  Now when you want to launch an instance with multiple SecurityGroups, instead of specifying multiple parameters you can specify a single List<AWS::EC2::SecurityGroup::Id> parameter.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## 0.5.4 (???)
+
+- Add support for List<AWS::EC2::*> parameters
+
 ## 0.5.3 (2015-11-03)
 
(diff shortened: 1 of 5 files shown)


phobologic:
Cool idea - I really like this feature.  I want to do another run through of it (later today), but can you also update the doc strings where they reference the output lookups here?

synfinatic:
I suppose I can update _resolve_parameters()... was there any other docstrings you had in mind?

phobologic:
Looks good!  Thanks!

PR #120, opened 2015-11-25
Parameters, if blank, will now not be submitted - falling back to the Parameter defaults

phobologic:
There's (almost?) never a good reason to pass 'None' as a Parameter to a
cfn template, so in the case that None is passed in (usually due to a
blank parameter in the config/env), change it to a blank string. I think
this is almost always what is expected.

This also makes a function called resolve_parameters, pulling it out of
the build.Action class, so it is easier to test.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -74,6 +74,14 @@ When updating an existing stack, if you don't supply a parameter in either the
 config or CLI, it will fall back on checking the existing stack for the
 parameter. If it finds it, it will use that automatically.
 
+One additional bit of functionality that should be noted is that if a Parameter
+is defined, but has no value, for example in the stack config like this::
+
+    MyParameter:
+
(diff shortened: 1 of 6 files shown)


phobologic:
So from the latest commit, I've had a change of heart on this, and I think the new method makes more sense:

```
Blank parameters now result in removing the param
I think this actually makes more sense - it allows us to fall back to
defaults for Parameters a lot easier by just not specifying a value.
Ie:

Param:

^^ Nothing after the Param, which is valid yaml, results in that Param
not being submitted to the blueprint, which means it'll fall back on
it's default.  If we want blank strings, people should be explicit about
that, ie:

Param: !!str
or
Param: ""
```

mhahn:
+1 makes sense to me. i haven't come across a use case for not passing anything as a param

PR #126, opened 2015-12-22
KMS Master Key encryption translator

phobologic:


mhahn:
+1

mhahn:
i'm most likely going to switch to this from vault, not sure if we should then remove the vault translator or keep it around as an implementation example. i don't really like leaving around stuff i don't use

PR #127, opened 2015-12-30
Fixed typo mixing up Launch Config and ASG

jonatanblue:
Small fix. It would be good to write up some tests for this, but perhaps better to do in stacker_blueprints.


phobologic:
Thanks @jordansmith108 - yeah, this definitely needs to be moved over to the blueprints module.

PR #128, opened 2016-01-04
Support storing the kms encrypted value in a file

mhahn:


stacker/config/translators/kms.py
```diff
@@ -23,9 +23,24 @@ def kms_simple_decrypt(value):
...
+
+        and reference it within stacker:
+
+        conf_key: !kms file://kms_value.txt
+
+        # Both of the above would resolve to
         conf_key: PASSWORD
+
     """
+    if value.startswith('file://'):
```
phobologic: What do you think about making this more generic so that it's used by every translator?

Also is it worth using urlparse instead of the raw split you do here?  You could parse it with urlparse, and if there is no scheme, then just use the raw value?  Maybe it's not worth it at this point.

stacker/config/translators/kms.py
```diff
@@ -23,9 +23,24 @@ def kms_simple_decrypt(value):
...
 
-        # The above would resolve to
+        You can optionally store the encrypted value in a file, ie:
+
+        kms_value.txt
+        us-east-1@CiD6bC8t2Y<...encrypted blob...>
+
+        and reference it within stacker:
+
+        conf_key: !kms file://kms_value.txt
```
mhahn: the one confusing thing here is this path is relative to where you run the stacker function -- not where the conf file is located.

i'm not sure we can easily get the location of the conf file here

stacker/config/translators/kms.py
mhahn: i refactored this to a function that could potentially be used by other translators. i'd wait on making it more complex until we need to

mhahn:
not sure if we should cache the result of `_get_config_directory` since we have argparse open the file and parse the env file, it could have some overhead if you have multiple kms values

stacker/config/translators/kms.py
```diff
@@ -1,7 +1,27 @@
 import base64
+import os
+
 import botocore.session
 
 
+def _get_config_directory():
```
phobologic: Both of these helper functions seem useful/generic enough that I think they should probably be in the base translators module - I could see other translators, in the future, wanting to use them.

phobologic:
While we don't have full docs, this feature is useful enough that I think it might be worth mentioning the use of file:// as well as just putting a string in the field in the section of the README.rst that discusses translators.

stacker/config/translators/kms.py
mhahn: makes sense

phobologic:
:+1:

PR #130, opened 2016-01-05
Fixing incorrect behaviour of --force flag when using destroy command

oliviervg1:
The user is asked to confirm stack deletions when the force flag is set. This should be the other way around.

### DIFF
--- a/stacker/commands/stacker/destroy.py
+++ b/stacker/commands/stacker/destroy.py
@@ -31,7 +31,7 @@ class Destroy(BaseCommand):
             '\n'.join(stack_names),
         )
         force = False
-        if options.force:
+        if not options.force:
             confirm = raw_input(message)
             force = confirm.lower() == 'yes'
             if not force:


phobologic:
I actually think the original behavior was intended on the part of @mhahn - going to wait for him to comment/+1 this before pulling it in.  Basically I think he wanted the `destroy` command to require a lot of authorization.

oliviervg1:
@phobologic I can see the rationale behind that. However, in that case I'd suggest renaming the flag to something more appropriate such as "--dry-run=(false | true)".

It would also still be nice to be able to have an actual "--force" flag which doesn't require human input. You can always protect a stack with the Stacker lock feature or AWS stack policies if you're scared of accidentally deleting a stack.

mhahn:
yea, i was being overly cautious.

i think we should adjust the logic to:

`stacker destroy <environment> <config>`
- prints out the delete plan and asks you to enter the name of the namespace

`stacker destroy <environment> <config> --no-input`
- prints out the delete plan and then starts deleting without requiring you to enter the namespace

mhahn:
i simplified the logic here: https://github.com/remind101/stacker/pull/131

phobologic:
Closing this for #131

PR #132, opened 2016-01-12
Add support for diffing existing vs. new CF Templates

synfinatic:
Sometimes it's hard to figure out what is different between your
existing CloudFormation deployment and the template changes you've made.

Now you can specify --diff to the 'stacker build' command to see
what has changed in one or more stacks and get a feel for what
changes AWS will make when you deploy the changes for real.

### DIFF
--- a/stacker/actions/build.py
+++ b/stacker/actions/build.py
@@ -4,8 +4,13 @@ from .base import BaseAction
 from .. import exceptions, util
 from ..exceptions import StackDidNotChange
 from ..plan import COMPLETE, SKIPPED, SUBMITTED, Plan
+import json
+import os
+import subprocess
+import tempfile
 
 logger = logging.getLogger(__name__)
+DIFF_ARGS="-C3"
 
 
 def should_update(stack):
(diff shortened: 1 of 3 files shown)


phobologic:
This is awesome - haven't gone through the diff yet, but I have actually started work on this before.  The only thing I think is probably worth being different (which I think is illustrated by how many pieces of code you had to change for this) is that I think this should probably be its own subcommand, not a flag on build.

stacker/actions/build.py
```diff
@@ -228,17 +233,76 @@ def _handle_missing_parameters(self, params, required_params,
...
+            new = tempfile.NamedTemporaryFile(
+                prefix="new_stack_",
+                delete=False,
+            )
+            try:
+                old.write(old_stack)
+                new.write(new_stack)
+                old.close()
+                new.close()
+                pobj = subprocess.Popen(
```
phobologic: If you use difflib here, you won't need all the temporary files, etc.

stacker/actions/build.py
```diff
@@ -228,17 +233,76 @@ def _handle_missing_parameters(self, params, required_params,
...
+                    ['/usr/bin/diff', DIFF_ARGS, old.name, new.name],
```
phobologic: One other thing is that this assumes that diff will be at this place, which is another reason to use difflib rather than shelling out to another command.

synfinatic:
going to abandon this based on above comments

PR #133, opened 2016-01-14
Add support for diffing existing vs. modified CF templates (redux)

synfinatic:
Sometimes it's hard to figure out what is different between your
existing CloudFormation deployment and the template changes you've made.

Now you can specify the 'stacker diff' command to see
what has changed in one or more stacks and get a feel for what
changes AWS will make when you deploy the changes for real.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -24,8 +24,14 @@ Stacker Command
 ===============
 
 The stacker command is built to have sub-commands, much like git. Currently the
-only implemented command is ``build``, which handles taking your stack config
-and then launching or updating stacks as necessary.
+comands are:
+
+- ``build`` which handles taking your stack config and then launching or 
+  updating stacks as necessary.
+- ``destroy`` which tears down your stacks
+- ``diff`` which compares your currently deployed stack templates to your 
+  config files 
+- ``info`` which prints information about your currently deployed stacks
 
 Example
 =======
--- a/README.rst
+++ b/README.rst
@@ -26,11 +26,11 @@ Stacker Command
 The stacker command is built to have sub-commands, much like git. Currently the
 comands are:
 
-- ``build`` which handles taking your stack config and then launching or 
+- ``build`` which handles taking your stack config and then launching or
   updating stacks as necessary.
 - ``destroy`` which tears down your stacks
-- ``diff`` which compares your currently deployed stack templates to your 
-  config files 
+- ``diff`` which compares your currently deployed stack templates to your
+  config files
 - ``info`` which prints information about your currently deployed stacks
 
 Example
--- a/stacker/actions/diff.py
+++ b/stacker/actions/diff.py
@@ -5,9 +5,53 @@ from ..plan import COMPLETE, SKIPPED, Plan
 from . import build
 import difflib
 import json
-import sys
 
 logger = logging.getLogger(__name__)
+
+
+def diff_dictionaries(old_dict, new_dict):
+    """Diffs two single dimension dictionaries
+
+    Returns the number of changes and an unordered list
+    expressing the common entries and changes.
+
+    Args:
+        old_dict(dict): old dictionary
+        new_dict(dict): new dictionary
+
+    Returns: list()
+        int: number of changed records
+        list: [str(<change type>), <key>, <value>]
+
+        Where <change type>: +, - or empty string
+    """
+
+    old_set = set(old_dict)
+    new_set = set(new_dict)
+
+    added_set = new_set - old_set
+    removed_set = old_set - new_set
+    common_set = old_set & new_set
+
+    changes = 0
+    output = []
+    for key in added_set:
+        changes += 1
+        output.append(['+', key, new_dict[key]])
+
+    for key in removed_set:
+        changes += 1
+        output.append(['-', key, old_dict[key]])
+
+    for key in common_set:
+        if str(old_dict[key]) != str(new_dict[key]):
+            changes += 1
+            output.append(['-', key, old_dict[key]])
+            output.append(['+', key, new_dict[key]])
+        else:
+            output.append(['', key, new_dict[key]])
+
+    return [changes, output]
 
 
 class Action(build.Action):
@@ -23,32 +67,24 @@ class Action(build.Action):
     """
 
     def _diff_parameters(self, old_params, new_params):
-        """Compares the old vs. new parameters and returns a 'diff'
+        """Compares the old vs. new parameters and prints a 'diff'
+
+        If there are no changes, we print nothing.
 
         Args:
             old_params(dict): old paramters
             new_params(dict): new parameters
-
-        Returns:
-            int: number of changed entries
-            string: "diff" of all parameters and how they changed
         """
-        output = []
-        count = 0
-        for k, v in old_params.iteritems():
-            new_val = new_params.get(k, "_Undef_")
-            if str(v) != str(new_val):
-                output.append("CHANGED   %s: %s => %s\n" % (k, v, new_val))
-                count += 1
-            else:
-                output.append("NO CHANGE %s: %s\n" % (k, v))
+        [changes, diff] = diff_dictionaries(old_params, new_params)
+        if changes == 0:
+            return
 
-        for k, v in new_params.iteritems():
-            if not k in old_params:
-                output.append("NEW       %s: %s\n" % (k, v))
-                count += 1
+        print """*** New Parameters
+--- Old Parameters
+******************"""
 
-        return [count, "".join(output)]
+        for line in diff:
+            print "%s\t%s = %s" % (line[0], line[1], line[2])
 
     def _normalize_json(self, template):
         """Normalizes our template for diffing
@@ -67,6 +103,32 @@ class Action(build.Action):
             result.append(line + "\n")
         return result
 
+    def _print_new_stack(self, stack, parameters):
+        """Prints out the parameters & stack contents of a new stack"""
+        print "New template parameters:"
+        for key in parameters.keys().sort():
+            print "%s = %s" % (key, parameters[key])
+
+        print "New template contents:"
+        print "".join(stack)
+
+    def _print_stack_changes(self, stack_name, new_stack, old_stack,
+                             new_params, old_params):
+        """Prints out the paramters (if changed) and stack diff"""
+        from_file = "old_%s" % (stack_name,)
+        to_file = "new_%s" % (stack_name,)
+        lines = difflib.context_diff(
+            old_stack, new_stack,
+            fromfile=from_file, tofile=to_file)
+
+        # lines is a generator, not a list hence this is a bit fugly
+        template_changes = list(lines)
+        if not template_changes:
+            print "*** No changes to template ***"
+        else:
+            self._diff_parameters(old_params, new_params)
+            print "".join(template_changes)
+
     def _diff_stack(self, stack, **kwargs):
         """Handles the diffing a stack in CloudFormation vs our config"""
         if not build.should_submit(stack) or not build.should_update(stack):
@@ -82,48 +144,21 @@ class Action(build.Action):
 
         # generate our own template & params
         new_template = stack.blueprint.rendered
-        resolved_parameters = self._resolve_parameters(stack.parameters,
-                                                       stack.blueprint)
-        required_params = [k for k, v in stack.blueprint.required_parameters]
-        parameters = self._handle_missing_parameters(resolved_parameters,
-                                                     required_params)
+        parameters = self.build_parameters(stack)
         new_params = dict()
         for p in parameters:
             new_params[p[0]] = p[1]
         new_stack = self._normalize_json(new_template)
-        [param_changes, param_diff] = self._diff_parameters(
-            old_params, new_params)
 
-        print "============== Stack: %s ==============" % (stack,)
+        print "============== Stack: %s ==============" % (stack.name,)
+        # If this is a completely new template dump our params & stack
         if not old_template:
-            print "Input parameters:"
-            print param_diff
-            print "New template contents:"
-            sys.stdout.write(''.join(new_stack))
+            self._print_new_stack(new_stack, parameters)
         else:
+            # Diff our old & new stack/parameters
             old_stack = self._normalize_json(old_template)
-
-            lines = difflib.context_diff(
-                old_stack, new_stack,
-                fromfile="old_stack", tofile="new_stack")
-
-            count = 0
-            # lines is a generator, not a list hence this is a bit fugly
-            for line in lines:
-                if count == 0:
-                    print "Input parameters:"
-                    print param_diff
-
-                sys.stdout.write(line)
-                count += 1
-
-            if count == 0:
-                print "*** No changes to template ***"
-                return SKIPPED
-            elif param_changes:
-                print "Input parameters:"
-                print param_diff
-
+            self._print_stack_changes(stack.name, new_stack, old_stack,
+                                      new_params, old_params)
         return COMPLETE
 
     def _generate_plan(self):
--- a/stacker/actions/diff.py
+++ b/stacker/actions/diff.py
@@ -23,7 +23,7 @@ def diff_dictionaries(old_dict, new_dict):
         int: number of changed records
         list: [str(<change type>), <key>, <value>]
 
-        Where <change type>: +, - or empty string
+        Where <change type>: +, - or <space>
     """
 
     old_set = set(old_dict)
@@ -49,7 +49,7 @@ def diff_dictionaries(old_dict, new_dict):
             output.append(['-', key, old_dict[key]])
             output.append(['+', key, new_dict[key]])
         else:
-            output.append(['', key, new_dict[key]])
+            output.append([' ', key, new_dict[key]])
 
     return [changes, output]
 
@@ -79,12 +79,12 @@ class Action(build.Action):
         if changes == 0:
             return
 
-        print """*** New Parameters
---- Old Parameters
+        print """--- Old Parameters
++++ New Parameters
 ******************"""
 
         for line in diff:
-            print "%s\t%s = %s" % (line[0], line[1], line[2])
+            print "%s%s = %s" % (line[0], line[1], line[2])
 
     def _normalize_json(self, template):
         """Normalizes our template for diffing
@@ -121,7 +121,6 @@ class Action(build.Action):
             old_stack, new_stack,
             fromfile=from_file, tofile=to_file)
 
-        # lines is a generator, not a list hence this is a bit fugly
         template_changes = list(lines)
         if not template_changes:
             print "*** No changes to template ***"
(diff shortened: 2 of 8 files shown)


synfinatic:
One thing this doesn't do (and I can't figure out a good way to solve) is how to diff parameters passed in to a stack which would be nice.... doesn't seem that AWS has an API call to return those values that I can see.

phobologic:
Yeah, this was where I actually got hung up when working on my version of this.  That and the use of Ref() throughout templates makes it hard to say what exactly has changed.  You can actually get the parameters from the output of DescribeStacks: http://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_DescribeStacks.html

README.rst
```diff
@@ -24,8 +24,14 @@ Stacker Command
 ===============
 
 The stacker command is built to have sub-commands, much like git. Currently the
-only implemented command is ``build``, which handles taking your stack config
-and then launching or updating stacks as necessary.
+comands are:
+
+- ``build`` which handles taking your stack config and then launching or
```
phobologic: Thanks for catching this!

stacker/commands/stacker/diff.py
```diff
@@ -0,0 +1,35 @@
...
+from ...actions import diff
+
+
+class Diff(BaseCommand):
+    name = 'diff'
+    description = __doc__
+
+    def add_arguments(self, parser):
+        super(Diff, self).add_arguments(parser)
+        parser.add_argument("--force", action="append", default=[],
```
phobologic: Probably not necessary?

synfinatic: you mean --force?  No, actually it is useful (and necessary) since determining which stacks are processed works just like the build command.

stacker/actions/diff.py
```diff
@@ -0,0 +1,83 @@
...
+        try:
+            old_stack = self.provider.get_template(stack.fqn)
+        except exceptions.StackDoesNotExist:
+            old_stack = None
+
+        new_stack = self._normalize_json(stack.blueprint.rendered)
+        logger.info("============== Stack: %s ==============", stack)
+        if not old_stack:
+            logger.info("New template contents:")
+            sys.stdout.write(''.join(new_stack))
```
phobologic: This seems like it'd be a bit verbose - it's probably worth just flagging it as a totally new stack, don't you think?  I guess I could see use cases for both.  Anyway, I'm fine with going with this for now, just curious on your thoughts.

stacker/actions/diff.py
```diff
@@ -0,0 +1,83 @@
...
+                sys.stdout.write(line)
+                count += 1
+            if not count:
+                print "*** No changes ***"
+                return SKIPPED
+        return COMPLETE
+
+    def _generate_plan(self):
+        plan_kwargs = {}
+        plan = Plan(description='Create/Update stacks', **plan_kwargs)
```
phobologic: Copy/paste leftover

stacker/commands/stacker/diff.py
phobologic: Yeah, just got back to that - it's kind of a weird syntax here. This really makes me wish we had 'reasons' attached to SKIPPED results.

stacker/actions/diff.py
synfinatic: I decided to follow the diff command's lead of showing the new file/contents.  I briefly considered a --verbose option, but decided it wasn't worth the effort.

stacker/commands/stacker/diff.py
synfinatic: Yeah, kinda.  Honestly, I like the fact that diff and build take the same args so once I've checked everything is good in my diff, I just need to edit the command on the cli and hit enter... less chance of making Bob unhappy. :)

synfinatic:
btw, i'm looking at adding the parameters to the diff.

phobologic:
Awesome!

synfinatic:
ping

phobologic:
Thanks for pinging - github doesn't send notifications when someone updates code in a PR, only comments.  I'll try to look at this soon.

stacker/providers/aws.py
```diff
@@ -183,3 +183,27 @@ def get_outputs(self, stack_name, *args, **kwargs):
...
+                parameters[p.key] = p.value
+            ret = stack.get_template()
+            template = ret['GetTemplateResponse']['GetTemplateResult']
+            template = template['TemplateBody']
+            return [template, parameters]
+        except boto.exception.BotoServerError as e:
+            if 'does not exist' not in e.message:
+                raise
+            raise exceptions.StackDoesNotExist(stack_name)
+        except KeyError:
```
phobologic: This KeyError seems pretty broad, I'd narrow it's scope to only the piece of code you're actually checking (I think this is really only supposed to be around the parts dealing with the template?)

stacker/providers/aws.py
```diff
@@ -183,3 +183,27 @@ def get_outputs(self, stack_name, *args, **kwargs):
...
+                kwargs=dict(stack_name_or_id=stack_name))
+            stack = stacks[0]
+            parameters = dict()
+            for p in stack.parameters:
+                parameters[p.key] = p.value
+            ret = stack.get_template()
+            template = ret['GetTemplateResponse']['GetTemplateResult']
+            template = template['TemplateBody']
+            return [template, parameters]
+        except boto.exception.BotoServerError as e:
```
phobologic: This also seems pretty broad - you really only need this around the self.cloudformation.describe_stacks call, right?

stacker/actions/diff.py
```diff
@@ -0,0 +1,110 @@
...
+            for line in lines:
+                sys.stdout.write(line)
+                count += 1
+            if not count:
+                print "*** No changes ***"
+                return SKIPPED
+        return COMPLETE
+
+    def _generate_plan(self):
+        plan_kwargs = {}
```
phobologic: Doesn't seem like this is used really.

stacker/actions/diff.py
```diff
@@ -0,0 +1,110 @@
...
+    Generates the build plan based on stack dependencies (these dependencies
+    are determined automatically based on references to output values from
+    other stacks).
+
+    The plan is then used to pull the current CloudFormation template from
+    AWS and compare it to the generated templated based on the current
+    config.
+    """
+
+    def _normalize_json(self, json_str, parameters):
```
phobologic: `json_str` should probably template at this point, right?  Also, what was the reasoning behind having the parameters & template normalized into a single string, rather than comparing each separately?  This feels a bit complicated and could be refactored out so that the normalization (and comparison, used in `_diff_stack`) could be easier to read/understand.

synfinatic:
new patch

stacker/actions/diff.py
```diff
@@ -0,0 +1,146 @@
...
+
+        Args:
+            old_params(dict): old paramters
+            new_params(dict): new parameters
+
+        Returns:
+            int: number of changed entries
+            string: "diff" of all parameters and how they changed
+        """
+        output = []
```
phobologic: You might be able to make this logic a little easier if you use sets on the keys - an example:

```
>>> old = {'a': 'Apple', 'b': 'Banana', 'c': 'Carrot'}
>>> new = {'b': 'Banana', 'a': 'Apple', 'o': 'Orange', 'l': 'Lettuce'}
>>>
>>> old_set = set(old)
>>> new_set = set(new)
>>>
>>> added = new_set - old_set
>>> removed = old_set - new_set
>>> common = old_set & new_set
>>>
>>> added
set(['l', 'o'])
>>> removed
set(['c'])
>>> common
set(['a', 'b'])
```

Then you can iterate over the common ones to see if the values are actually different or not.  The other thing this gives you is the ability to print the removed values and not rely on a special string like `_Undef_`

stacker/actions/diff.py
```diff
@@ -0,0 +1,146 @@
...
+    Generates the build plan based on stack dependencies (these dependencies
+    are determined automatically based on references to output values from
+    other stacks).
+
+    The plan is then used to pull the current CloudFormation template from
+    AWS and compare it to the generated templated based on the current
+    config.
+    """
+
+    def _diff_parameters(self, old_params, new_params):
```
phobologic: This logic is probably worth writing a test for.  It'll likely be easier to make it testable by breaking it out as a function.  You can see an example of that here:

https://github.com/remind101/stacker/blob/master/stacker/actions/build.py#L115

stacker/actions/diff.py
```diff
@@ -0,0 +1,146 @@
...
+                count += 1
+            else:
+                output.append("NO CHANGE %s: %s\n" % (k, v))
+
+        for k, v in new_params.iteritems():
+            if not k in old_params:
+                output.append("NEW       %s: %s\n" % (k, v))
+                count += 1
+
+        return [count, "".join(output)]
```
phobologic: BTW, testing the output as a string might be harder than if it's another data type.  Maybe a dictionary of added/removed/changed?  Up to you though, I'm fine with this, but a string seems a lot more fragile.

stacker/providers/aws.py
```diff
@@ -183,3 +183,30 @@ def get_outputs(self, stack_name, *args, **kwargs):
...
+            if 'does not exist' not in e.message:
+                raise
+            raise exceptions.StackDoesNotExist(stack_name)
+
+        stack = stacks[0]
+        parameters = dict()
+        for p in stack.parameters:
+            parameters[p.key] = p.value
+        ret = stack.get_template()
+        try:
```
phobologic: Is this actually valid?  The stack already exists - otherwise it would have failed much higher than this, right?

stacker/actions/diff.py
```diff
@@ -0,0 +1,146 @@
...
+            old_stack = self._normalize_json(old_template)
+
+            lines = difflib.context_diff(
+                old_stack, new_stack,
+                fromfile="old_stack", tofile="new_stack")
+
+            count = 0
+            # lines is a generator, not a list hence this is a bit fugly
+            for line in lines:
+                if count == 0:
```
phobologic: If there is no change to teh templates, but there is a change to the parameters, won't this get skipped?  I'm assuming if there is no change to the templates, then 'lines' is an empty generator?

stacker/actions/diff.py
```diff
@@ -0,0 +1,146 @@
...
+        else:
+            old_stack = self._normalize_json(old_template)
+
+            lines = difflib.context_diff(
+                old_stack, new_stack,
+                fromfile="old_stack", tofile="new_stack")
+
+            count = 0
+            # lines is a generator, not a list hence this is a bit fugly
+            for line in lines:
```
phobologic: Maybe this would at least clean it up some:

```
if param_changes:
    print "Input parameters:"
    print param_diff

template_changes = list(lines)
if not template_changes:
    print "*** No changes to stack %s template ***" % (stack.name)
    return SKIPPED

for line in template_changes:
    sys.stdout.write(line)
```

Not sure about the `SKIPPED` line, since I'm not sure what SKIPPED means in the case the parameters changed but the template didn't.  Maybe it doesn't quite apply here.

stacker/actions/diff.py
synfinatic: Not sure I understand?  output is the diff.  I suppose that I could just suppress the "NO CHANGE" lines, but decided I kinda liked seeing all the parameters if the parameters or template itself changed.

phobologic: I'm just saying that if you write tests for this function (like I mentioned above) the output may be harder to test when you have a giant text string to test - you have to make sure all the white space is correct, etc, which doesn't seem like something you'd actually want to test.

mhahn: yea, i would break up the "determine whats changed" vs. "format a message to the user"

stacker/actions/diff.py
```diff
@@ -0,0 +1,146 @@
...
+            old_params(dict): old paramters
+            new_params(dict): new parameters
+
+        Returns:
+            int: number of changed entries
+            string: "diff" of all parameters and how they changed
+        """
+        output = []
+        count = 0
+        for k, v in old_params.iteritems():
```
mhahn: i'm personally not a fan of these variable names. i would go with the more verbose option of:
```
for old_key, new_value in old_params.iteritems():
```

then the logic below, ie:

```
new_value = new_params.get(old_key, "_Undef_")
```

is much clearer.

stacker/actions/diff.py
```diff
@@ -0,0 +1,146 @@
...
+        try:
+            [old_template, old_params] = self.provider.get_stack_info(
+                stack.fqn)
+        except exceptions.StackDoesNotExist:
+            old_template = None
+            old_params = {}
+
+        # generate our own template & params
+        new_template = stack.blueprint.rendered
+        resolved_parameters = self._resolve_parameters(stack.parameters,
```
mhahn: We use the underscore prefix to denote that a method is "private" within the class, meaning there is no guarantee about what the function does or if the function will be there in the future. If this is necessary, I would remove the underscore prefix from these methods in the Build class to make it clear that it has (or can have) outside dependencies.

mhahn: i would also roll these into one method: https://github.com/remind101/stacker/blob/master/stacker/actions/build.py#L173-L178 which you could call here instead of copying over the same workflow which could change in the future

stacker/actions/diff.py
```diff
@@ -0,0 +1,146 @@
...
+                                                     required_params)
+        new_params = dict()
+        for p in parameters:
+            new_params[p[0]] = p[1]
+        new_stack = self._normalize_json(new_template)
+        [param_changes, param_diff] = self._diff_parameters(
+            old_params, new_params)
+
+        print "============== Stack: %s ==============" % (stack,)
+        if not old_template:
```
mhahn: i'm not a fan of large if/else statements. i would prefer:
```
if not old_template:
    self._print_no_change(...)
else:
     self._output_diff()
```

stacker/commands/stacker/diff.py
mhahn: i think the help method explains this well too

mhahn:
sweet! tests please :)

stacker/actions/diff.py
phobologic: Did you mean `for old_key, old_value in old_params.iteritems():` @mhahn ?

mhahn: yes :)

stacker/actions/diff.py
```diff
@@ -0,0 +1,181 @@
...
+    """
+
+    old_set = set(old_dict)
+    new_set = set(new_dict)
+
+    added_set = new_set - old_set
+    removed_set = old_set - new_set
+    common_set = old_set & new_set
+
+    changes = 0
```
phobologic: Is there any reason you need this?  It looks like it's only used to tell if there were any changes, which you should be able to tell with `if output` since output will be False if nothing has changed, right?

synfinatic: I don't return just the changed lines, but rather all the lines since I felt that helped with context.  For example, I have some parameters passed into my stack which are related and updating one may require updating the other.  This allows me to detect not updating them together when necessary.

stacker/actions/diff.py
```diff
@@ -0,0 +1,181 @@
...
+
+        Args:
+            old_params(dict): old paramters
+            new_params(dict): new parameters
+        """
+        [changes, diff] = diff_dictionaries(old_params, new_params)
+        if changes == 0:
+            return
+
+        print """*** New Parameters
```
phobologic: Is this header correct?  I think this is a bit confusing - the output when there was an update seems like it's unnecessarily noisy?  Can you share an example of what it looks like?

synfinatic: Example when changes != 0:

```
*** New Parameters
--- Old Parameters
******************
-       EBSVolumeSize = 0
-       ProfileArn = 
-       EnableEIP = False
-       WavefrontToken = xxxx-xxxx-xxxx-xxxx
-       ConfigStore = 
        SSHAuthorizedKeys = ssh-rsa XXXXXXXXXXXXX
        Subnets = subnet-51194c26,subnet-b15166d4
        ExternalZoneName = demo.aaron.foo.net
        VpcId = vpc-4867472d
        ServiceName = WavefrontProxy
        AzCount = 2
        Namespace = container-dev
        ImageName = wavefront
        SecurityGroups = sg-a14a59c5,sg-f1174095,sg-f0174094
        InternalZoneId = Z32SXJ04VCA6WZ
-       InstanceType = m3.medium
+       InstanceType = m4.medium
        Size = 2
```

phobologic: Gotcha - the header is incorrect right?  `*** New Parameters` - it seems like new parameters should be led by 3 `*`'s.

stacker/actions/diff.py
phobologic: Ahh, ok, seems a little verbose, but I'll wait to use it for a bit before really bothering about it :)

stacker/actions/diff.py
```diff
@@ -0,0 +1,181 @@
...
+        """
+        obj = json.loads(template)
+        json_str = json.dumps(obj, sort_keys=True, indent=4)
+        result = []
+        lines = json_str.split("\n")
+        for line in lines:
+            result.append(line + "\n")
+        return result
+
+    def _print_new_stack(self, stack, parameters):
```
phobologic: This seems really verbose as well - but we can always flag it away later if that ends up being the case.  I could actually see a really cool other flag for this, like `-q` that would just list what stacks have changed, but that can come later.

stacker/actions/diff.py
```diff
@@ -0,0 +1,181 @@
...
+    def _print_stack_changes(self, stack_name, new_stack, old_stack,
+                             new_params, old_params):
+        """Prints out the paramters (if changed) and stack diff"""
+        from_file = "old_%s" % (stack_name,)
+        to_file = "new_%s" % (stack_name,)
+        lines = difflib.context_diff(
+            old_stack, new_stack,
+            fromfile=from_file, tofile=to_file)
+
+        # lines is a generator, not a list hence this is a bit fugly
```
phobologic: can probably ditch this comment

stacker/actions/diff.py
synfinatic: I used the exact same header format that difflib generates for it's diff for the template.  Yes, they're different diff types, but this makes it easier to parse if for some reason someone wanted to do that.

phobologic:
Ok, I'm definitely :+1: on this!  Looks great.  @mhahn, want to give it a :+1: and then I'll merge?  Thanks @synfinatic

stacker/actions/diff.py
phobologic: Yeah - different diff types, and this is a bit confusing since the New Parameters/Old Parameters header is just incorrect in this case.

synfinatic:
fyi, uploaded a small change based on the last set of comments

mhahn:
:+1:

phobologic:
Awesome, thanks @synfinatic!

Issue #135, opened 2016-01-20
Boolean parameters should be converted to strings


phobologic:
`true` in YAML converts to python's `True` boolean. Parameters are never booleans, and the strings `true` and `false` are somewhat common as strings in Parameters.  I can't think of a reason to leave them as the python booleans, as right now to use `true` for example you have to use `!!str true`

PR #136, opened 2016-01-24
Convert python booleans in parameters to strings

phobologic:
Fixes GH-135

Some resources have the concept of boolean Parameters, but these are
booleans in json, which is a lowercase string.  This allows people to
use `true` or `false` in boolean parameters, without having to specify
`!!str` before it in order to force the yaml to not turn them into
python booleans.


brianz:
:+1:

PR #138, opened 2016-02-02
Fix outputting of parameters for new stacks

synfinatic:
After I did the last refactor, parameters is no longer a dict, but rather
a list of tuples which causes a crash when processing new templates.


phobologic:
Thanks!

PR #140, opened 2016-02-04
No reason to always require mappings

phobologic:


phobologic:
Basically before, if you had no mappings defined in your config it'd just break.  Doesn't make sense, you don't need mappings for everything.  this makes it so you don't have to have a mapping

mhahn:
+1

PR #142, opened 2016-02-05
Fix exception: post_build hook takes the region as the 3rd arg

synfinatic:
python doesn't like it when you pass 3 args and 4 are required :)


phobologic:
Thanks!

PR #143, opened 2016-02-08
Add pre/post destroy hooks

synfinatic:
build already has pre/post hooks so you can create non-CloudFormation
resources, but currently Stacker has no way of deleting them which can
prevent destory from being able to tear down one or more stacks due
to dependancies.


phobologic:
Looks good - thanks!

PR #144, opened 2016-02-10
Make sure we don't run post_run & pre_run during diff

synfinatic:
pretty straight forward... diff is a subclass of build which means by default it was running the hooks.


phobologic:
Thanks!  Honestly at some point we should probably make a common master class for `Build` and `Diff`, rather than setting the precident we do here, since dangerous things like this could potentially be the result.  For now though, this should work.

PR #145, opened 2016-03-15
Allow users to override name of bucket to store templates

oliviervg1:
Hi,

We'd like to be able to override the default naming convention used to create the S3 bucket where templates are stored.

This pull request adds a `-b` / `--bucket-name` cli option to pass in the name to use. If the option is omitted then the behaviour remains unchanged.

Thanks!

### DIFF
--- a/stacker/context.py
+++ b/stacker/context.py
@@ -53,10 +53,14 @@ class Context(object):
         self.config = config or {}
         self.force_stacks = force_stacks or []
         self._base_fqn = self.namespace.replace('.', '-').lower()
+        self.bucket_name = 'stacker-%s' % (self.get_fqn(),)
 
     def load_config(self, conf_string):
         self.config = parse_config(conf_string, environment=self.environment)
         self.mappings = self.config.get('mappings', {})
+        bucket_name = self.config.get('stacker_bucket', None)
+        if bucket_name is not None:
+            self.bucket_name = bucket_name
 
     def _get_stack_definitions(self):
         if not self.stack_names:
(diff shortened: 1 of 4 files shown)


phobologic:
Hey @oliviervg1 - thanks for this.  I'm wondering if this would be better as a config option (probably top level) rather than a command line argument.  There are a few benefits and downsides to the config option:

- Folks on a team wouldn't have to make sure they're using the same CLI args (it'd be in the config, that's hopefully shared)
- You could use environment args to have a different bucket for environment, ie: `stacker_bucket: my-bucket-${environment}`
- There's less chance for typos.

The downside is that each person who wants to launch their own set of stacks would need to use a different environment file, though I think that's a good idea in general.

Anyway, what do you think?

oliviervg1:
@phobologic makes sense. I'll update the code when I've got some time :)

phobologic:
Thanks!  Looking forward to it!

oliviervg1:
@phobologic, should be good now :) You were right, the code is much cleaner.

oliviervg1:
@phobologic All done now!

stacker/context.py
```diff
@@ -53,10 +53,14 @@ def __init__(self, environment, stack_names=None,
         self.config = config or {}
         self.force_stacks = force_stacks or []
         self._base_fqn = self.namespace.replace('.', '-').lower()
+        self.bucket_name = 'stacker-%s' % (self.get_fqn(),)
 
     def load_config(self, conf_string):
         self.config = parse_config(conf_string, environment=self.environment)
         self.mappings = self.config.get('mappings', {})
+        bucket_name = self.config.get('stacker_bucket', None)
+        if bucket_name is not None:
```
phobologic: You should be able to use `if bucket_name:` here, and it's a bit more pythonic.

phobologic:
This looks great - can you add something to the README?  Thanks!

phobologic:
Sorry for the delay - this is great, thanks @oliviervg1 !

PR #148, opened 2016-04-04
Add --environment to CLI to pass in environment args on the CLI

synfinatic:
Hey, we can do this with --parameter, but sometimes you want to
pass in environment variables on the CLI.

### DIFF
--- a/stacker/commands/stacker/base.py
+++ b/stacker/commands/stacker/base.py
@@ -103,11 +103,13 @@ class BaseCommand(object):
             level=log_level,
         )
 
-    def parse_args(self):
+    def parse_args(self, *vargs):
         parser = argparse.ArgumentParser(description=self.description)
         self.add_subcommands(parser)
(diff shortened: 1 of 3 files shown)


stacker/commands/stacker/base.py
```diff
@@ -141,6 +143,13 @@ def add_arguments(self, parser):
                                  "that can be used inside any of the stacks "
                                  "being built. Can be specified more than "
                                  "once.")
+        parser.add_argument("-e", "--env", dest="cli_envs",
+                            metavar="ENV=VALUE", type=key_value_arg,
+                            action=KeyValueAction, default={},
+                            help="Adds environment key/value pairs from "
+                            "the command line. Overrides your environment "
```
phobologic: Can you indent this to line up with the quote above?

phobologic:
Cool, thanks!

PR #150, opened 2016-04-25
build now optionally runs (only) the hooks

synfinatic:
pre/post hooks are powerful, but occasionly you need a little more
control over when they run.  This change extends the build command
to support two new flags: --skip-hook and --run-hook which skip
running the pre/post hooks or only run the corresponding hooks.

Testing:
- Manual testing
- Unit tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## x.x.x (2016-xx-xx)
 - Add support for passing environment variables on the cli via --env
+- Add support for build --skip-hook and --run-hook for pre/post_build hooks
 
 ## 0.6.1 (2016-02-11)
 - Add support for the 'stacker diff' command [GH-133]
--- a/stacker/actions/build.py
+++ b/stacker/actions/build.py
@@ -291,6 +291,10 @@ class Action(BaseAction):
             # need to generate a new plan to log since the outline sets the
             # steps to COMPLETE in order to log them
             debug_plan = self._generate_plan()
+
+            # don't actually deploy anything if we're just running the hooks
+            if self.context.run_hook:
+                return
             debug_plan.outline(logging.DEBUG)
             logger.info("Launching stacks: %s", ', '.join(plan.keys()))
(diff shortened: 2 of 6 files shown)


synfinatic:
Ping

mhahn:
Hey, sorry for lagging!

To me, hooks should be idempotent. The thing that makes me nervous with `--skip-hook` and `--run-hook` is that you could start to build up dependencies like:

"the first time you run the stack, execute the hook, but then skip it after that" or some other such thing that is no longer stored in the stacker config file that is reproducible by others.

What is the use case you're trying to solve?

The one thing I've seen with hooks is just that they add overhead.

I use the following hooks:```
pre_build:
  - path: stacker.hooks.keypair.ensure_keypair_exists
    required: true
    args:
      keypair: ${ssh_key_name}
    required: true
    args:
      domain: ${external_domain}
  - path: stacker.hooks.iam.ensure_server_cert_exists
    required: true
    args:
      cert_name: ${empire_controller_cert_name}
```

all of which are mostly useless once i bring up the stacks the first time. i don't mind the performance hit because i don't run stacker often and if i am, i'm expecting it to take forever because of CF. I could see that being a pain with a lot of hooks though.

synfinatic:
The particular use case I'm dealing with is blue/green deployments.  I have a shared "frontend" stack, like an ELB for example and a blue and green EC2 instances in separate stacks.  In our case right now, the EC2 instances can't fully bootstrap themselves so after deploying the stack we have to run some other scripts to boot strap them.

The hooks effectively allow us to migrate the ELB to the blue or green EC2 instances independently from the CF deploy which prevents downtime during the manual bootstrap step.  It also means we can effectively roll back to the previous color without tearing down the existing color.

phobologic:
Yeah, I think this actually encourages some bad patterns - we have something similar at Remind, and rather than writing special run cases of stacker, we wrote a separate tool for doing the out of band management things.  I'm going to go ahead and close this.

PR #153, opened 2016-05-08
Cleanup output on non-verbose runs

phobologic:
Removes a lot of the extraneous logging in non-verbose mode, adds an
immediate checkpoint dump as soon as the first run is complete (showing
you immediately the state of things)

This requires that each action handle it's own output, rather than
relying on the provider.

Eventually (maybe in this PR) I'd like to add color.


ejholmes:
Awesome 👍, especially dumping after the first run.

PR #155, opened 2016-05-17
Add command to diff environments

phobologic:
regular old diff is ok, but when I look at my `dev.env` to see what is
different from it and the checked in `stage.env` it can be a bit noisy,
when usually all I want to see is what keys have been added/removed

This should make that a little easier.

### DIFF
--- a/scripts/compare_env
+++ b/scripts/compare_env
@@ -1,0 +1,69 @@
+#!/usr/bin/env python
+""" A script to compare environment files. """
+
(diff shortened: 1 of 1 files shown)


mhahn:
i think this should be an actual command, like `stacker compare` so i can run it in the directory where my stacks live easily

mhahn:
nvm, i forgot how scripts work in python :)

PR #156, opened 2016-05-17
Support dumping out rendered cloudformation templates for each stack

mhahn:
Output looks like:

```
$ stacker build -r us-west-2 conf/lunohq/development.env conf/lunohq/config.yaml --dump dump
[2016-05-17T14:05:33] Dumping "Create/Update stacks"...
[2016-05-17T14:05:33] Writing stack "dev-lunohq-vpc" -> dump/vpc-3ea2d331.json
[2016-05-17T14:05:33] Writing stack "dev-lunohq-empireDB" -> dump/empireDB-2bc493a4.json
[2016-05-17T14:05:33] Writing stack "dev-lunohq-empireMinion" -> dump/empireMinion-6b3bcb53.json
[2016-05-17T14:05:33] Writing stack "dev-lunohq-empireController" -> dump/empireController-92678c3c.json
...
````

### DIFF
--- a/stacker/actions/build.py
+++ b/stacker/actions/build.py
@@ -297,17 +297,15 @@ class Action(BaseAction):
 
         """
         plan = self._generate_plan(tail=tail)
-        if outline:
-            plan.outline()
-        elif dump:
-            plan.dump(dump)
-        else:
-            # need to generate a new plan to log since the outline sets the
-            # steps to COMPLETE in order to log them
-            debug_plan = self._generate_plan()
-            debug_plan.outline(logging.DEBUG)
+        if not outline and not dump:
+            plan.outline(logging.DEBUG)
             logger.info("Launching stacks: %s", ', '.join(plan.keys()))
             plan.execute()
+        else:
+            if outline:
+                plan.outline()
+            if dump:
+                plan.dump(dump)
 
     def post_run(self, outline=False, *args, **kwargs):
         """Any steps that need to be taken after running the action."""
--- a/stacker/actions/build.py
+++ b/stacker/actions/build.py
@@ -290,22 +290,24 @@ class Action(BaseAction):
             util.handle_hooks('pre_build', pre_build, self.provider.region,
                               self.context)
 
-    def run(self, outline=False, tail=False, *args, **kwargs):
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: stacker/actions/build.py [hunk 1]
```diff
@@ -290,22 +290,24 @@ def pre_run(self, outline=False, *args, **kwargs):
...
         """Kicks off the build/update of the stacks in the stack_definitions.
 
         This is the main entry point for the Builder.
 
         """
         plan = self._generate_plan(tail=tail)
-        if not outline:
+        if outline:
+            plan.outline()
+        elif dump:
```
phobologic: Doesn't seem like these need to be mutually exclusive, right?  The command doesn't give any feedback like that, so might as well allow both.  Maybe ```
if not outline and not dump:
    ... actual build ...
else:
    if outline:
        ... outline ...
    if dump:
        ... dump ...
```

?

File: stacker/plan.py [hunk 2]
```diff
@@ -282,6 +285,24 @@ def outline(self, level=logging.INFO, message=''):
         if message:
             logger.log(level, message)
 
+    def dump(self, output):
```
phobologic: maybe change output to directory?

phobologic: Also, can you include `os.path.expanduser`?

phobologic:
Awesome 👍 Did you get a chance to look at the local parameters 'issue'?

stacker/actions/build.py
mhahn: ah yea, we can do that.

ejholmes:
Sweet 👍 this will be really nice to have.

I was thinking it would be cool if stacker just compiled all of the blueprints down to a single "metastack" template, which could then be executed by stacker itself. If we exposed a subcommand to execute a metastack template, then this could open up other use cases for stacker (people that want the "stack glue" aspect of stacker, but not troposphere/python), and you could just dump the metastack to stdout (e.g. `stacker compile | stacker execute`).

Anyways, definitely not something we should do in this PR, just something I was thinking about.

phobologic:
Honestly for the non-troposphere/python users, I think we should provide a concept of 'drivers'.  You don't need to use a Blueprint in that case, we could have a 'json' driver or something that just reads a regular old CF template and uses that.

PR #162, opened 2016-06-10
Remove extra docs in README

phobologic:
Don't want to maintain these docs in two places, and the docs in docs/
are much better at this point (and readthedocs is much easier to
navigate)


mhahn:
👍

PR #166, opened 2016-06-18
Cleanup various inconsistencies

phobologic:
- > 80 character lines
- mixed use of ' and " (only use " now)
- some minor grammar errors


mhahn:
😍

PR #167, opened 2016-06-27
Allowing users to enable s3 server side encryption when uploading CF templates

oliviervg1:
Some enterprises enforce the use of server side encryption when uploading objects to S3. This change is to allow Stacker to work in such environments.

### DIFF
--- a/stacker/actions/base.py
+++ b/stacker/actions/base.py
@@ -54,7 +54,6 @@ class BaseAction(object):
         self.context = context
         self.provider = provider
         self.bucket_name = context.bucket_name
-        self.bucket_sse = context.bucket_sse
         self._conn = None
         self._cfn_bucket = None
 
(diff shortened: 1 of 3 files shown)


phobologic:
Hey @oliviervg1 - thanks for this!  One question I have: Is there any reason to not make this the default?  I'm kind of embarrassed we didn't make this the default already :)  The encryption is per object, right?

oliviervg1:
Hi @phobologic, yep - the encryption is on a per object basis. I'm not against making it the default behaviour, however I'd just like to double check something first if we go ahead with it:

_Consider we have account A and account B. Stacker uploads server side encrypted templates to a bucket in account A. We then make the bucket available to account B using a **bucket policy**. Can account B read the objects from the object?_

phobologic:
I actually don't know the answer to that question, though I'm curious when you think that situation would arise?

oliviervg1:
I've confirmed that cross account bucket access still works with server side encryption. I'd never recommend anyone to use bucket policies to do cross account S3 access (using IAM role assumption is far better). I just wanted to make sure that we cover potential edge cases and not break anything for people using stacker in weird and wonderful ways :)

I'll update the pull request to force server side encryption if that's ok with you!

phobologic:
Awesome.  Looks good to me.  Thanks @oliviervg1 !

PR #171, opened 2016-07-18
Support tags as a new top-level keyword

### LINKED ISSUE #154: Look into using cloudformation tags again
We removed them with #44 - now CF does provide a method for updating tags, so we should look into whether or not they'd be useful again.

acmcelwee:
Adds a new top-level `tags` keyword. By default, the previously applied `stacker_namespace` tag is still applied to the stack. Alternatively, a user can now specify a different set of tags to apply, including no tags. `config.rst` is updated with the usage examples, including tags based on dynamic environment information.

Might resolve #154

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -71,6 +71,28 @@ the build action::
       args:
         domain: mydomain.com
 
+Tags
+----
+
+Cloudformation supports arbitrary key-value pair tags. All stack-level, including automatically created tags, are
+propagated to resources that AWS CloudFormation supports. See `AWS Cloudformation Resource Tags Type`_ for more details.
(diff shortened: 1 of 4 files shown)


phobologic:
Awesome!  Thanks @acmcelwee - I'm going to open an enhancement request (after closing #154) to add the ability to add tags to specific stacks as well as global tags like you've done here.  Thanks again for this!

acmcelwee:
Yeah, I'd considered adding that, but we didn't have a use case for it. Don't be surprised if you see a PR from me in the next few days, though 😄.

phobologic:
Awesome - thanks! :)

PR #181, opened 2016-08-02
update args comment to reflect current syntax

troyready:
Simple comment update -- should prevent confusion and make it easier to copy/paste

### DIFF
--- a/stacker/tests/fixtures/vpc-bastion-db-web.yaml
+++ b/stacker/tests/fixtures/vpc-bastion-db-web.yaml
@@ -5,8 +5,9 @@
 pre_build:
   - path: stacker.hooks.route53.create_domain
     required: true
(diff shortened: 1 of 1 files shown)


phobologic:
Not sure I understand what the change is here - it looks like its different now than what the comment suggests?

troyready:
As is, it looks like you could just drop the last `#` and do:
```
     required: true
     # Additional args can be passed as a dict of key/value pairs in kwargs
     kwargs:
       foo: bar
```
I want to make it more clear that it should be
```
     required: true
     # Additional args can be passed as a dict of key/value pairs in kwargs
     args:
       foo: bar
```
(dropping the kw)

troyready:
@phobologic sure thing; should be more clear now

phobologic:
Awesome, thanks a ton!

PR #183, opened 2016-08-03
support mixed parameters referencing / not referencing other stacks

troyready:
Currently parameters can reference multiple outputs in other stacks:
```
SomeParameter: stack1::Output1,stack2::Output2
```

This commit allows parameter lists to also include variables not
referencing another stack, e.g.:
```
SomeParameter: ${ThingInEnvFile},stack1::Output1
```

I considered updating the documentation in [Using Outputs as Parameters](https://github.com/remind101/stacker/blob/master/docs/config.rst#using-outputs-as-parameters), but I'm not sure how to change the last part ("You can also pull multiple...") to note the functionality. Maybe a new section? Maybe the new functionality doesn't need to be called out?

### DIFF
--- a/stacker/actions/build.py
+++ b/stacker/actions/build.py
@@ -83,11 +83,16 @@ def resolve_parameters(parameters, blueprint, context, provider):
                          blueprint.name, k)
             continue
         value = v
-        if isinstance(value, basestring) and "::" in value:
-            # Get from the Output(s) of another stack(s) in the stack_map
-            v_list = []
+        # v_list holds each value in the (potentially) comma-delimited list of
+        # values for each parameter
+        v_list = []
+        if isinstance(value, basestring):
             values = value.split(",")
-            for v in values:
+        else:
+            values = [value]
+        for v in values:
+            if isinstance(v, basestring) and "::" in v:
+                # Get from the Output(s) of another stack(s) in the stack_map
                 v = v.strip()
                 stack_name, output = v.split("::")
                 stack_fqn = context.get_fqn(stack_name)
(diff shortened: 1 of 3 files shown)


stacker/actions/build.py
```diff
@@ -89,6 +89,10 @@ def resolve_parameters(parameters, blueprint, context, provider):
             values = value.split(",")
             for v in values:
                 v = v.strip()
+                # Skip parameters not referencing another stack
```
phobologic: It feels like this logic is in the wrong place - honestly because the original logic was in the wrong place.  We should probably pull the for loop that iterates over each part of the Parameter outside of the logic that checks for `::`.  SOmething like:

```
v_list = []
values = value.split(",")
for v in values:
  ... handle regular values, output reference values, etc ...
```

does that make sense?  It'd make it easier, later, to add other special values other than just the `::` values as well.

phobologic: I also wonder if the whole logic of how we parse the values should be pulled out into it's own function, so we don't have to update it in two places like you had to here.

acmcelwee: I agree. Consolidating that would probably pave the way for the work that's being proposed in #184.

troyready:
@phobologic good thoughts, I agree. I've added another commit to rework the logic accordingly.

I tried to take a stab at consolidating the methods in the 2 files, but gave up for now -- since one of them is only checking for requirements and the other is actually getting stack outputs, I wasn't sure the best way to do it. Open to doing more on it depending on your thoughts.

Also, the circleci test is still failing but it doesn't look like it's due to this PR?

phobologic:
Hey @troyready, can you try merging master?  It's been a while, and there have been fixing changes that might be included in master already that could clear up the Circle errors.  So far the code looks good though, I wouldn't owrry about consolidating the methods, I agree with your assessment there.

Thanks!

troyready:
@phobologic thanks for the tip -- I didn't realize it had fallen so behind. I've squashed and rebased it, and everything is looking green now.

troyready:
@phobologic my cleanup yesterday had introduced a breaking change, sorry about that. I've cleaned that up now and added a test to cover it.

stacker/actions/build.py
```diff
@@ -96,17 +101,19 @@ def resolve_parameters(parameters, blueprint, context, provider):
...
+                             "it to cloudformation, default value should be "
+                             "used.", k)
+                continue
+            elif isinstance(v, bool):
+                logger.debug("Converting parameter %s boolean \"%s\" to "
+                             "string.", k, v)
+                v_list.append(str(v).lower())
+            else:
+                v_list.append(v)
+        if v_list != []:
```
phobologic: You can get rid of this if statement - if the list is empty, there will be nothing to join, and it'll result in a blank string.  Only updating the param if the list is populated is a change from the old behavior, and I'm not sure what issues it might cause.

troyready: @phobologic I think the `if` does match the old behavior, specifically the way `None`s have [been handled here](https://github.com/remind101/stacker/blob/02a5ad5c202be1ec8a4247a2984af599db9e17ef/stacker/actions/build.py#L104).

If a value is always returned, then it trips [test_resolve_parameters_none_conversion](https://github.com/remind101/stacker/blob/02a5ad5c202be1ec8a4247a2984af599db9e17ef/stacker/tests/actions/test_build.py#L267-L278).

troyready:
Commented more specifically on the line comment, but as a broad update just wanted to note that I've rebased on top of master and this should be merge-able now. I can squash the 2nd commit with the new test into a single commit if that'd be better.

mhahn:
Hey @troyready, I think the new lookups Variables and Lookups code (https://github.com/remind101/stacker/pull/194) we implemented should resolve your issue without this PR.

The new `variables` section will support:

```
SomeParameter: ${env_value},${stack1::Output1}
 ```

Let me know if that isn't true, closing this for now.

Issue #170, opened 2016-07-15
Reduce verbosity of UI


ejholmes:
https://github.com/remind101/stacker/pull/153 helped a lot, but I think there's still room to make the UI around progress updates more intuitive. I'm thinking, something similar to what `docker pull` does when it pulls layers. Each layer is rendered on a single line, and it's updated as it changes state.

Doing something similar within stacker might make it a little more pleasant to watch:

https://asciinema.org/a/d5vtkmdv5llrxk9jqz66xuzso

Generated with this script:

```bash
#!/bin/bash

count=7

clr() {
  echo -e "\033[${count}A"
}

active() {
  echo -e "\033[K\033[1m\033[33m$1\033[0m"
}

inactive() {
  echo -e "\033[K\033[1m$1\033[0m"
}

complete() {
  echo -e "\033[K\033[2m\033[32m$1\033[0m"
}

inactive "vpc:               Pending"
inactive "bastion:           Pending"
inactive "db:                Pending"
inactive "empire_controller: Pending"
inactive "empire_minion:     Pending"
inactive "empire_daemon:     Pending"

sleep 5

clr
active   "vpc:               Updating..."
inactive "bastion:           Waiting on vpc"
inactive "db:                Waiting on vpc"
inactive "empire_controller: Waiting on vpc"
inactive "empire_minion:     Waiting on vpc"
inactive "empire_daemon:     Waiting on db"

sleep 3

clr
complete "vpc:               Updated"
active   "bastion:           Updating..."
active   "db:                Creating..."
active   "empire_controller: Updating..."
active   "empire_minion:     Updating..."
inactive "empire_daemon:     Waiting on db"

sleep 5

clr
complete "vpc:               Updated"
complete "bastion:           Updated"
active   "db:                Creating..."
complete "empire_controller: Updated"
active   "empire_minion:     Updating..."
inactive "empire_daemon:     Waiting on db"

sleep 5

clr
complete "vpc:               Updated"
complete "bastion:           Updated"
complete "db:                Created"
complete "empire_controller: Updated"
active   "empire_minion:     Updating..."
active   "empire_daemon:     Creating..."

sleep 3

clr
complete "vpc:               Updated"
complete "bastion:           Updated"
complete "db:                Created"
complete "empire_controller: Updated"
complete "empire_minion:     Updated"
active   "empire_daemon:     Creating..."

sleep 10

clr
complete "vpc:               Updated"
complete "bastion:           Updated"
complete "db:                Created"
[...]

PR #190, opened 2016-08-04
Reduce the verbosity of stacker output

mhahn:
This leverages colorama and a custom log handler to continually update the same lines when outputting the plan status.

Output looks like:

image

and will continue to update the same lines as the stack updates, ie:

image

Fixes: https://github.com/remind101/stacker/issues/170

### DIFF
--- a/stacker/logger/handler.py
+++ b/stacker/logger/handler.py
@@ -1,0 +1,35 @@
+from logging import StreamHandler
+from colorama.ansi import (
+    Cursor,
+    clear_line,
+)
+
+
+class LogLoopStreamHandler(StreamHandler):
+
+    def __init__(self, *args, **kwargs):
+        super(LogLoopStreamHandler, self).__init__(*args, **kwargs)
+        self.loops = {}
+
+    def format(self, record):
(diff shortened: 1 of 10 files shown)


stacker/logger/handler.py
```diff
@@ -0,0 +1,39 @@
+from logging import StreamHandler
+from colorama.ansi import (
+    Cursor,
+    clear_line,
+)
+
+
+class LogLoopStreamHandler(StreamHandler):
+
```
phobologic: docstring would be nice here just to explain what this is

stacker/plan.py
```diff
@@ -235,16 +241,10 @@ def execute(self):
         submit them in parallel based on their dependencies.
         """
 
-        attempts = 0
         try:
             while not self.completed:
-                attempts += 1
-                if not attempts % 10:
-                    self._check_point()
```
phobologic: We should try and keep this - logging every attempt gets really noisy, since CF isn't that fast.

mhahn:
@phobologic made changes based on your comments.

Issue #175, opened 2016-07-23
Move to local parameters


mhahn:
We've talked about this a bunch, we should move to local parameters since we don't actually leverage the parameters being a part of the cloudformation template.

When we make this move, we should also drop defining parameters as a class variable. It makes subclassing a pain, ie:

```
    def get_local_parameters(self):
        params = getattr(self, "LOCAL_PARAMETERS", {})
        params.update({
            "DataDogAPIKey": {
                "type": str,
                "description": "API key for Data Dog",
            },
            "DataDogTags": {
                "type": str,
                "description": "Tags to include for Data Dog",
            },
        })
        setattr(self, "LOCAL_PARAMETERS", params)
        return super(Base, self).get_local_parameters()
```

mhahn:
local parameters also doesn't handle resolving dependencies from other outputs

mhahn:
I'd like to move away from the current `PARAMETERS` and `LOCAL_PARAMETERS` and move to a new concept of parameters with the following behavior:
- should be a simple python dictionary of values. this lets us leverage the YAML config to pass complex types (lists/dicts). (similar to `LOCAL_PARAMETERS`)
- should support resolving stack outputs, similar to `PARAMETERS`
- subclassing should be straightforward
- referencing within the blueprint should be straightforward

We can make this work in a backwards compatible way by leaving `PARAMETERS` and `LOCAL_PARAMETERS` the way they are (with plans to deprecate in the future) and move to this UX:

```
from stacker.blueprints.base import Blueprint


class NewBlueprint(Blueprint):

    BLUEPRINT_PARAMETERS = {
        'SecurityGroups': {
            'type': list,
        }
    }

    def create_template(self):
        # get_parameters returns `BLUEPRINT_PARAMETERS` by way of the
        # `defined_parameters` which makes it easy for subclasses to add
        # parameters
        params = self.get_parameters()
        for sg in params['SecurityGroups']:
            print sg


class NewSubclassBlueprint(NewBlueprint):

    def defined_parameters(self):
        params = super(Blueprint, self).defined_parameters()
        params['NewSubclassParameter'] = {
            'type': str,
            'description': 'Parameter',
        }
        return params

    def create_template(self):
        super(NewSubclassBlueprint, self).create_template()
        params = self.get_parameters()
        print params['NewSubclassParameter']
```

**changes**:

[...]

PR #194, opened 2016-08-12
Implement Variables and Lookups

mhahn:
These will deprecate PARAMETERS and LOCAL_PARAMETERS.

Advantages:
- No need to use CF parameters
- Leverage yaml lists and dicts
- Blueprint subclasses can easily add/overwrite parameters

Fixes: https://github.com/remind101/stacker/issues/175

I added: https://github.com/remind101/stacker_blueprints/pull/24/commits/93db0d2b385f6eb769893de27b756aa5493fe638 to stacker_blueprints as a POC for showing how we can leverage the new `BLUEPRINT_PARAMETERS`


stacker/blueprints/base.py
```diff
@@ -152,6 +158,97 @@ def setup_parameters(self):
...
+        """
+        if self.resolved_parameters is None:
+            raise UnresolvedBlueprintParameters(self)
+        return self.resolved_parameters
+
+    def resolve_parameters(self, parameter_values, provider, context):
+        """Resolve the values of the blueprint parameters.
+
+        resolve_parameters will resolve the values of the
+        `BLUEPRINT_PARAMETERS` with values from the command lie, the env file,
```
mhahn: typo

mhahn:
Now supports doing the following:

`DATABASE_URL: postgres://${empiredb_user}:${empiredb_password}@${empireDB::DBCname}/empire
`

`empireDB::DBCname` will get resolved at build time, `empiredb_user` will get resolved when we parse the config

stacker/blueprints/base.py
```diff
@@ -152,6 +173,95 @@ def setup_parameters(self):
...
+
+        These parameters will have been defined within `BLUEPRINT_PARAMETERS`
+        or `self.defined_parameters`. Any parameter value that depends on stack
+        output will have been resolved.
+
+        Returns:
+            dict: parameters available to the template
+
+        """
+        if self.resolved_parameters is None:
```
phobologic: Why not make this do the parameter resolution when it's called if resolved_parameters is None?

stacker/blueprints/base.py
```diff
@@ -152,6 +173,95 @@ def setup_parameters(self):
...
+        """Resolve the values of the blueprint parameters.
+
+        resolve_parameters will resolve the values of the
+        `BLUEPRINT_PARAMETERS` with values from the command lie, the env file,
+        the config, and any references to other stack output.
+
+        Args:
+            parameter_values (dict): dictionary of parameter key to parameter
+                value
+            provider (stacker.Provider): provider to use to resolve references
```
phobologic: Can you use this syntax for classes - it makes the API docs easier to use: https://github.com/remind101/stacker/blob/700ec034f2a9ef5711b8c21b5705eb9d5993b091/stacker/actions/build.py#L24

stacker/blueprints/base.py
```diff
@@ -152,6 +173,95 @@ def setup_parameters(self):
...
+                to output
+            context (stacker.Context): context to pull stack fqn from
+
+        """
+
+        def _resolve_string(value, provider, context):
```
phobologic: What about breaking these methods out and writing some test for them directly?

stacker/blueprints/base.py
```diff
@@ -152,6 +173,95 @@ def setup_parameters(self):
...
+            for match in OUTPUT_REGEX.finditer(value):
+                # Get from the Output(s) of another stack(s) in the
+                # stack_map
+                stack_output = match.group()
+                stack_name, output = stack_output.split("::")
```
phobologic: Does this still work when a value doesn't have any `::` in it?  It seems like it'd break with a `ValueError` when it tries to unpack it unsuccessfully.

phobologic: NM, it's iterating only over the parts that have `::` in them

stacker/blueprints/base.py
```diff
@@ -152,6 +173,95 @@ def setup_parameters(self):
...
+                stack_fqn = context.get_fqn(stack_name)
+                try:
+                    output_value = provider.get_output(stack_fqn, output)
+                except KeyError:
+                    raise OutputDoesNotExist(stack_fqn, stack_output)
+                value = OutputTemplate(value).safe_substitute(
```
phobologic: It took me a second to understand why you do the safe_substitute followed by the replace.  Might be worth commenting.

phobologic:
👍 other than the small comments I made.  Thanks!

stacker/blueprints/base.py
mhahn: I didn't want `get_parameters` to take any arguments and was invoking `resolve_parameters` where we were before: https://github.com/remind101/stacker/pull/194/files#diff-92ed81f5017042692188db7b5961c619R203. We need `stack.parameters`, `provider` and `context` which was readily available there, i'm not sure if we can get all of those values when we init the blueprint to make this resolve when called.

stacker/blueprints/base.py
mhahn: yea, i was being lazy

stacker/tests/blueprints/test_base.py
```diff
@@ -181,3 +183,48 @@ class TestBlueprint(Blueprint):
         parameters = blueprint.get_parameters()
         self.assertEqual(parameters["Param1"], 1)
         self.assertEqual(parameters["Param2"], "Test Output")
+
+    def test_resolve_string_comma_delmited_list(self):
+        provider = MagicMock()
+        context = MagicMock()
+        provider.get_output.return_value = "Test Output"
+        output = resolve_string("some-stack::Output, other-stack::Output",
```
phobologic: We should ditch this old syntax when using `blueprint_parameters` in preference for the new template syntax.

phobologic:
One thing I thought of - what if instead of `${stack::output}` being the generic syntax, what if we combined this with the concept of translators?  Like, what if the output lookup was a sort of translator, but with the new syntax?  Something like `${output MyStack::StackOutput}`?  Seems like that'd be more flexible (maybe it'd replace translators entirely?) and it'd make things more explicit.  I suppose, since the output lookup is such a common thing, we could keep generic `${MyStack::StackOutput}` to default to the output lookup.

mhahn:
we talked about this a little on slack, but to repeat for the record:

my concern with this is just that it gets confusing when the translator gets resolved. right now translators get resolved after the config is parsed, whereas outputs need to be resolved at build time. i think it would get confusing to have some translators resolved pre-build and others during build.

troyready:
Curious how this change would impact the generated templates' CFN parameters? Sounds like there won't be any, so I'm wondering if we could still add them to templates (using troposphere.add_parameter) while still using this new functionality?

It may be an errand of folly, but I'd like to keep adding CFN parameters for templates that I might share in built json form.

phobologic:
@troyready The idea was to eventually remove stacker support for parameters & local parameters in favor of these.  That wouldn't mean that you couldn't still add them directly via troposphere, it just wouldn't have all the helper logic around building them into the blueprints.

The reason for that is that using real CFN parameters in stacker has sort of proven to be pointless.  In theory you use parameters to make the templates re-usable, but in stacker you never actually re-use the templates (at least in the use cases I've seen - different environments usually have different buckets/namespaces - we end up with copies of the same template in two different buckets).

Do you often share the generated templates with others?

troyready:
It's definitely not something used in a normal stacker workflow, but it is nice to be able to share the json templates at times. It's something that I've done with colleagues (e.g. "even though you don't have stacker setup, you can just grab the template from this stack and use it with regular CFN").

Don't get me wrong, I'm not trying to argue for some [crazy workflow](https://xkcd.com/1172/) that depends on that. I just have some reservation about a change that significantly alters what stacker can output (maybe that does mean I depend on spacebar heating?).

mhahn:
@troyready 😆 re: https://xkcd.com/1172/

We're now calling these `variables` so it isn't out of the realm of possibility that we keep `PARAMETERS` around in some form, but I'd like to remove them and allow people to just use `add_parameter` in a blueprint if they really want it.

mhahn:
After talking with @phobologic about the initial implementation of blueprint parameters, we decided to go with a concept of "variables" that wouldn't be as easy to confuse with CF parameters. Variables are only related to how you can build templates within stacker.

This PR now introduces two new classes to stacker:

`Variable` and `Lookup`. A `Variable` is a python class that represents the variable passed within the stacker config, ie:

```
  # Create a simple app within empire named "acme-inc"
  - name: acmeApp
    class_path: stacker_blueprints.empire.app.App
    variables:
      ServiceToken: ${empireDaemon::CustomResourcesTopic}
      Name: "acme-inc"
```

`ServiceToken` and `Name` are `Variable`s.

A `Variable` can have zero or many `Lookup`s within its value that will be resolved at build time and will then be available within the `Blueprint` as a dictionary of `<variable name>: <resolved value>`.

A `Lookup` is defined as `${<lookup type> <lookup input>}` within the stacker config. We default to the `output` lookup since this is the most common use case. ie `${empireDaemon::CustomResourcesTopic}` could be more verbosely defined as `${output empireDaemon::CustomResourcesTopic}`.

Custom lookups can be registered within the config: https://github.com/remind101/stacker_blueprints/pull/24/files#diff-8f84670404b5e0466057d169e260c840R58, https://github.com/remind101/stacker_blueprints/pull/24/files#diff-7cb05c06f16d1eea2c363ecf8e0d1463R3

They can also be nested: https://github.com/remind101/stacker_blueprints/pull/24/files#diff-8f84670404b5e0466057d169e260c840R207

When we parse the config, we register any custom lookups that have been defined. At build time, we resolve all of the variables, invoking the lookup handlers that have been defined. When the `Blueprint` renders its template, variables will have been translated into a dictionary with any lookups resolved.

In the near future, we'll collapse the concept of "translators" into lookups given that their execution at build time and access to `context` and `provider` makes them much more powerful.

stacker/blueprints/base.py
```diff
@@ -98,13 +104,15 @@ class Blueprint(object):
             template.
 
     """
-    def __init__(self, name, context, mappings=None):
+
+    def __init__(self, name, context, variables=None, mappings=None):
```
mhahn: this isn't used

stacker/blueprints/base.py
```diff
@@ -152,6 +160,61 @@ def setup_parameters(self):
...
+
+        """
+        return getattr(self, "VARIABLES", {})
+
+    def get_variables(self):
+        """Return a dictionary of variables available to the template.
+
+        These variables will have been defined within `VARIABLES` or
+        `self.defined_variables`. Any variable value that depends on stack
+        output will have been resolved.
```
mhahn: switch this to talking about lookups instead of outputs

stacker/blueprints/base.py
```diff
@@ -152,6 +160,61 @@ def setup_parameters(self):
...
+        """
+        if self.resolved_variables is None:
+            raise UnresolvedVariables(self)
+        return self.resolved_variables
+
+    def resolve_variables(self, variables):
+        """Resolve the values of the blueprint variables.
+
+        This will resolve the values of the `VARIABLES` with values from the
+        env file, the config, and any references to other stack output.
```
mhahn: switch to talking about lookups

stacker/blueprints/base.py
```diff
@@ -152,6 +160,61 @@ def setup_parameters(self):
...
+        for var_name in defined_variables.iterkeys():
+            if var_name not in variable_dict:
+                raise MissingVariable(self, var_name)
+
+            variable = variable_dict[var_name]
+            if not variable.resolved:
+                raise UnresolvedVariable(self, variable)
+
+            if variable.value is None:
+                logger.debug("Got None value for parameter %s, not submitting "
```
mhahn: s/parameter/variable

stacker/config/__init__.py
```diff
@@ -30,6 +30,10 @@ def parse_config(raw_config, environment=None):
         buff.write(t.substitute(environment))
     except KeyError, e:
         raise exceptions.MissingEnvironment(e.args[0])
+    except ValueError:
+        # Support "invalid" placeholders for output values (ie. placeholders
```
mhahn: this should change to mention lookups

stacker/stack.py
```diff
@@ -94,11 +105,23 @@ def requires(self):
                 continue
             for stack_name in stack_names:
                 stack_fqn = self.context.get_fqn(stack_name)
-                if stack_fqn not in requires:
-                    requires.add(stack_fqn)
+                requires.add(stack_fqn)
+
+        for lookup in self.lookups:
```
mhahn: explain this

stacker/stack.py
```diff
@@ -94,11 +105,23 @@ def requires(self):
...
+            d = deconstruct(lookup.input)
+            stack_fqn = self.context.get_fqn(d.stack_name)
+            requires.add(stack_fqn)
+
         return requires
 
     @property
+    def lookups(self):
```
mhahn: docstring

stacker/variables.py
```diff
@@ -0,0 +1,132 @@
...
+)
+
+
+class OutputTemplate(Template):
+    """A custom string template we use to replace lookup values"""
+    idpattern = r'[_a-z][_a-z0-9\s\:\-\.\,]*'
+
+
+def resolve_string(value, replacements):
+    """Resolve any output lookups within a string.
```
mhahn: remove output

stacker/variables.py
```diff
@@ -0,0 +1,132 @@
...
+
+    Args:
+        value (str): string value we're resolving lookups within
+        replacements (dict): resolved output values
```
mhahn: remove output

stacker/blueprints/base.py
```diff
@@ -152,6 +160,60 @@ def setup_parameters(self):
...
+        return self.resolved_variables
+
+    def resolve_variables(self, variables):
+        """Resolve the values of the blueprint variables.
+
+        This will resolve the values of the `VARIABLES` with values from the
+        env file, the config, and any lookups resolved.
+
+        Args:
+            variables (list): list of variables
```
phobologic: This is a list of a specific class, right?  Can you use the `:class:` syntax to point at the actual class that you want a list of?  Right now it's hard to tell what a `list of variables` means.

stacker/lookups/output.py
```diff
@@ -0,0 +1,35 @@
+from collections import namedtuple
```
phobologic: worth moving this into a subdirectory called `handlers/` or something like that?

stacker/context.py
```diff
@@ -70,7 +71,11 @@ def load_config(self, conf_string):
             self.bucket_name = bucket_name
         tags = self.config.get("tags", None)
         if tags is not None:
-            self.tags = dict([(str(tag_key), str(tag_value)) for tag_key, tag_value in tags.items()])
+            self.tags = dict([(str(tag_key), str(tag_value)) for tag_key,
+                              tag_value in tags.items()])
+        lookups = self.config.get("lookups", {})
+        for key, handler in lookups.iteritems():
```
phobologic: Will all lookups need to be registered in the config, or will some be loaded automatically?  If they do have to be loaded, is there any reason why you chose that over loading all of them automatically?

stacker/variables.py
```diff
@@ -0,0 +1,132 @@
...
+            value[key] = resolve(v, replacements)
+        return value
+    return value
+
+
+def resolve_variables(variables, context, provider):
+    """Given a list of variables, resolve all of them.
+
+    Args:
+        variables (list): list of `stacker.variables.Variables`
```
phobologic: class syntax

stacker/variables.py
```diff
@@ -0,0 +1,132 @@
...
+        """
+        while self.lookups:
+            resolved_lookups = resolve_lookups(self.lookups, context, provider)
+            self.replace(resolved_lookups)
+
+    def replace(self, resolved_lookups):
+        """Replace lookups in the Variable with their resolved values.
+
+        Args:
+            resolved_lookups (dict): dict of Lookup -> resolved value.
```
phobologic: `:class:` syntax for Lookup

phobologic:
This is looking so awesome.  A couple of nits, and a few questions, but I'm really excited for this.

stacker/context.py
mhahn: You only need to register the lookup in the config if it is a custom one that isn't provided by stacker. https://github.com/remind101/stacker_blueprints/pull/24/files#diff-8f84670404b5e0466057d169e260c840R58 is an example of that. Anyone defined within stacker should register itself via: https://github.com/remind101/stacker/pull/194/files#diff-00537bebd4f02f9f6ab51659e6716ef9R51. Output was the only one I did for now, but KMS etc. should be translated to stacker provided lookups that you don't have to do anything special to register in the config. Does that answer your questio?

phobologic: Cool, yeah, that makes sense.  Thanks.

phobologic:
Also, once you're feeling comfortable with this, can you add some documentation?  Thanks!

stacker/lookups/output.py
mhahn: lookups and handlers are essentially the same. any new lookup should have its own file within the `lookups` directory

phobologic: Sorry, I meant a new subdirectory in: `lookups/handlers` since this directory also has the registry in it.  Probably not a big deal, just thought it'd make sense to separate them out for future readers.

mhahn: ah i see what you mean. let me convert the KMS translator and lets see how it looks. just not sure what the right syntax should be since i've only done outputs so far

mhahn: i think you're right, i'll do it in this one

mhahn:
going to add docs in a separate PR, merging this in now!

PR #198, opened 2016-08-23
Fix traceback on interactive provider when adding resources

danielkza:
It seems it's possible for resources to omit the 'Replacement' key
instead of setting it to false. Handle that case instead of assuming the
key is always present.

Fixes a traceback like the following: 

```
  File "/home/danielkza/bin/stacker", line 6, in <module>
    exec(compile(open(__file__).read(), __file__, 'exec'))
  File "/home/danielkza/src/stacker/scripts/stacker", line 9, in <module>
    args.run(args)
  File "/home/danielkza/src/stacker/stacker/commands/stacker/build.py", line 52, in run
    dump=options.dump)
  File "/home/danielkza/src/stacker/stacker/actions/base.py", line 125, in execute
    self.run(*args, **kwargs)
  File "/home/danielkza/src/stacker/stacker/actions/build.py", line 301, in run
    plan.execute()
  File "/home/danielkza/src/stacker/stacker/plan.py", line 251, in execute
    if not self._single_run():
  File "/home/danielkza/src/stacker/stacker/plan.py", line 219, in _single_run
    status = step.run()
  File "/home/danielkza/src/stacker/stacker/plan.py", line 72, in run
    return self._run_func(self.stack, status=self.status)
  File "/home/danielkza/src/stacker/stacker/actions/build.py", line 220, in _launch_stack
    tags)
  File "/home/danielkza/src/stacker/stacker/providers/aws/interactive.py", line 164, in update_stack
    replacements_only=self.replacements_only)
  File "/home/danielkza/src/stacker/stacker/providers/aws/interactive.py", line 83, in output_summary
    replacement = resource['Replacement'] == 'True'
KeyError: 'Replacement'
```

### DIFF
--- a/stacker/providers/aws/interactive.py
+++ b/stacker/providers/aws/interactive.py
@@ -80,7 +80,7 @@ def output_summary(fqn, action, changeset, replacements_only=False):
     changes = []
     for change in changeset:
         resource = change['ResourceChange']
-        replacement = resource['Replacement'] == 'True'
+        replacement = resource.get('Replacement') == 'True'
         summary = '- %s %s (%s)' % (
             resource['Action'],
             resource['LogicalResourceId'],


stacker/providers/aws/interactive.py
```diff
@@ -80,7 +80,7 @@ def output_summary(fqn, action, changeset, replacements_only=False):
     changes = []
     for change in changeset:
         resource = change['ResourceChange']
-        replacement = resource['Replacement'] == 'True'
+        replacement = resource.get('Replacement', '') == 'True'
```
phobologic: Only a minor nit: You can just do `resource.get('Replacement') == 'True'`

`.get()` will return `None` by default if the key doesn't exist.

phobologic:
This looks good - one minor change, and I'll go ahead and merge.  Thanks @danielkza !

stacker/providers/aws/interactive.py
danielkza: Right. I got into the habit of making it explicit for clarity, and I thought it would maybe be better to return a string for uniformity, but it's only a matter of preference, and keeping it like the rest of the codebase is best.

phobologic: Yeah - I think I prefer just letting it use the usual default, unless there's actually something useful about specifying a new default.  If you can update that, I'll pull this in ASAP :)  Thanks @danielkza

danielkza: @phobologic Done.

phobologic:
Thanks!

PR #200, opened 2016-08-29
Convert KMS translator to a Lookup

mhahn:
This converts the KMS translator to a Lookup.

This changes the syntax for referencing `kms` encrypted values within environment & configs, instead of `conf_value: !kms <encrypted value>` you should use the new lookup syntax: `conf_value: ${kms <encrypted value>}`.


phobologic:
Just off the top of my head, lets not remove the translator right away.  We should keep it around, and slowly migrate away from the old syntax to the new.

mhahn:
yea, probably a good call

phobologic:
👍 looks good, looks like there's a conflict from other changes, can you refresh it and deal with the conflicts, and then go ahead and merge.  Thanks.

PR #201, opened 2016-08-29
Use compatible release versions for all dependencies

mhahn:
Use https://www.python.org/dev/peps/pep-0440/#compatible-release to keep dependencies stable between stacker and stacker_blueprints.


phobologic:
Nice find.

Issue #184, opened 2016-08-03
Allow cross-namespace output gathering for parameters


phobologic:
This comes from @acmcelwee in https://github.com/remind101/stacker/pull/169#issuecomment-233201214.  This is something I've thought about before, but hadn't come up with a good conclusion.

One idea is that we could have stacker attempt to find the stack with the namespace, and if that fails fall back to looking for the stack without the namespace.  The only downside to that is that it feels like it could do something accidentally, but in that case I get the feeling it'd just fail to create the stack at all.

Another option is to come up with another syntax that indicates to look outside of the stack definition at stacks.  not sure what that syntax would be.

CC: @ejholmes @mhahn

phobologic:
And I realized now that I missed  @acmcelwee 's proposal for a syntax - basically a leading `::` in front would indicate that it's a fully qualified stack name - that seems fine to me.

acmcelwee:
@phobologic, thanks for pulling this over. It hasn't been a blocker for me yet, so I haven't spent any brain cycles on it lately. I'm quoting your contract expectation from [#169 (comment)](https://github.com/remind101/stacker/pull/169#issuecomment-234603481)

> I don't think it has to be that complicated - the contract could be that if you specify a full stack name for an Output that it no longer guarantees much other than it's in a 'consistent' state (ie: UPDATE_COMPLETE, CREATE_COMPLETE, etc).

The combination of a leading `::` for fully-qualified stacks + the contract stated above seems like enough to provide a lot of useful functionality without complicating things.

mhahn:
@acmcelwee can you take a look at: https://github.com/remind101/stacker/pull/202 and see if that addresses your issue here?

PR #202, opened 2016-08-30
xref lookup

mhahn:
Depends on this PR: https://github.com/remind101/stacker/pull/201. I was having some issues with the version of `mock` and what was required by `moto`.

This should fix: https://github.com/remind101/stacker/issues/184. It is the same as the `output` Lookup, except we don't use the `Context` to resolve the fully qualified name of the stack.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -7,20 +7,20 @@ VERSION = "0.6.3"
 src_dir = os.path.dirname(__file__)
 
 install_requires = [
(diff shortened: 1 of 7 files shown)


phobologic:
Awesome - the new Lookup concept makes this stuff so easy.

acmcelwee:
This looks great. It'll provide some nice flexibility in breaking up some of the stack contexts from a single large file for us.

PR #203, opened 2016-08-30
Add documentation on lookups and variables

mhahn:


docs/config.rst
```diff
@@ -174,18 +197,21 @@ Here's an example from stacker_blueprints_, used to create a VPC::
...
+
+Parameters are a CloudFormation concept that allow you to re-use an existing
+CloudFormation template, but modify its behavior by passing in different
 values.
 
 stacker tries to make working with Parameters a little easier in a few ways:
 
-YAML anchors & references
--------------------------
+Parameter YAML anchors & references
```
phobologic: These can still be used in blueprint variables, right?  It's more a yaml concept, than a stacker one.

phobologic: NM, see you added it into blueprint variables as well.  Ignore.

docs/lookups.rst
```diff
@@ -0,0 +1,124 @@
...
+
+  DockerConfig: ${kms file://dockercfg}
+
+.. note::
+  Lookups resolve the path specified with `file://` relative to
+  the location of the config file, not where the stacker command is run.
+
+.. _xref:
+
+XRef Lookup
```
phobologic: Probably worth mentioning that the xref lookup will not setup any dependencies like output does.

phobologic:
👍

docs/config.rst
mhahn: yea was just trying to avoid the duplicate reference. this parameter section should be deleted

docs/lookups.rst
mhahn: yea, i was waffling on it but its probably better to just be explicit

PR #204, opened 2016-08-30
Docstring consistency

mhahn:
Some small fixes to docstring values i noticed when going through the api docs.

Trying to follow: http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html


phobologic:
Thanks for cleaning this up.

This makes me think we should have a `CONTRIBUTING.rst` file to detail the styleguide for things like this (also to mention that when you contribute things you should be sure to write docs/tests, mention moto, etc).

Issue #213, opened 2016-09-01
Having a slash inside a lookup seems to cause infinite recursion


ttarhan:
While working in the ${file...} lookup, I discovered that having a slash inside the lookup, such as:

`${file parameterized-b64:file://userdata.txt}`

causes infinite recursion in Variable.resolve, on line 118 (while self.lookups). Simply removing the slashes in the path above avoids the problem.

I haven't looked into the root cause yet.

phobologic:
I'm taking a guess, but it's likely due to this line: https://github.com/remind101/stacker/blob/master/stacker/variables.py#L11 - you might need to include `\/` in the idpattern.  @mhahn will know best.

ttarhan:
@phobologic indeed, that fixes it. I'll put in a simple PR just for that fix right now.

PR #215, opened 2016-09-01
Resolves issue #213 - slashes inside variable lookups

ttarhan:
Simple fix to include "/" in the idpattern for the variable template

### DIFF
--- a/stacker/variables.py
+++ b/stacker/variables.py
@@ -8,7 +8,7 @@ from .lookups import (
 
 class LookupTemplate(Template):
     """A custom string template we use to replace lookup values"""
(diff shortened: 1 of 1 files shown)


phobologic:
Thanks for this - since I haven't had a chance to become super acquainted with the lookups code as much as @mhahn, I'm going to leave it to him to review.

mhahn:
I think we can just change this to something like `^[\$\{\}]` otherwise we'll keep having to add values here.

ttarhan:
@mhahn - changed, but I just made the change quickly online and haven't tested yet.

mhahn:
looks good!

Issue #214, opened 2016-09-01
Lookups can no longer return complex objects


ttarhan:
While working on the ${file. ...} lookup, I ran into an issue: the lookups serve as simple string substitution. The file lookup needs to return troposphere objects (Base64 or GenericHelperFn) to return the required JSON structure:

         { "Fn::Join" : ["", [
           "#!/bin/sh\naws s3 sync s3://",
           {"Ref" : "BucketName"},
           "/somepath /somepath"
         ]] }

which cannot be represented as a String. I'm not sure where to go from here, as this worked fine with translators (since they could return any object), but it's broken with the current implementation of lookups.

mhahn:
I need to think about this a little more.

We need to change: https://github.com/remind101/stacker/blob/master/stacker/variables.py#L26 to support this.

off the top of my head, one option (which would break nested lookups):

- determine what lookups are within the string
- if any of the lookups have a non-string value, that should be returned and all the others will be ignored. Any wrapping lookup would be a noop.

ie. if you had something like `${custom ${file ...}}`, since file returns an object, `custom` would never get called.

I was thinking another option could be storing a map between any non-string lookup value and a UUID. We could replace the UUID in the string and then nested lookups could lookup the value when they get invoked, but there are too many cases where that gets weird, ie:

`${custom ${file ...},${output someStack::someOutput}}`

if that resolved to something like: `${custom _lookup-res:123456,Output Value}` I'm not sure how useful that would be.

Right now the input for a lookup handle is unstructured, you get whatever string value is present. We could add more structure and treat the input value as something that will be resolved to `args` and `kwargs`. In the above example, we would split the input value on `,` and then invoke `custom` with:

`custom([<result of file lookup>, "Output Value"])`

I may be thinking about this too much though.

ttarhan:
I think we should go one of two extremes:

1) Very Simple -- non-string lookups don't support nesting, and it's invalid to nest them. If any lookup within a set of nested lookups returns a non-string value, it's an error. So you can create lookups like ${file} that return whatever you want them to return, but you can't nest them.

2) Go All The Way -- use kwawgs (I would't bother with args and kwawgs, personally) and have a syntax for lookups that's structured ${something(a=b,c=d,e=${another(v="asd")}}. I'm not married to the syntax, but you get the idea. With this structured (as opposed to string) input, it's up to the outer lookup to decide what to do with a non-string value.

mhahn:
Yea, I say we go with 1 until we have another use case.

[...]

PR #216, opened 2016-09-01
Support lookups returning non-string values

mhahn:
Fixes: https://github.com/remind101/stacker/issues/214

One caveat is that you need to make sure the variable definition uses the correct type in the blueprint. 

ie: https://github.com/remind101/stacker/pull/216/files#diff-a5840ad0d3853c3a46851aabf2324832R150


phobologic:
👍

Issue #217, opened 2016-09-02
Should blueprint variables accept non-python built in types


phobologic:
Right now we only accept basic python types and CFNTypes due to this code:

https://github.com/remind101/stacker/blob/378d0bfb989900a4c688cc51f25f7702cdbfac87/stacker/blueprints/base.py#L285

That works great, but what if we want more specific types.  For example, I am building a blueprint that will allow you to build an SNS queue with multiple subscriptions.  My idea for the config would be something like:

```
  topics:
    topic1:
      - subscription1
      - subscription2
```

Right now I can only define `topics` as a `dict` type, but that leaves a lot of room for error.  Instead, what I'd like to do is provide my own validation function, something like:

```
def validate_topics(value):
    if not isinstance(value, dict):
        raise InvalidVariable("topics must be a dictionary")
    for k, v in value.items():
        if v and not isinstance(v, list):
            raise InvalidVariable("Topics should include a list of subscriptions.")
    return True
```

phobologic:
@mhahn assigning you  just to get your thoughts, you don't have to handle this if you're ok with it, I can write a PR

phobologic:
Talked it over with @mhahn, and I think instead I'm going to work on an optional `validator` setting for Variables, that will let you do custom validation of the contents of the variable.

mhahn:
👍

mhahn:
@phobologic you can close this now right?

phobologic:
Yup, closed.

PR #218, opened 2016-09-03
Add custom validator to blueprint variables

phobologic:
- Now all blueprint variables must provide a `type`, otherwise an
  exception is thrown.

Before it wasn't required, and the type was ignored if it wasn't
provided. That didn't seem right to me, but I realize I may just not
have understood the implications.

Other changes:

- variable resolution now done in a function not a class method
- moved variable type validation to a function
- instead of passing the whole blueprint just to use it's name, I just
  pass the blueprints name into the exceptions that use them (that was
  all the blueprint was used for it appeared)

I'll add docs in the next checkin.

### DIFF
--- a/stacker/blueprints/base.py
+++ b/stacker/blueprints/base.py
@@ -11,8 +11,9 @@ from troposphere import (
 from ..exceptions import (
     MissingLocalParameterException,
     MissingVariable,
-    UnresolvedVariables,
     UnresolvedVariable,
+    UnresolvedVariables,
+    VariableTypeRequired,
 )
 from .variables.types import CFNType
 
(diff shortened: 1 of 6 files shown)


phobologic:
@mhahn @jeffpierce

phobologic:
Fixes #217

stacker/blueprints/base.py
```diff
@@ -130,6 +131,90 @@ def build_parameter(name, properties):
...
+        # Variable value not provided, try using the default, if it exists
+        # in the definition
+        try:
+            value = var_def["default"]
+        except KeyError:
+            raise MissingVariable(blueprint_name, var_name)
+
+    # If no validator, return the value as is, otherwise apply validator
+    validator = var_def.get("validator", lambda v: v)
+    value = validator(value)
```
mhahn: I think we should have a preferred exception that validators raise that gets caught here. Then we can include more details about which variable was invalid etc.

Right now if you raise an exception within the validator, you can't tell which variable was being validated.

mhahn: or maybe just any exception that is raised by the validator we wrap with some other info?

phobologic: Good call - I'll think about it and come up with something that makes it easy to tell which variable failed it's validator, and why.

stacker/blueprints/base.py
```diff
@@ -130,6 +132,93 @@ def build_parameter(name, properties):
...
+            value = var_def["default"]
+        except KeyError:
+            raise MissingVariable(blueprint_name, var_name)
+
+    # If no validator, return the value as is, otherwise apply validator
+    validator = var_def.get("validator", lambda v: v)
+    try:
+        value = validator(value)
+    except Exception as exc:
+        raise ValidatorError(var_name, validator.__name__, value, exc)
```
mhahn: add this to docstring above

mhahn:
👍

PR #222, opened 2016-09-05
Cleanup misc flake8 errors in prep for auto-running

phobologic:
In preparing for running flake8 during CI, go ahead and cleanup all the flake8 errors I could find.  The only case I have not cleaned up is the use of setUp not being capitalized correctly per pylint when used in unittest. I have other plans for fixing that.


phobologic:
Closing in favor of another PR with more work around this.

PR #229, opened 2016-09-15
Remove deprecated parameter logic

mhahn:
This is tough to break up into smaller pull requests since it touches every level of stacker.

This moves everything over to variables, some notable changes:

- you can no longer pass `-p` on the command line, this has been replaced with `-var`
- `parameters` are no longer recognized within stack definitions
- Support for `PARAMETERS`  and `LOCAL_PARAMETERS` has been dropped


danielkza:
Would it be possible to make a stable release before landing this? Otherwise I'll have to migrate a bunch of stacks from the old parameters to get other bug fixes.

mhahn:
@danielkza yes! thanks for reminding me, this should be going into the `release-1.0` branch, not master.

stacker/actions/build.py
```diff
@@ -53,119 +60,112 @@ def should_submit(stack):
...
             dict: The parameters for the given stack
+
         """
-        parameters = self._resolve_parameters(stack.cfn_parameters,
-                                              stack.blueprint)
-        required_params = [k for k, v in stack.blueprint.required_parameters]
-        parameters = self._handle_missing_parameters(parameters,
-                                                     required_params,
-                                                     provider_stack)
+        resolved = _resolve_parameters(stack.parameters, stack.blueprint)
```
phobologic: We talked about this, but figured I'd put this here - I think we should start referring to `parameters` in this usage instead as `parameter_values` to be more explicit, and help avoid confusion going forward.

stacker/blueprints/base.py
```diff
@@ -239,63 +187,63 @@ def __init__(self, name, context, mappings=None):
...
+        be returned as a CloudFormation Parameter.
 
         Returns:
             dict: parameter definitions. Keys are parameter names, the values
-                are dicts containing key/values for various parameter
-                properties.
-        """
-        parameters = getattr(self, "CF_PARAMETERS",
-                             getattr(self, "PARAMETERS", {}))
+            are dicts containing key/values for various parameter properties.
```
phobologic: I think there's a whitespace issue here - this line should be indented further.

stacker/blueprints/base.py
```diff
@@ -239,63 +187,63 @@ def __init__(self, name, context, mappings=None):
...
             var_type = attrs.get("type")
             if isinstance(var_type, CFNType):
                 cfn_attrs = copy.deepcopy(attrs)
                 cfn_attrs["type"] = var_type.parameter_type
-                parameters[var_name] = cfn_attrs
-        return parameters
+                output[var_name] = cfn_attrs
+        return output
+
+    def get_parameters(self):
```
phobologic: `get_parameter_values`

stacker/blueprints/base.py
```diff
@@ -239,63 +187,63 @@ def __init__(self, name, context, mappings=None):
...
+
+        Returns:
+            dict: variables that need to be submitted as CloudFormation
+                Parameters.
+
+        """
+        variables = self.get_variables()
+        output = {}
+        for key, value in variables.iteritems():
+            if hasattr(value, "to_parameter_value"):
```
phobologic: Should probably replace this with:

```
try:
    output[key] = value.to_parameter_value()
except AttributeError:
    continue
```

It's a nit, but doing hasattr doesn't check that `to_parameter_value()` is actually callable.

stacker/commands/stacker/base.py
```diff
@@ -116,10 +116,10 @@ def get_context_kwargs(self, options, **kwargs):
 
     def add_arguments(self, parser):
         # global arguments that should be available on all stacker subcommands
-        parser.add_argument("-p", "--parameter", dest="parameters",
-                            metavar="PARAMETER=VALUE", type=key_value_arg,
+        parser.add_argument("-var", "--variable", dest="variables",
```
phobologic: Does this work?  Can you have 'long' variables with single dashes?  Maybe we should just pick one of these, and not have a short option.

stacker/commands/stacker/base.py
```diff
@@ -116,10 +116,10 @@ def get_context_kwargs(self, options, **kwargs):
...
+                            metavar="VARIABLE=VALUE", type=key_value_arg,
```
phobologic: Maybe use `BLUEPRINT_VARIABLE=VALUE` here

stacker/context.py
```diff
@@ -41,7 +41,7 @@ class Context(object):
 
     def __init__(self, environment,  # pylint: disable-msg=too-many-arguments
                  stack_names=None,
-                 parameters=None, mappings=None, config=None,
+                 variables=None, mappings=None, config=None,
```
phobologic: I know you didn't introduce this, but why don't we just split these up on multiple lines - this seems out of place.

stacker/stack.py
```diff
@@ -11,48 +11,47 @@
...
 
     Returns:
-        dict: Contains key/value pairs of the collected parameters.
+        dict: Contains key/value pairs of the collected variables.
+
     """
-    parameters = copy.deepcopy(stack_def.get("parameters", {}))
-    stack_specific_params = {}
-    for key, value in context_parameters.iteritems():
+    variables = copy.deepcopy(stack_def.get('variables', {}))
```
phobologic: Rename variable to `variable_values` maybe?

stacker/commands/stacker/base.py
mhahn: this seems to work:

```
$ stacker build -var test
usage: stacker build [-h] [-var VARIABLE=VALUE] [-e ENV=VALUE] [-r REGION]
                     [-v] [-i] [--replacements-only] [-m MAX_ZONES] [-o]
                     [--force STACKNAME] [--stacks STACKNAME] [-t] [-d DUMP]
                     environment config
stacker build: error: argument -var/--variable: test does not match KEY=VALUE format
```

I'm fine just picking one too though

phobologic: The unix nerd in me is freaked out by single dash long options.  Lets just have `--var` :)

Issue #228, opened 2016-09-12
Commands hang with circular variable


syphon7:
Very small issue -

When an output is used as a variable to the stack that creates the output stacker hangs rather than errors.

E.g.
`
-name: main-vpc`
`
   variables:`
`
        TestVar: ${main-vpc::TestVar}
`

This is an invalid variable, but it should detect that rather than hang indefinitely.

phobologic:
@syphon7 thanks for noticing this, will try to get it cleaned up asap.

PR #230, opened 2016-09-15
Don't allow circular references within stack variables

mhahn:
Fixes: https://github.com/remind101/stacker/issues/228

### DIFF
--- a/stacker/stack.py
+++ b/stacker/stack.py
@@ -131,14 +131,6 @@ class Stack(object):
         return requires
 
     @property
-    def lookups(self):
-        """Return a set of lookups contained by stack variables"""
-        lookups = set()
-        for variable in self.variables:
-            lookups = lookups.union(variable.lookups)
-        return lookups
-
-    @property
     def blueprint(self):
         if not hasattr(self, "_blueprint"):
             class_path = self.definition["class_path"]
--- a/stacker/stack.py
+++ b/stacker/stack.py
@@ -111,16 +111,22 @@ class Stack(object):
                 requires.add(stack_fqn)
 
         # Add any dependencies based on output lookups
-        for lookup in self.lookups:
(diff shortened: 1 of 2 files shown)


stacker/stack.py
```diff
@@ -114,6 +114,11 @@ def requires(self):
         for lookup in self.lookups:
             if lookup.type == OUTPUT_LOOKUP_TYPE_NAME:
                 d = deconstruct(lookup.input)
+                if d.stack_name == self.name:
+                    message = (
+                        "Lookup: \"%s\" is a circular reference to: %s"
```
phobologic: Can we provide the stack & variable that this lookup happens in as part of this error?  Might make it easier to chase down what is actually broken in some cases.  Especially with anchors and what not.

mhahn: Yea, error message now looks like this:

```
Variable Var1 in stack vpc.1 has a ciruclar reference within lookup: vpc.1::FakeOutput
```

stacker/stack.py
```diff
@@ -111,11 +111,22 @@ def requires(self):
                 requires.add(stack_fqn)
 
         # Add any dependencies based on output lookups
-        for lookup in self.lookups:
-            if lookup.type == OUTPUT_LOOKUP_TYPE_NAME:
-                d = deconstruct(lookup.input)
-                stack_fqn = self.context.get_fqn(d.stack_name)
-                requires.add(stack_fqn)
+        for variable in self.variables:
```
mhahn: in stacker v1.0 this is the only code within this function, so i don't think it needs to be broken out more

phobologic: Cool - is there any reason to do away with `stack.lookups`?

mhahn: yea we can kill them.

PR #231, opened 2016-09-15
Support resolving variables when calling `dump`

mhahn:
We now resolve variables when calling `dump`.

We resolve the variables with a `provider` set to`None` to prevent calls to the provider. `dump` is analogous to "dry run without building anything", so there should be no need to access the provider, run `pre_build` or `post_build` hooks.

### DIFF
--- a/stacker/actions/build.py
+++ b/stacker/actions/build.py
@@ -256,7 +256,12 @@ class Action(BaseAction):
     def pre_run(self, outline=False, dump=False, *args, **kwargs):
         """Any steps that need to be taken prior to running the action."""
         pre_build = self.context.config.get("pre_build")
-        if not outline and not dump and pre_build:
+        should_run_hooks = (
+            not outline and
+            not dump and
+            pre_build
+        )
+        if should_run_hooks:
             util.handle_hooks("pre_build", pre_build, self.provider.region,
                               self.context)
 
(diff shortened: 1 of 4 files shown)


mhahn:
Depends on merging: https://github.com/remind101/stacker/pull/229

stacker/actions/build.py
```diff
@@ -253,10 +253,10 @@ def _get_dependencies(self):
             dependencies[stack.fqn] = stack.requires
         return dependencies
 
-    def pre_run(self, outline=False, *args, **kwargs):
+    def pre_run(self, outline=False, dump=False, *args, **kwargs):
         """Any steps that need to be taken prior to running the action."""
         pre_build = self.context.config.get("pre_build")
-        if not outline and pre_build:
+        if not outline and not dump and pre_build:
```
phobologic: This logic is getting a little long, should we maybe make this a method? `should_run_hooks` or something?

stacker/lookups/registry.py
```diff
@@ -24,6 +24,19 @@ def register_lookup_handler(lookup_type, handler_or_path):
     LOOKUP_HANDLERS[lookup_type] = handler
 
 
+def unregister_lookup_handler(lookup_type):
```
phobologic: Is this only used in tests?

mhahn: yes

mhahn: the docstring:

```    """Unregister the specified lookup type.

    This is useful when testing various lookup types if you want to unregister
    the lookup type after the test runs.

    Args:
        lookup_type (str): Name of the lookup type to unregister

    """
```

PR #235, opened 2016-09-28
Increase how often we output the plan check point if we're using the loop logger

mhahn:
Depends on merging: https://github.com/remind101/stacker/pull/229

Before, we were only displaying the plan check point on every 10 attempts. This made sense if we were outputting the plan status each time. Now that we use the loop logger and update the same lines, we can output the plan status each time.

### DIFF
--- a/stacker/plan.py
+++ b/stacker/plan.py
@@ -1,4 +1,5 @@
 from collections import OrderedDict
+import hashlib
 import logging
 import multiprocessing
 import os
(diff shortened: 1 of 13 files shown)


stacker/plan.py
```diff
@@ -317,6 +328,15 @@ def dump(self, directory):
 
         self.reset()
 
+    @property
+    def md5(self):
```
phobologic: Might be worth a docstring here detailing what this is used for.

PR #238, opened 2016-10-05
add UsePreviousValue support for NoEcho parameters

troyready:
Some values (e.g. passwords) are useful to pass in once during stack
creation with a NoEcho flag. In future stack updates, it should then be
possible to omit those parameters and have CFN use the previous value
submitted.

This commit enables that functionality, when using NoEcho parameters.

### DIFF
--- a/stacker/actions/build.py
+++ b/stacker/actions/build.py
@@ -166,10 +166,19 @@ class Action(BaseAction):
         parameters = self._handle_missing_parameters(parameters,
                                                      required_params,
                                                      provider_stack)
-        return [
+
+        # Handle values that shouldn't be sent ('UsePreviousValue')
+        final_list = []
+        for p in [
             {'ParameterKey': p[0],
-             'ParameterValue': str(p[1])} for p in parameters
-        ]
+             'ParameterValue': str(p[1])} for p in parameters]:
+            stack_p = stack.blueprint.parameters[p['ParameterKey']]
+            if p['ParameterValue'] == '****' and stack_p.properties['NoEcho']:
+                final_list.append({'ParameterKey': p['ParameterKey'],
+                                   'UsePreviousValue': True})
+            else:
+                final_list.append(p)
+        return final_list
 
     def _build_stack_tags(self, stack):
         """Builds a common set of tags to attach to a stack"""


troyready:
Not sure if this is too narrow/opinionated to only add the functionality for NoEcho params, but if nothing else I figured it was super safe to start there and open the conversation.

mhahn:
Is this for a stack that isn't created within stacker?

For passwords, I use the kms translator (now lookup) to store the encrypted value within the environment file. This is useful so you can track when this value changes and should be safe to store in version control since it is encrypted

troyready:
@mhahn yeah, that's probably preferable in most scenarios.

I still run into this need when creating stacker stacks using a db password not stored in KMS (e.g. LastPass, 1Password, etc).

stacker/actions/build.py
```diff
@@ -166,10 +166,19 @@ def build_parameters(self, stack, provider_stack=None):
         parameters = self._handle_missing_parameters(parameters,
                                                      required_params,
                                                      provider_stack)
-        return [
+
+        # Handle values that shouldn't be sent ('UsePreviousValue')
+        final_list = []
+        for p in [
```
mhahn: can we break this up to something like:

```
final_list = []
aws_params = [{"ParameterKey" ... for p in parameters]
for parameter in aws_params:
    ...
```

i'm not a fan of the list comp within the for statement

stacker/actions/build.py
```diff
@@ -166,10 +166,19 @@ def build_parameters(self, stack, provider_stack=None):
...
             {'ParameterKey': p[0],
-             'ParameterValue': str(p[1])} for p in parameters
-        ]
+             'ParameterValue': str(p[1])} for p in parameters]:
+            stack_p = stack.blueprint.parameters[p['ParameterKey']]
+            if p['ParameterValue'] == '****' and stack_p.properties['NoEcho']:
```
mhahn: is this always equal to `****`?

troyready:
@mhahn after testing this again today I realized that I had done this before converting to the new Variables which means it needs to be rethought a bit.

I'd still like to get a form of this implemented, but I'm not sure the cleanest way to do that and not mess up the new variables validation (e.g. (here)[https://github.com/remind101/stacker/blob/0.8.1/stacker/blueprints/base.py#L219]).

Curious if you can think of a good option to pursue? I'm not sure what kind of solution I'd prefer. A new object type? Some new key/list designating variables to not send in the yaml? A magic string that bypass the validation/sending (j/k)?

troyready:
Closing for now -- to be revisited after 1.0 release

PR #239, opened 2016-10-06
fix file handler lookups with pseudo parameters

troyready:
If the file lookup includes a reference with `::` in it (e.g.
`AWS::Region`), the regex currently doesn't extract it.


troyready:
@phobologic good thought -- I've added a new commit expanding this lookups' test to explicitly test what should & shouldn't match.

PR #243, opened 2016-10-12
Fix dump command unresolved variables error

mchristen:
My quick and easy way of fixing the following error while trying to dump the stacks

```
stacker.exceptions.UnresolvedVariables: Blueprint: "zone_test" hasn't resolved it's variables
```

### DIFF
--- a/stacker/actions/build.py
+++ b/stacker/actions/build.py
@@ -313,7 +313,7 @@ class Action(BaseAction):
             if outline:
                 plan.outline()
             if dump:
-                plan.dump(dump)
+                plan.dump(dump, self.context, self.provider)
 
     def post_run(self, outline=False, *args, **kwargs):
         """Any steps that need to be taken after running the action."""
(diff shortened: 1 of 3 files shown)


mhahn:
I think we need a better idea of what we want "dump" to accomplish. I thought of `dump` as essentially a "dry run" since it doesn't launch any stacks, just iterates through the plan and manually sets the status of each step to `COMPLETE`. Right now, this PR will only work if the stack has been brought up before, otherwise the output values won't exist for the provider to lookup.

I opted to not allow any access to the provider from `dump`. If you had lookup values, you could pass in dummy values with the `-var` option on the command line. This would let you run it at any stage: pre-creation, post-creation etc

I've only used dump a couple of times to debug blueprints, which is why the dummy value seemed acceptable to me, but that probably breaks down with a larger template.

mchristen:
Make sense, after I did this I discovered your 1.0 branch and saw that this was fixed on there already anyways...

PR #245, opened 2016-10-13
Support "allowed_values" within variable definitions

mhahn:


stacker/blueprints/base.py
```diff
@@ -179,6 +200,11 @@ def resolve_variable(var_name, var_def, provided_variable, blueprint_name):
     var_type = var_def.get("type")
     value = validate_variable_type(var_name, var_type, value)
 
+    allowed_values = var_def.get('allowed_values')
+    valid = validate_allowed_values(allowed_values, value)
+    if not valid:
+        raise UnallowedValue(var_name, value, allowed_values)
```
phobologic: Do you think we need a special exception here?  Seems like a `ValueError`, though I guess at the point where we move to all stacker exceptions being from a base exception this might be useful.

stacker/exceptions.py
```diff
@@ -128,3 +128,13 @@ def __init__(self, variable, validator, value, exception=None):
 
     def __str__(self):
         return self.message
+
+
+class UnallowedValue(Exception):
+    """Exception raised when the value provided to a variable is not allowed.
+    """
+    def __init__(self, variable, value, allowed_values, *args, **kwargs):
```
phobologic: This might be hard to find, since it doesn't list what blueprint it's a part of.  I actually am starting to feel like we should be doing more with catching exceptions at the blueprint, plan or stack level to provide more context, rather than needing to pass that context down to each exception in order to figure out where things are breaking.

stacker/blueprints/base.py
mhahn: Yea, I was just using the `UnallowedValue` error to construct the message which probably isn't worth it

stacker/exceptions.py
mhahn: Yea, a base exception that includes blueprint and some other data would be useful. I'm going to switch this to a `ValueError` for now and we can address this with the issue to track better exception handling

PR #252, opened 2016-10-20
Add hook for uploading AWS Lambda functions 

danielkza:
Add a hook to upload AWS Lambda payloads to S3, so that complex code can be easily
used in templates.

The payload ZIP is built from a base folder with a list of
include/exclude patterns (using the formic library), then uploaded to S3
to either the default stacker bucket, or a custom bucket. The uploads
contain the MD5 of the payload in their key, to allow idempotent
deployments - unchanged payloads are not uploaded multiple times.

Usage examples for the YAML config and blueprint code have been added.

### DIFF
--- a/stacker/hooks/aws_lambda.py
+++ b/stacker/hooks/aws_lambda.py
@@ -273,8 +273,10 @@ def _upload_function(s3_conn, bucket, name, options):
                     absolute, it will be interpreted as relative to the stacker
                     configuration file directory, then converted to an absolute
                     path. See :func:`stacker.util.get_config_directory`.
-                * include: file patterns to include in the payload (optional).
-                * exclude: file patterns to exclude from the payload (optional).
+                * include:
+                    file patterns to include in the payload (optional).
+                * exclude:
+                    file patterns to exclude from the payload (optional).
 
     Returns:
         troposphere.awslambda.Code: CloudFormation AWS Lambda Code object,
@@ -343,28 +345,28 @@ def upload_lambda_functions(region, namespace, mappings, parameters,
 
                     Files in this directory will be added to the payload ZIP,
                     according to the include and exclude patterns. If not
-                    patterns are provided, all files in this directory (respecting
-                    default exclusions) will be used.
+                    patterns are provided, all files in this directory
+                    (respecting default exclusions) will be used.
 
                     Files are stored in the archive with path names relative to
                     this directory. So, for example, all the files contained
                     directly under this directory will be added to the root of
                     the ZIP file.
                 * include(str or list[str], optional):
-                    Pattern or list of patterns of files to include in the payload.
-                    If provided, only files that match these patterns will
-                    be included in the payload.
+                    Pattern or list of patterns of files to include in the
+                    payload. If provided, only files that match these
+                    patterns will be included in the payload.
 
-                    Omitting it is equivalent to accepting all files that are not
-                    otherwise excluded.
+                    Omitting it is equivalent to accepting all files that are
+                    not otherwise excluded.
                 * exclude(str or list[str], optional):
                     Pattern or list of patterns of files to exclude from the
                     payload. If provided, any files that match will be ignored,
                     regardless of whether they match an inclusion pattern.
 
-                    Commonly ignored files are already excluded by default, such as
-                    ``.git``, ``.svn``, ``__pycache__``, ``*.pyc``, ``.gitignore``,
-                    etc.
+                    Commonly ignored files are already excluded by default,
+                    such as ``.git``, ``.svn``, ``__pycache__``, ``*.pyc``,
+                    ``.gitignore``, etc.
 
     Examples:
         .. Hook configuration.
--- a/stacker/hooks/aws_lambda.py
+++ b/stacker/hooks/aws_lambda.py
@@ -131,7 +131,7 @@ def _head_object(s3_conn, bucket, key):
     """Retrieve information about an object in S3 if it exists.
 
     Args:
-        s3_conn (.S3.Client): S3 connection to use for operations.
+        s3_conn (botocore.client.S3): S3 connection to use for operations.
         bucket (str): name of the bucket containing the key.
         key (str): name of the key to lookup.
 
@@ -156,7 +156,7 @@ def _ensure_bucket(s3_conn, bucket):
     """Create an S3 bucket if it does not already exist.
 
     Args:
-        s3_conn (.S3.Client): S3 connection to use for operations.
+        s3_conn (botocore.client.S3): S3 connection to use for operations.
         bucket (str): name of the bucket to create.
 
     Returns:
@@ -189,7 +189,7 @@ def _upload_code(s3_conn, bucket, name, contents):
     expected contents.
 
     Args:
-        s3_conn (.S3.Client): S3 connection to use for operations.
+        s3_conn (botocore.client.S3): S3 connection to use for operations.
         bucket (str): name of the bucket to create.
         name (str): desired name of the Lambda function. Will be used to
             construct a key name for the uploaded file.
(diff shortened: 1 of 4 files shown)


stacker/hooks/aws_lambda.py
```diff
@@ -0,0 +1,318 @@
...
+
+
+def _upload_code(s3_conn, bucket, name, contents):
+    """
+    Upload a ZIP file to S3 for use by Lambda.
+
+    The file will be stored with it's MD5 appended to the key, to allow
+    avoiding repeated uploads in subsequent runs with unchanged content.
+
+    Returns a troposphere.aws_lambda.Code object containing the bucket and key
```
phobologic: `aws_lambda` -> `awslambda`

danielkza:
@phobologic 

> I'm a little nervous about allowing the hooks themselves to chose top-level keys for the hook_data dictionary (though I like the concept of the hook_data dict itself) - but I don't have a good idea how we could make that better.

I initially namespaced by the last component of the path name, but it didn't seem good to me: it would mean there is no way to easily have two hooks implement the same functionality in a different manner, or to replace a default hook with a customized version. It was also quite cumbersome to do:

    self.context.hook_data['upload_lambda_functions']['MyFunction']

> Also, can you add some docs about the new hook? This is great stuff - thanks!

Anything particular you are looking for? Most of the documentation is the `upload_lambda_functions` docstrings, and I added it to the doc. index together with the other hooks.

stacker/hooks/aws_lambda.py
danielkza: Fixed.

stacker/hooks/aws_lambda.py
```diff
@@ -0,0 +1,318 @@
...
+
+
+def _find_files(root, includes, excludes):
+    """
+    Generate a list of files relative to a root path, applying inclusion and
+    exclusion patterns. The documentation for the patterns can be found at:
+    http://www.aviser.asia/formic/doc/index.html
+    """
+
+    root = os.path.abspath(root)
```
danielkza: Done. I didn't end up using `full_path` because I don't want to call `get_config_directory` unless the path is relative, as that would force me to mock it in all the tests. I just call `expanduser` right when I retrieve the config key.

stacker/hooks/aws_lambda.py
```diff
@@ -0,0 +1,431 @@
...
+    if not os.path.isabs(root):
+        root = os.path.abspath(os.path.join(get_config_directory(), root))
+    zip_contents = _zip_from_file_patterns(root, includes, excludes)
+
+    return _upload_code(s3_conn, bucket, name, zip_contents)
+
+
+def upload_lambda_functions(region, namespace, mappings, parameters,
+                            context=None, **kwargs):
+    """Builds Lambda payloads from user configuration and uploads then to S3.
```
ejholmes: Little typo here: `s/then/them`

stacker/hooks/aws_lambda.py
```diff
@@ -166,12 +259,31 @@ def _check_pattern_list(patterns, key, default=None):
...
-    """
-    Processes and uploads a complete Lambda payload to S3.
-
-    The `path` option is mandatory, and can be a relative or absolute path.
-    In the former case, it is interpreted relative to the stacker configuration
-    path.
+    """Builds a Lambda payload from user configuration and uploads it to S3.
+
+    Args:
+        s3_conn (.S3.Client): S3 connection to use for operations.
```
phobologic: Should this be boto.S3.Client?

stacker/hooks/aws_lambda.py
```diff
@@ -100,7 +153,20 @@ def _head_object(s3_conn, bucket, key):
 
 
 def _ensure_bucket(s3_conn, bucket):
-    """Create an S3 bucket if it doesn't already exist"""
+    """Create an S3 bucket if it does not already exist.
+
+    Args:
+        s3_conn (.S3.Client): S3 connection to use for operations.
```
phobologic: same here - botocore.S3.Client?  not sure what the actual path should look like

danielkza: I was actually experimenting with cross-referencing to the boto docs, which seem to be quite broken - they don't prefix any of the classes with the correct package in the indexes.

I think the correct type is actually `botocore.client.S3`.

phobologic:
@mhahn @danielkza Any comment on my previous comment?

```
Yeah, it almost makes me think that we should have given hooks a name or namespace property that would have allowed us to namespace this stuff naturally. I suppose we could add something like that and make it optional, and not do anything with the return value of the hook unless it comes from a hook with the name. What do you think about that?
```

I'm even considering just breaking the old API with this, since we're moving into 1.0.  It seems like the right time to make a change like this.

danielkza:
Adding a key to the hook configuration could make sense. Maybe call it `data_key` or something of the sort? That makes it clear what the purpose is, since a name would seem redundant when the path is already specified. Then it can be used to namespace the hook data.

I personally would prefer that the API didn't break, as am I selfishly and lazily delaying migrating my stacks to the new 1.0 parameter. It also doesn't seem strictly necessary, as ignoring the return values of hooks that are not properly tagged (or even just storing `True`/`False`) should not break anything.

Having the hook data be a flat dictionary has some advantages though: there's no need to merge data that comes from different hooks that happen to have the same key (as the hook itself can decide whether to override something or not), and it's easy to have a hook replace/complement another one.

It might be wise to try to think of a second use case other than the Lambda one, to figure out which solution works best overall.

stacker/hooks/aws_lambda.py
danielkza: Fixed now, sorry for the delay.

stacker/hooks/aws_lambda.py
danielkza: Fixed.

danielkza:
I rebased to the latest master and fixed some typos. I'm not really sure what's the best way to proceed about the hook API dilemma, but please feel free to make/suggest any changes from my first attempt.

phobologic:
Ok, I've thought about it some more, and I think we should go with the idea I had before:

```
Yeah, it almost makes me think that we should have given hooks a name or namespace property that would have allowed us to namespace this stuff naturally. I suppose we could add something like that and make it optional, and not do anything with the return value of the hook unless it comes from a hook with the name. What do you think about that? @mhahn any thoughts?
```

I'm fine with using `data_key` in place of `name` or `namespace`, but I think that we should ignore any returned data on a hook that doesn't have an identifier.  We can still use the idea that returning anything that evaluates to True means a positive run, and anything that returns a value that is False (or raises an exception) is broken.

Sorry for the delay again, @danielkza, but could you take a stab at this?  If not, I can go ahead and write a separate PR that creates the basics of the new system.

One thing we should consider is what to do if two hooks have the same `data_key` - my guess is it would just overwrite at this point.  Or we could throw an error - assuming that it's a mistake more often than not.

danielkza:
@phobologic I think it would be better if you implemented the hook data system first as you find best, then I can fix my code to adapt to it afterwards.

phobologic:
@danielkza sounds good, I'll work on that now.  Thanks for your patience!

phobologic:
@danielkza @mhahn check out #270 - one design decision I made was to disallow duplicate data_keys.

phobologic:
Thanks @danielkza - I've gone ahead and merged #270.

danielkza:
@phobologic  Rebased to the latest master with the update to match the final API.

phobologic:
This looks good @danielkza - thanks!  Can you do one more pass and take a look at all the logger calls?  It looks like they're all at `info`, which may be a bit noisy.  I'm going to leave it up to your judgement though if that's the right thing to do or not.  Thanks again!

danielkza:
@phobologic I changed printing the ZIP hash and list of files to the debug level, see if it helps.

phobologic:
Thanks @danielkza !

PR #253, opened 2016-10-24
Allow use of later boto3

jonathanunderwood:
This PR just broadens the range of required boto3 to include the 1.4.n release series.


phobologic:
Looks good to me - thanks!

PR #257, opened 2016-10-29
Troposphere types

mhahn:
Adds a new custom variable type: `TroposphereType`. This builds off @phobologic's improvements to `from_dict` within Troposphere. `TroposphereType` lets you pass through the yaml from a config directly to troposphere types which saves you the boilerplate of specifying variables within the blueprint or instantiating the nested types yourself.

Open to suggestions about naming or other syntax.

An example:

**config file**:

```
- name: myBuckets
  class_path: stacker_blueprints.s3.Buckets
  variables:
    Buckets:
      - BucketName: my-bucket
      - BucketName: my-other-bucket
```

**stacker_blueprints.s3.Buckets**:

```
from stacker.blueprints.base import Blueprint
from stacker.blueprints.variables.types import TroposphereType
from troposphere import s3


class Buckets(Blueprint):

    VARIABLES = {
        "Buckets": {
            "type": TroposphereType([s3.Bucket]),
            "description": "s3 buckets to create",
        },
    }

    def create_template(self):
        t = self.template
        [t.add_resource(bucket) for bucket in variables["Buckets"]]
```


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## TBD
+
+- Add support for `TroposphereType`
```
phobologic: I'd just move this into the 1.0.0a1 section - I plan to just update that to be the 1.0.0 release when we're ready to go.

docs/blueprints.rst
```diff
@@ -95,6 +95,43 @@ Variable Types
...
+
+
+  class Bucket(Blueprint):
+
+      VARIABLES = {
+          "Buckets": {
+              # Specify that Buckets will be a list of s3.Bucket types.
+              This means the config should take a list of dictionaries
+              which will be converted into troposphere buckets.
+              "type": TroposphereType([s3.Bucket]),
```
phobologic: Is this, and a single type, the only applicable syntax for `TroposphereType`?  If so, probably worth documenting both.

stacker/blueprints/variables/types.py
```diff
@@ -1,15 +1,79 @@
...
+    def __init__(self, defined_type, resource_name=None):
+        """Represents a Troposphere type.
+
+        :class:`TroposphereType` will convert the value for the variable to the
+        given Troposphere type.
+
+        Args:
+            defined_type (Union[list, type]): List of or single Troposphere
+                type
+            resource_name (str): The name to use for the resource when creating
```
phobologic: Rather than this, what about making it so that every resource should have a `title` attribute that is required?  That's the name of the variable that is used in troposphere.  Another option is that we could do away with both the list & plain type syntax, and instead replace it with a single format that uses dictionaries like I did with the SecurityGroup blueprint: https://github.com/remind101/stacker_blueprints/pull/56/commits/ab094c32b1bcea2978796d6118bf43365571f2b7#diff-e1bf8fc9a0d5fa2e2b507057d3fcd3aaR29

phobologic: I think the way I did it in the security group blueprint actually makes a lot of sense.  You have to give unique names to all resources in a blueprint, so you might as well make it a requirement in the config.  Going to hold off on further review till you reply to this.

mhahn: good call

mhahn:
@phobologic updated based on your feedback

docs/blueprints.rst
```diff
@@ -122,16 +122,44 @@ Below is an annotated example::
...
+
+A sample config for the above::
+
+  stacks:
+    - name: buckets
+      class_path: path.to.above.Buckets
+      variables:
+        Buckets:
+          # resource name that will be added to CloudFormation
+          FirstBucket:
```
phobologic: Honestly, at this point, my only concern is the fact we call it a `list` type, but when you define it in the config its actually a dict.  Not sure if this is a big deal or not - dict makes more sense from the config perspective, but maybe making it a list of single item dicts would bridge the gap:

```
Buckets:
    - FirstBucket:
        BucketName: my-first-bucket
```

What do you think?

phobologic: Hmm, the other option is to step away from the python types altogether and instead go with our own syntax.  `TroposphereType` `ManyTroposphereType` or something?

mhahn: We talked about this on slack, but we decided we should do `TroposphereType(<type>, many=True)`

mhahn:
@phobologic i think this is good to go

stacker/blueprints/variables/types.py
```diff
@@ -42,22 +35,17 @@ def create(self, value):
...
-                message = (
-                    "Only one resource can be provided for single "
-                    "TroposphereType"
-                )
-                raise ValueError(message)
-
-        output = [new_type.from_dict(title, values) for title, values in
+        if not self._many and len(value) > 1:
+            message = (
+                "Only one resource can be provided for single "
```
phobologic: Is there anyway we can make this error more helpful in determining where the issue is?  Maybe we need to catch this higher and point out what caused this?

phobologic: NM, just realized this is caught and re-raised as a ValidatorError, which will point out where the issue stems from.  Awesome, going to go ahead and merge this.

PR #260, opened 2016-11-02
Add currently working directory to path

chrishenry:
This allows stacker to be called from other directories containing the standard blueprints / conf folder structure.

### DIFF
--- a/stacker/util.py
+++ b/stacker/util.py
@@ -140,6 +140,10 @@ def load_object_from_string(fqcn):
     load_object_from_string("logging.Logger")
     load_object_from_string("LocalClassName")
     """
+
+    import os, sys
+    sys.path.insert(0, os.getcwd())
+
     module_path = "__main__"
     object_name = fqcn
     if "." in fqcn:


phobologic:
Hey @chrishenry, thanks for this, but we already have the ability for someone to add their CWD to their sys path.  You can see that here: http://stacker.readthedocs.io/en/latest/config.html#module-paths

chrishenry:
👍 Thanks!

Out of curiosity, why isn't `cwd` added to the path by default? I can imagine a lot of conf files have the same line to add `cwd` to `sys_path`.

phobologic:
I personally don't like adding the local path automatically - it could be viewed as a security risk, and has some pretty scary implications if someone were to exploit it.  For many people the benefit may outweigh the risk,and it's very likely that I'm being paranoid, but that's why I like making it easy to add - just not automatic and enforced for everyone.

(@ Remind we don't add it, for example, and do everything out of a docker container where our private stacks get installed in the regular path)

PR #261, opened 2016-11-04
Add support for getting info on stacker blueprints

mwildehahn:
This depends on the `troposphere-type` branch since its most useful with that.

I wanted to get this out there for feedback, right now it is pretty simple but I think will make people's lives much easier when working with the new troposphere types.

Any input on how the output should look is useful. I'm going to look into how argparse formats long descriptions of variables and try and do something similar, I just went with what was fastest for now.

I think this could get super useful if we pulled in cloudformation variable descriptions from the help site: http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-s3-bucket.html#cfn-s3-bucket-lifecycleconfig. I'm not sure if there is a way to do that right now, it could be a useful thing to add to troposphere if we can figure out how to automate it, then we could get it here mostly for free.

Here are some sample outputs:

**output info for an old school blueprint**
```
$ stacker blueprint info stacker_blueprints.vpc.VPC

VPC

Variables:
	- BaseDomain (CFN::String): Base domain for the stack.
	- CidrBlock (CFN::String): Base CIDR block for subnets.
	- ImageName (CFN::String): The image name to use from the AMIMap (usually found in the config file.)
	- UseNatGateway (CFN::String): If set to true, will configure a NAT Gatewayinstead of NAT instances.
	- InternalDomain (CFN::String): Internal domain name, if you have one.
	- AZCount (int)
	- PublicSubnets (CFN::CommaDelimitedList): Comma separated list of subnets to use for public hosts. NOTE: Must have as many subnets as AZCount
	- SshKeyName (CFN::AWS::EC2::KeyPair::KeyName)
	- InstanceType (CFN::String): NAT EC2 instance type.
	- PrivateSubnets (CFN::CommaDelimitedList): Comma separated list of subnets to use for non-public hosts. NOTE: Must have as many subnets as AZCount
```

**output info for new troposphere types**
```
$ stacker blueprint info stacker_blueprints.s3.Buckets

Buckets

Variables:
	- Buckets ([troposphere.s3.Bucket]): S3 Buckets to create
```

where it gets really useful now is being able to dig into what fields the `troposphere.s3.Bucket` takes:

```
$ stacker blueprint info stacker_blueprints.s3.Buckets:Buckets

Buckets ([troposphere.s3.Bucket]): S3 Buckets to create

Fields:
- AccessControl (basestring)
- LoggingConfiguration (troposphere.s3.LoggingConfiguration)
- LifecycleConfiguration (troposphere.s3.LifecycleConfiguration)
- BucketName (troposphere.validators.s3_bucket_name)
- CorsConfiguration (troposphere.s3.CorsConfiguration)
- Tags (troposphere.Tags)
- ReplicationConfiguration (troposphere.s3.ReplicationConfiguration)
- WebsiteConfiguration (troposphere.s3.WebsiteConfiguration)
- NotificationConfiguration (troposphere.s3.NotificationConfiguration)
- VersioningConfiguration (troposphere.s3.VersioningConfiguration)
```

and you can continue to dig in by making the path after the `:` more specific:
```
$ stacker blueprint info stacker_blueprints.s3.Buckets:Buckets.LifecycleConfiguration

LifecycleConfiguration (troposphere.s3.LifecycleConfiguration)

Fields:
- Rules ([troposphere.s3.LifecycleRule], required)

$ stacker blueprint info stacker_blueprints.s3.Buckets:Buckets.LifecycleConfiguration.Rules

Rules (troposphere.s3.LifecycleRule)

Fields:
- Status (basestring, required)
- Prefix (basestring)
- Transitions ([troposphere.s3.LifecycleRuleTransition])
- ExpirationDate (basestring)
- NoncurrentVersionExpirationInDays (troposphere.validators.positive_integer)
- ExpirationInDays (troposphere.validators.positive_integer)
- NoncurrentVersionTransition (troposphere.s3.NoncurrentVersionTransition)
- Transition (troposphere.s3.LifecycleRuleTransition)
- Id (basestring)
```

**TODO**
- [ ] support listing out blueprints if given a directory: ie `stacker blueprint info stacker_blueprints`
- [ ] support listing out blueprints if given a file: ie. `stacker blueprint info stacker_blueprints.s3`
- [ ] better error handling if given wrong paths etc.
- [ ] list out required variables at the top and denote they're required
- [ ] display default values for variables


stacker/blueprints/info.py
```diff
@@ -0,0 +1,244 @@
...
+
+    Returns:
+        str: formatted string describing the field for the given variable.
+
+    """
+    var_type = variable["type"]
+    # TODO does this handle cases where we don't have a troposphere type here?
+    troposphere_type = get_troposphere_type(var_type)
+
+    while field_path:
```
phobologic: Why not `for prop_name in field_path:` here instead of the while + pop method?

stacker/blueprints/info.py
```diff
@@ -0,0 +1,244 @@
...
+
+    """
+    is_list = False
+    if isinstance(var_type, TroposphereType):
+        if var_type.many:
+            is_list = True
+
+        var_type = var_type.type
+
+    if isinstance(var_type, (list, tuple)) and len(var_type):
```
phobologic: Not sure I understand why the `len(var_type)` is here?  What's case is this trying to deal with?  Can the var_type be an empty list?

stacker/blueprints/info.py
```diff
@@ -0,0 +1,244 @@
...
+
+    """
+    var_type = variable["type"]
+    # TODO does this handle cases where we don't have a troposphere type here?
+    troposphere_type = get_troposphere_type(var_type)
+
+    while field_path:
+        prop_name = field_path.pop(0)
+        prop_type, required = troposphere_type.props[prop_name]
+        if isinstance(prop_type, list) and len(prop_type) == 1:
```
phobologic: What's the case when the length of prop_type will be anything other than 1?

stacker/blueprints/info.py
```diff
@@ -0,0 +1,244 @@
...
+            prop_type = prop_type[0]
+
+        if hasattr(prop_type, "props"):
+            troposphere_type = prop_type
+
+    fields = ""
+    description = [format_var_type(prop_type)]
+    if not hasattr(prop_type, "props"):
```
phobologic: This is all getting a little complicated - maybe it'd be good to break it up into smaller functions, or possibly just throw some comments in here?

PR #266, opened 2016-11-21
Remove transformation of variables

### LINKED ISSUE #255: Command line parsing of variables breaks for lists variables
In a blueprint, I have part of the VARIABLES attribute defined as:

``` python
        'SecurityGroups': {
            'type': list,
            'description': 'List of Security Groups for the instance'
        },
```

Calling stacker like this:

```
stacker build -r eu-west-1 -t -var 'SecurityGroups=sg-0d60076a' <env> <conf>
```

Results in a CloudFormation template containing:

```
                "SecurityGroupIds": [
                    "s",
                    "g",
                    "-",
                    "0",
                    "d",
                    "6",
                    "0",
                    "0",
                    "7",
                    "6",
                    "a"
]
```

So, somehow the security group id has become tokenized. It's unclear from the docs/help how -var is supposed to be used, but I don't think this is intended behavior?

phobologic:
This stops the type checker from transforming variables, making it more
strict. Magically changing the variable type was a little janky to begin
with, so I think this is the right thing to do.

Should partially fix #255


phobologic:
@mhahn any thoughts on this?

mhahn:
If we're going to do this, I think we should prevent any transformations of the variable. ie. `validate_variable_type` and use something more like this: https://github.com/remind101/stacker/pull/266/files#diff-9e12e7feca740fe22caedce88b419cbcR199. I think the main driver for `validate_variable_type` was so that we could transform to troposphere types, but since we should have native support for those soon, i don't think it is necessary.

mhahn:
Right now you could still transform the value with: https://github.com/remind101/stacker/pull/266/files#diff-9e12e7feca740fe22caedce88b419cbcR196

phobologic:
Ok, @mhahn, I think this should be good to go now.  Let me know what you think.

PR #276, opened 2016-11-29
Relax troposphere dependency

ejholmes:
This could probably be relaxed even more, since there's nothing that stacker explicitly depends on in troposphere 1.9.

stacker_blueprints still has a troposphere dependency on ~=1.8.1, which makes them incompatible with each other right now.


phobologic:
👍 This looks good to me.

phobologic:
Closing since stacker_blueprints 64 should fix it.

PR #278, opened 2016-11-29
Support destroying a subset of stacks

rouge8:
This patch adds a `--stacks` argument to `stacker destroy` to enable destroying only some of the stacks, similar to `stacker build --stacks`.

This is useful if you have one config file for several similar stacks, rather than just interdependent stacks.


phobologic:
Thanks @rouge8 - looks good!

PR #280, opened 2016-12-01
DAG

ejholmes:
This is an internal refactor of `stacker.plan.Plan` to use a DAG to internally represent steps and their dependencies. The net result provides a simpler method to walk the graph, and update stacks in order of their dependencies.

Currently, this doesn't walk the graph in parallel, which is implemented in https://github.com/remind101/stacker/pull/281.

The DAG implementation is pulled from https://github.com/thieman/py-dag with the addition of a `walk` method for walking the graph in topological order.

**TODO**

* [x] Better reverse walk.

### DIFF
--- a/stacker/actions/build.py
+++ b/stacker/actions/build.py
@@ -1,6 +1,6 @@
 import logging
 
-from .base import BaseAction
+from .base import BaseAction, cancel
 from .. import exceptions, util
 from ..exceptions import StackDidNotChange
 from ..plan import Plan
@@ -292,7 +292,10 @@ class Action(BaseAction):
         if not outline and not dump:
             plan.outline(logging.DEBUG)
             logger.debug("Launching stacks: %s", ", ".join(plan.keys()))
-            plan.execute(self._launch_stack)
+            plan.execute(
+                self._launch_stack,
+                parallel=self.provider.supports_parallel,
+                cancel=cancel())
         else:
             if outline:
                 plan.outline()
--- a/stacker/actions/build.py
+++ b/stacker/actions/build.py
@@ -271,25 +271,9 @@ class Action(BaseAction):
         return params.items()
 
     def _generate_plan(self, tail=False):
-        plan_kwargs = {}
-        if tail:
-            plan_kwargs["watch_func"] = self.provider.tail_stack
-        plan = Plan(description="Create/Update stacks", **plan_kwargs)
-        stacks = self.context.get_stacks_dict()
-        dependencies = self._get_dependencies()
-        for stack_name in self.get_stack_execution_order(dependencies):
-            plan.add(
-                stacks[stack_name],
-                run_func=self._launch_stack,
-                requires=dependencies.get(stack_name),
-            )
+        plan = Plan(description="Create/Update stacks")
+        plan.build(self.context.get_stacks())
         return plan
-
-    def _get_dependencies(self):
-        dependencies = {}
-        for stack in self.context.get_stacks():
-            dependencies[stack.fqn] = stack.requires
-        return dependencies
 
     def pre_run(self, outline=False, *args, **kwargs):
         """Any steps that need to be taken prior to running the action."""
@@ -308,7 +292,7 @@ class Action(BaseAction):
         if not outline and not dump:
             plan.outline(logging.DEBUG)
             logger.debug("Launching stacks: %s", ", ".join(plan.keys()))
-            plan.execute()
+            plan.execute(self._launch_stack)
         else:
             if outline:
                 plan.outline()
--- a/stacker/dag/__init__.py
+++ b/stacker/dag/__init__.py
@@ -74,7 +74,20 @@ class DAG(object):
             raise KeyError('this edge does not exist in graph')
         graph[ind_node].remove(dep_node)
 
-    def walk(self, walk_func, graph=None, reverse=True):
+    def transpose(self, graph=None):
+        """ Builds a new graph with the edges reversed. """
+        if not graph:
+            graph = self.graph
+        transposed = DAG()
+        for node, edges in graph.items():
+            transposed.add_node(node)
+        for node, edges in graph.items():
+            # for each edge A -> B, transpose it so that B -> A
+            for edge in edges:
+                transposed.add_edge(edge, node)
+        return transposed
+
+    def walk(self, walk_func, graph=None):
         """ Walks each node of the graph in reverse topological order.
 
         This can be used to perform a set of operations, where the next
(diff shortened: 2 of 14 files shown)


ejholmes:
Ran this against our staging environment without issue. Gonna take a look at adding a `walk_parallel` method to the DAG.

stacker/plan.py
```diff
@@ -97,174 +86,79 @@ def submit(self):
...
-        wait_func (func, optional): the function to be called after each pass
-            of running stacks. This defaults to :func:`time.sleep` and will
-            sleep for the given `sleep_time` before starting the next pass.
-            Default: :func:`time.sleep`
-
-    """
-
-    def __init__(self, description, sleep_time=5, wait_func=None,
-                 watch_func=None, *args, **kwargs):
+class Plan():
```
phobologic: this should be `class Plan(object):`

stacker/plan.py
```diff
@@ -97,174 +86,79 @@ def submit(self):
...
+    def __init__(self, description, reverse=False, sleep_func=sleep):
```
phobologic: Probably worth adding back the doc string here.

stacker/plan.py
```diff
@@ -97,174 +86,79 @@ def submit(self):
...
-    def list_status(self, status):
-        """Returns a list of steps in the given status.
-
-        Args:
-            status (:class:`Status`): The status to match steps against.
-
-        Returns:
-            list: A list of :class:`Step` objects that are in the given status.
+    def build(self, stacks):
+        """ Builds an internal dag from the stacks and their dependencies """
```
phobologic: Add the full docstring here for the API docs.

stacker/plan.py
```diff
@@ -97,174 +86,79 @@ def submit(self):
...
+            dag.add_node(fqn)
+            self._steps[fqn] = Step(
+                stack=stack)
+
+        for stack in stacks:
+            for dep in stack.requires:
+                try:
+                    dag.add_edge(stack.fqn, dep)
+                except DAGValidationError:
+                    raise CyclicDependencyError(stack.fqn)
```
phobologic: You should add the dependency to this exception as well, just so it's easier to see what dependency caused the error.

stacker/plan.py
```diff
@@ -97,174 +86,79 @@ def submit(self):
...
+
+        self._dag = dag
+        return None
```
phobologic: You don't have to do this - 

```
>>> def blah():
...   print "WOO"
...
>>>
>>> x = blah()
WOO
>>> x
>>> type(x)
<type 'NoneType'>
```

stacker/dag/__init__.py
```diff
@@ -0,0 +1,221 @@
...
+            nodes.reverse()
+        for n in nodes:
+            walk_func(n)
+
+    def rename_edges(self, old_task_name, new_task_name, graph=None):
+        """ Change references to a task in existing edges. """
+        if not graph:
+            graph = self.graph
+        for node, edges in graph.items():
+
```
phobologic: Weird blank line here.

stacker/dag/__init__.py
```diff
@@ -0,0 +1,221 @@
...
+            graph = self.graph
+
+        dependent_nodes = set(
+            node for dependents
+            in graph.itervalues() for node in dependents)
+        return [node for node in graph.keys() if node not in dependent_nodes]
+
+    def validate(self, graph=None):
+        """ Returns (Boolean, message) of whether DAG is valid. """
+        graph = graph if graph is not None else self.graph
```
phobologic: This is different than the way it's done throughout this class - might be worth changing it just for consistency.

ejholmes:
Ok. Think I got all the changes in based on feedback.

We should probably hold off on merging until https://github.com/remind101/stacker/pull/281 is ready, since this will be slower to start without a parallel walk.

stacker/dag/__init__.py
```diff
@@ -0,0 +1,224 @@
...
+        """ Walks each node of the graph in reverse topological order.
+
+        This can be used to perform a set of operations, where the next
+        operation depends on the previous operation. It's important to note
+        that walking happens serially, and is not paralellized.
+        """
+        if not graph:
+            graph = self.graph
+        nodes = self.topological_sort(graph=graph)
+        if reverse:
```
ejholmes: I think this is kinda a naive implementation for reversing a dag. Doing this, we can't walk the graph concurrently in reverse. Instead, we should do a depth first walk, and build a new graph with the edges reversed.

ejholmes: Added a commit that adds a `tranpose` method on DAG to build a new DAG with all the edges transposed. Plan uses this now for destroy actions, so it can be parallelized.

stacker/actions/build.py
```diff
@@ -271,26 +271,10 @@ def _handle_missing_parameters(self, params, required_params,
         return params.items()
 
     def _generate_plan(self, tail=False):
-        plan_kwargs = {}
-        if tail:
-            plan_kwargs["watch_func"] = self.provider.tail_stack
```
ejholmes: This gets removed in this PR, which may not be correct. This seems to tail the events, but I'm not sure I've ever actually seen stacker do this.

phobologic: Yeah - this is used by some people with `--tail`.  Does it cause issues if you leave it in?

stacker/tests/test_plan.py
```diff
@@ -60,20 +60,47 @@ def test_build_plan(self):
...
+        app = Stack(
+            definition=generate_definition(
+                'app', 1, requires=['namespace-db.1']),
             context=self.context)
 
         plan = Plan(description="Test", sleep_func=None)
 
-        with self.assertRaises(CyclicDependencyError):
-            plan.build([vpc, bastion])
+        try:
```
phobologic: I think what you want here is `self.assertRaises()`.  You can see an example of that here:

https://github.com/remind101/stacker/blob/6a141b6d9c429d89158ad7ea274cdfcfbdc6060d/stacker/tests/test_config.py#L15

phobologic:
Oh, also, once you submit this, could you open another PR to merge this against the `release-1.0` branch?

stacker/tests/test_plan.py
ejholmes: Oh sweet. I couldn't find an example online that used assertRaises to check the message. Much nicer.

ejholmes:
Alright, this has gotten behind `master` which is moving towards 1.0, and my original version of this was based on top of 0.8.5. I've updated this for 0.8.6, if anybody wants to use it ([dag-0.8.6 branch](https://github.com/remind101/stacker/tree/dag-0.8.6), much faster than 0.8.6), but it won't be merged into 0.8 stable.

I'll follow up with porting this forward to 1.0 once we've moved internally to 1.0.

Issue #279, opened 2016-11-30
General slowness


ejholmes:
I've noticed recently (I think it started with the 0.8+ releases) that stacker seems to run a lot slower. For a project containing ~60 stacks, a no-op run through took 6 minutes. Given that most of the stacks don't depend on each other (which means most updates should be running in parallel), 6 minutes seems incredibly high.

We should dig into why this is.

ejholmes:
On a side note, I've noticed that the plan status doesn't seem to be updated periodically anymore. Maybe related?

ejholmes:
Ok, I did a little digging and it looks like the logic in [stack.py](https://github.com/remind101/stacker/blob/0cf1df67b4ae5aeda5845442c84905909101c238/stacker/plan.py#L189-L230) for doing parallelization doesn't really work the way you'd expect it to. Essentially, it only spins up a separate process that watches for a stack to complete, but doesn't actually do the execution in a separate process, so a single stack update will block any other stack update from executing.

This also explains why the plan status is never updated, because `_single_run` is only called once, and the plan status is only output once for each `_single_run`.

Maybe another reason to do an internal refactor to use a DAG, which would make it a lot easier to walk the graph in parallel.

PR #281, opened 2016-12-01
Execute plan in parallel

ejholmes:
https://asciinema.org/a/37li2129w03khzxdqt4pacq77

This allows for parallel executions of plans, by walking each node in a thread.

Each thread blocks and waits for it's dependencies to complete, using a condition variable. When it executes it's walk function, it notifies other threads waiting on it so that they can continue.

Fixes https://github.com/remind101/stacker/issues/279. I tested this under the same conditions and it took ~2 minutes. Even that still seems to long to me, so there's probably some more areas that can be improved.

**TODO**

* [x] Exception handling in threads
* [x] Disable parallel walk in interactive mode.
* [x] Synchronization around calling `check_point`.
* [x] Cancel execution when a stack errors
* [x] Handle SIGINT

### DIFF
--- a/stacker/plan.py
+++ b/stacker/plan.py
@@ -114,8 +114,8 @@ class Plan(object):
         self._reverse = reverse
         self._sleep_func = sleep_func
         self.id = uuid.uuid4()
-        # Manages synchronization around calling `check_point`.
-        self._check_point_lock = threading.Lock()
+        # Manages synchronization around calling `fn` within `execute`.
+        self._lock = threading.Lock()
 
     def build(self, stacks):
         """ Builds an internal dag from the stacks and their dependencies.
@@ -156,6 +156,7 @@ class Plan(object):
         """
         check_point = self._check_point
         sleep_func = self._sleep_func
+        lock = self._lock
 
         check_point()
 
@@ -163,11 +164,14 @@ class Plan(object):
         # for managing the lifecycle of the step until completion.
         def step_func(step):
             while not step.done:
+                lock.acquire()
                 current_status = step.status
                 status = fn(step.stack, status=step.status)
                 step.set_status(status)
                 if status != current_status:
                     check_point()
+                lock.release()
+
                 if sleep_func and not step.done:
                     sleep_func()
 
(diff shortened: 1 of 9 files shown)


ejholmes:
One thing that's annoying about this is it produces a lot of flicker when `check_point` is called. Dunno what's up with that.

ejholmes:
Added a commit that allows providers to describe whether they support parallel execution of the plan or not. For the AWS interactive provider, parallel execution is disabled.

Also check_point is only called when the status of a step changes. Hopefully that will produce less flicker.

ejholmes:
Added a commit to handle cancelation if a stack update fails with an exception:

In the following, the dependency graph is:

```
c1 -> b1
c1 -> a
b1 -> a
b2 -> a
```

The execution order would be `a -> (b1 -> (c1), b2)` and b1 and b2 are set to a broken template. `a` executes, then `b1` and `b2` start executing, both of which error. `c1` is cancelled.

https://asciinema.org/a/cuus5bf9roxpwb1bbtzyju21n

ejholmes:
Added a commit to handle SIGTERM and SIGINT gracefully, so that threads are notified to cancel.

stacker/plan.py
```diff
@@ -224,9 +230,15 @@ def walk_func(fqn):
...
 
-        return dag.walk(walk_func)
+        walk = dag.walk
+        if parallel:
+            walk = dag.walk_parallel
+
+        return walk(walk_func, cancel=cancel)
 
     def _check_point(self):
+        lock = self._check_point_lock
```
ejholmes: I think this probably needs to be extended to lock calls to `fn` in `execute`. I ran into a case where I was using an aws profile with MFA, and got hit with 20+ prompts at once. Individual `fn` calls should be pretty quick, so holding the lock shouldn't be too bad.

phobologic:
👍 looks good - my guess is that we might be able to make this faster w/ the SNS+SQS stuff, but probably best to have that as another change.

ejholmes:
I'll go ahead and merge this into the dag branch, and we can play with it internally to work out any bugs.

Issue #186, opened 2016-08-04
Allow for sub-tree provisioning


ejholmes:
If we think about it, stacker configs are just a dependency tree (a DAG to be exact), where stacks depend on other stacks. It would be useful if you had the ability to start provisioning from an explicit root.

For example, say I have the following stacker config:

``` yaml
stacks:
  - name: vpc
  - name: bastion
    parameters:
      VpcId: vpc::Id
  - name: db
    parameters:
      VpcId: vpc::Id
  - name: api
    parameters:
      DatabaseUrl: db::Url
```

It would be handy to be able to specify the starting point for provisioning, like:

``` console
$ stacker build --root api
api -> db -> vpc
```

``` console
$ stacker build --root bastion
bastion -> vpc
```

I think this would be more powerful than the current `--stacks` flag, since that currently doesn't respect dependencies.

ejholmes:
Potentially related to https://github.com/remind101/stacker/issues/184?

mhahn:
i think this would be pretty straightforward. i'll take a crack at it

PR #282, opened 2016-12-02
Make --stacks flag build dependencies

ejholmes:
Closes https://github.com/remind101/stacker/issues/186

This allows the --stacks flag to support building the stacks dependencies, so you don't need to do it manually. For example, say you had this configuration:

```yaml
stacks:
  - name: a
    class_path: stacker.dev.blueprints.Dummy
  - name: b
    class_path: stacker.dev.blueprints.Dummy
    parameters:
      DummyParameter: a::DummyOutput
  - name: c
    class_path: stacker.dev.blueprints.Dummy
    parameters:
      DummyParameter: b::DummyOutput
```

Previously, if I wanted to build only `c`, I would have to specify `--stacks a --stacks b --stacks c`. With this change, you only need to specify `--stacks c` and stacker will build `a` and and `b` implicitly.

This depends on the DAG refactor in https://github.com/remind101/stacker/pull/280

### DIFF
--- a/stacker/actions/build.py
+++ b/stacker/actions/build.py
@@ -270,9 +270,9 @@ class Action(BaseAction):
 
         return params.items()
 
-    def _generate_plan(self, tail=False):
+    def _generate_plan(self, tail=False, stack_names=None):
         plan = Plan(description="Create/Update stacks")
-        plan.build(self.context.get_stacks())
+        plan.build(self.context.get_stacks(), stack_names=stack_names)
         return plan
 
     def pre_run(self, outline=False, *args, **kwargs):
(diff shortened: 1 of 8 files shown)


phobologic:
Looks good - it's a backwards incompatible change, but I don't think it's an unexpected one.  One thing - I just merged #278 and I think this will break that, can you merge that in here and make it so this works with `destroy` as well?

ejholmes:
Yep, totally. It makes conceptual sense to do the same for destroying. In the example above, if I want to destroy stack `b`, then I need to destroy stack `c` first since it depends on `b`.

ejholmes:
See https://github.com/remind101/stacker/pull/280#issuecomment-288599922. I'll follow up on this later.

PR #285, opened 2016-12-04
Update all hooks to use advanced results

phobologic:
This takes advantage of GH-270 for all hooks now.

### DIFF
--- a/stacker/hooks/ecs.py
+++ b/stacker/hooks/ecs.py
@@ -26,7 +26,9 @@ def create_clusters(region, namespace, mappings, parameters, **kwargs):
     if isinstance(clusters, basestring):
         clusters = [clusters]
 
+    cluster_info = {}
     for cluster in clusters:
         logger.debug("Creating ECS cluster: %s", cluster)
-        conn.create_cluster(clusterName=cluster)
-    return {"clusters": clusters}
+        r = conn.create_cluster(clusterName=cluster)
+        cluster_info[r['clusterName']] = r
+    return {"clusters": cluster_info}
--- a/stacker/hooks/ecs.py
+++ b/stacker/hooks/ecs.py
@@ -29,4 +29,4 @@ def create_clusters(region, namespace, mappings, parameters, **kwargs):
     for cluster in clusters:
         logger.debug("Creating ECS cluster: %s", cluster)
         conn.create_cluster(clusterName=cluster)
-    return True
+    return {"clusters": clusters}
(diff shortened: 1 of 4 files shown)


phobologic:
@danielkza - since you're probably the most familiar with this feature, any chance you could take a look?

stacker/hooks/ecs.py
```diff
@@ -29,4 +29,4 @@ def create_clusters(region, namespace, mappings, parameters, **kwargs):
     for cluster in clusters:
         logger.debug("Creating ECS cluster: %s", cluster)
         conn.create_cluster(clusterName=cluster)
-    return True
+    return {"clusters": clusters}
```
danielkza: Wouldn' it be more useful to return the results of the `create_cluster` calls?

stacker/hooks/iam.py
```diff
@@ -139,10 +141,15 @@ def ensure_server_cert_exists(region, namespace, mappings, parameters,
...
+        result = "uploaded"
         logger.info(
             "uploaded certificate: %s (%s)",
             cert_name,
             cert_arn,
         )
 
-    return True
+    return {
+        "result": result,
```
danielkza: What about using `status` instead of `result`? Just a nitpick.

stacker/hooks/keypair.py
```diff
@@ -26,7 +26,11 @@ def ensure_keypair_exists(region, namespace, mappings, parameters, **kwargs):
                     keypair["KeyName"],
                     keypair["KeyFingerprint"],
                     "exists")
-        return True
+        return {
+            "result": "exists",
```
danielkza: What about using status instead of result? Just a nitpick.

stacker/hooks/keypair.py
```diff
@@ -50,7 +54,12 @@ def ensure_keypair_exists(region, namespace, mappings, parameters, **kwargs):
                     keypair["KeyName"],
                     keypair["KeyFingerprint"],
                     "imported")
-        return True
+        return {
+            "result": "imported",
```
danielkza: What about using status instead of result? Just a nitpick.

stacker/hooks/keypair.py
```diff
@@ -71,10 +80,15 @@ def ensure_keypair_exists(region, namespace, mappings, parameters, **kwargs):
...
                     "created")
-        f = open(os.path.join(full_path, file_name), "w")
-        f.write(keypair["KeyMaterial"])
-        f.close()
-        return True
+        with open(os.path.join(full_path, file_name), "w") as f:
+            f.write(keypair["KeyMaterial"])
+
+        return {
+            "result": "created",
```
danielkza: What about using status instead of result? Just a nitpick.

phobologic:
@danielkza updated, thanks for the review!

PR #286, opened 2016-12-04
Use sys_path for hooks and lookups

### LINKED ISSUE #283: Apply `sys_path` when finding handlers/lookups
It would be helpful if the `sys_path` config setting applied to lookups and handlers, not just blueprints. I would expect this layout to work, but instead it raises an `ImportError`:

#### `stacks.yml`:

```yaml
---

sys_path: .

lookups:
  foo: mymod.foo
```

#### `stacks.env`:

```
namespace: mystacks
```

#### `mymod.py`:

```python
def foo(value, **kwargs):
    return 'foobar'
```

phobologic:
Fixes #283

This causes the `sys_path` config setting to be appended to `sys.path`
at the context level when it loads the config, rather than at the action
level.  This means it can be used everywhere, including lookups & hooks.

### DIFF
--- a/stacker/actions/base.py
+++ b/stacker/actions/base.py
@@ -1,6 +1,5 @@
 import copy
 import logging
-import sys
 
 import boto3
 import botocore.exceptions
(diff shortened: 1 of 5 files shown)


phobologic:
@ejholmes I figured out what it was - lately I've been naming my branches `mike/` - and I think that's what is causing the regexp to fail for docker.  Going to stop doing that :)  The tests themselves pass, so we should end up with a good image on merge.

PR #293, opened 2017-01-12
Avoid dictionary sharing pollution

### LINKED ISSUE #292: calling super() on `defined_variables` returns the base classes `VARIABLES` attribute
This doesn't bite you, unless you have two classes that subclass a single class that then subclasses Blueprint, and all call super() on `defined_variables`.  Example:

- class `BaseRDS` is a subclass of `Blueprint`
- class `MasterInstance` is a subclass of `BaseRDS`
- class `ReadReplica` is a subclass of `BaseRDS`

You try to build an environment with a `MasterInstance` and a `ReadReplica`.  First `MasterInstance` is built, since `ReadReplica` needs it to come up.  The following calls happen:

```
MasterInstance.defined_variables() -> BaseRDS.defined_variables() -> Blueprint.defined_variables()
```

With each class modifying the variables along the way.  The issue is that this is how the Blueprint executes `defined_variables`:

https://github.com/remind101/stacker/blob/master/stacker/blueprints/base.py#L301
```
        return getattr(self, "VARIABLES", {})
```

Since it returns the dictionary attached to the class, it is returning the dictionary defined on `BaseRDS`, and since it's a dictionary all the changes being made further down the line result in the `VARIABLES` dictionary being changed.

This works fine in most cases, but when it gets around to building the `ReadReplica` now the `VARIABLES` on `BaseRDS` are filled with variables from `MasterInstance`.

This should be an easy fix thankfully, but what a painful and tricky bug to find.  Python dictionaries are scary.

phobologic:
Fixes #292

### DIFF
--- a/stacker/blueprints/base.py
+++ b/stacker/blueprints/base.py
@@ -308,7 +308,7 @@ class Blueprint(object):
             dict: variables defined by the blueprint
 
         """
-        return getattr(self, "VARIABLES", {})
+        return copy.deepcopy(getattr(self, "VARIABLES", {}))
 
     def get_variables(self):
         """Return a dictionary of variables available to the template.


mhahn:
+1

ttaub:
+1

PR #305, opened 2017-01-24
Self parse user data

ttaub:
This PR adds the ability to parse variables referenced in a userdata file.

_UserData.yaml_

```
#!/bin/sh
echo Hey, {{name}}
```

_stack.yaml_

```
- name: stack
   variables: 
     name: tom
     UserData: ${file plain:file:/path/to/file}
```

_stack.py - somewhere where usedata needs to be passed in_

```
...
UserData = self.parse_user_data( variables['Userdata'] )
...
```
this will convert all variables enclosed in {{ ... }} to their respective values and base64 encode the entire string.

### DIFF
--- a/docs/blueprints.rst
+++ b/docs/blueprints.rst
@@ -103,7 +103,7 @@ types by directly passing in the value from the config to the specified
 Troposphere type.
 
 Example
-=======
+^^^^^^^
 
 Below is an annotated example::
 
@@ -171,7 +171,7 @@ specific Parameter types like ``List<AWS::EC2::Image::Id>``. See
 ``CFNType``.
 
 Example
-=======
+^^^^^^^
 
 Below is an annotated example::
 


stacker/blueprints/base.py
```diff
@@ -402,6 +405,36 @@ def render_template(self):
         version = hashlib.md5(rendered).hexdigest()[:8]
         return (version, rendered)
 
+    def parse_user_data(self, raw_userdata):
```
mhahn: i would be consistent about how you reference `user_data` vs `userdata`. i prefer `user_data`

stacker/blueprints/base.py
```diff
@@ -402,6 +405,36 @@ def render_template(self):
...
+        """Translate a userdata file to into the file contents.
+
+        It supports referencing template variables to create userdata
+        that's supplemented with information from the data, as commonly
```
mhahn: this is kind of a confusing comment. "from the data" is vague.

stacker/blueprints/base.py
```diff
@@ -402,6 +405,36 @@ def render_template(self):
...
+        required when creating EC2 userdata files. Automatically, encodes
+        the data file to base64 after it is processed.
+
+        """
+        pattern = re.compile(r'{{([::|\w]+)}}')
+        userdata = ""
+        s_index = 0
```
mhahn: i prefer to just be explicit and call this `start_index`

stacker/blueprints/base.py
```diff
@@ -402,6 +405,36 @@ def render_template(self):
...
+        variables = self.get_variables()
+
+        for match in pattern.finditer(raw_userdata):
+            userdata += raw_userdata[s_index:match.start()]
+
+            key = match.group(1)
+
+            if key in variables:
+                userdata += variables[key]
```
mhahn: one thing here is this would break if something was a `CFNString` since the value has to be `Ref`'d i can't remember exactly what that looks like off the top of my head, but i would add a test case to be sure.

one thing that would be cool is if this handled adding the `Ref` for you similar to the lookup.

stacker/blueprints/base.py
```diff
@@ -402,6 +405,36 @@ def render_template(self):
...
+            else:
+                raise MissingVariable(self.name, key)
+
+            s_index = match.end()
+
+        userdata += raw_userdata[s_index:]
+
+        return base64.b64encode(userdata)
```
mhahn: are there ever cases where we don't want to b64 this? not sure, but i remember the parameterized lookup had support for not b64 encoding.

stacker/blueprints/base.py
mhahn: you should also test this without just strings. what happens with array etc.

PR #306, opened 2017-01-25
Self parse user data

ttaub:
This PR adds the ability to parse variables referenced in a userdata file.

_UserData.yaml_

```
#!/bin/sh
echo Hey, {{name}}
```

_stack.yaml_

```
- name: stack
   variables: 
     name: tom
     UserData: ${file plain:file:/path/to/file}
```

_stack.py - somewhere where userdata needs to be passed in_

```
...
UserData = self.parse_user_data( variables['Userdata'] )
...
```
this will convert all variables enclosed in {{ ... }} to their respective values and base64 encode the entire string.


stacker/blueprints/base.py
```diff
@@ -402,6 +406,44 @@ def render_template(self):
...
+    def parse_user_data(self, raw_user_data):
+        """Translate a userdata file to into the file contents.
+
+        It supports referencing template variables to create userdata
+        that's supplemented with information from the data, as commonly
+        required when creating EC2 userdata files. Automatically, encodes
+        the data file to base64 after it is processed.
+
+         Args:
+            raw_userdata (str of userdata):
```
phobologic: Just `str` instead of `str of userdata`.  Also, make sure you put the return statement in the docstring as well - see: http://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html

stacker/blueprints/base.py
```diff
@@ -402,6 +406,44 @@ def render_template(self):
...
+                str of the given userdata to be parsed
+
+        """
+        pattern = re.compile(r'{{([::|\w]+)}}')
+        user_data = []
+        start_index = 0
+        variables = self.get_variables()
+        parameters = self.get_cfn_parameters()
+
+        for match in pattern.finditer(raw_user_data):
```
phobologic: Lets go with the standard python `string.Template` method of templating: https://docs.python.org/3/library/string.html#template-strings

Also, please double check that the parameters aren't availabe in `get_variables` - seems strange, but worth double checking (@mhahn might know)

stacker/blueprints/base.py
```diff
@@ -402,6 +406,44 @@ def render_template(self):
...
+                user_data.append(parameters[key])
+            elif key in variables:
+                user_data.append(variables[key])
+            else:
+                raise MissingVariable(self.name, key)
+
+            start_index = match.end()
+
+        user_data.append(raw_user_data[start_index:])
+        res = Join("", user_data)
```
phobologic: Since this is just a template, and you're not doing anything w/ Ref, etc, you probably won't need any of this join logic - it'll just be a template.

stacker/blueprints/base.py
ttaub: 1. I decided to stick to just using a regex because the standard delimiter in the Template strings is $ which is very commonly used in UserData for other purposes. We can subclass the Template class and define our own custom rules for the delimiters, but I would argue that what we have right now is cleaner.

2. Parameters are in get_variables, however, they are stored as CFNParameter objects. In my new commit I addressed this is a cleaner way.

stacker/blueprints/base.py
ttaub: Done.

stacker/blueprints/base.py
ttaub: Done.

stacker/blueprints/base.py
```diff
@@ -402,6 +405,46 @@ def render_template(self):
...
+
+        Args:
+            raw_user_data (str): the user data with the cloud-init info
+
+        Returns:
+            str: The parsed user data, with all the variables values and
+                 refs replaced with their resolved values.
+
+        """
+        pattern = re.compile(r'{{([::|\w]+)}}')
```
phobologic: One issue w/ building your own templating engine is that there's no way to escape this with the current implementation, right?  Python actually has a second string formatting/templating system that you can use perhaps? https://docs.python.org/3/library/string.html#format-string-syntax

stacker/blueprints/base.py
```diff
@@ -402,6 +405,46 @@ def render_template(self):
...
+        res = ""
+        start_index = 0
+        variables = self.get_variables()
+
+        for match in pattern.finditer(raw_user_data):
+            res += raw_user_data[start_index:match.start()]
+
+            key = match.group(1)
+
+            if key in variables:
```
phobologic: I wouldn't usually check for the existence of a key - instead you should use try/except:

```
try:
    v = variables[key]
except KeyError:
    raise MissingVariable(self.name, key)
```

Then work with v for the rest of the logic, outside of the block.

stacker/blueprints/base.py
```diff
@@ -402,6 +405,46 @@ def render_template(self):
...
+                else:
+                    res += variables[key]
+            else:
+                raise MissingVariable(self.name, key)
+
+            start_index = match.end()
+
+        res += raw_user_data[start_index:]
+
+        return base64.b64encode(res)
```
phobologic: It might be better to not encode this at all, letting the user choose whether to encode it either using base64 or the Base64 troposphere method.  what do you think?

ttaub: I think that's fair. They could extend the script if they wanted to.

ttaub:
The way userdata templating works is fixed, it uses the default string template provided by python. The main problem with my previous attempt was the way I was passing in the userdata was as a plain string using the file lookup handler which was passing its output to the default output handler and causing all sorts of problems. Now, the function only requires the path and it reads the file itself.

Example config file: test.yaml
```
...
UserDataPath: file://tmp.txt
#file:// is required for it to work 
...
```

Example template file: test.py

```
...
...
userdata=self.read_user_data(variables['UserDataPath'])
...
```

ttaub:
I can move this to stacker_blueprints, I think it may be a better fit.

stacker/blueprints/base.py
```diff
@@ -220,7 +222,48 @@ def resolve_variable(var_name, var_def, provided_variable, blueprint_name):
     return value
 
 
+def parse_user_data(variables, raw_user_data, blueprint_name):
+    """Translate a userdata file to into the file contents.
+
+        It supports referencing template variables to create userdata
+        that's supplemented with information from the data, as commonly
+        required when creating EC2 userdata files. Automatically, encodes
+        the data file to base64 after it is processed.
```
phobologic: I don't think this is true any longer, right?

stacker/blueprints/base.py
```diff
@@ -1,11 +1,13 @@
 import copy
 import hashlib
 import logging
+from string import Template as StringTemplate
```
phobologic: I would just use `import string` here, and then use Template as `string.Template` later

stacker/blueprints/base.py
```diff
@@ -220,7 +222,48 @@ def resolve_variable(var_name, var_def, provided_variable, blueprint_name):
...
+
+        Returns:
+            str: The parsed user data, with all the variables values and
+                 refs replaced with their resolved values.
+
+        Raises:
+            MissingVariable: Raised when a variable is in the user_data that
+                             is not given in the blueprint
+
+        """
```
phobologic: This appears to be over indented - I think the whole docstring block is actually.

stacker/blueprints/base.py
```diff
@@ -220,7 +222,48 @@ def resolve_variable(var_name, var_def, provided_variable, blueprint_name):
...
+def parse_user_data(variables, raw_user_data, blueprint_name):
+    """Translate a userdata file to into the file contents.
+
+        It supports referencing template variables to create userdata
+        that's supplemented with information from the data, as commonly
+        required when creating EC2 userdata files. Automatically, encodes
+        the data file to base64 after it is processed.
+
+        Args:
+            raw_user_data (str): the user data with the cloud-init info
```
phobologic: missing some Args here.

stacker/blueprints/base.py
mhahn: this indentation looks off

stacker/blueprints/base.py
```diff
@@ -220,7 +222,48 @@ def resolve_variable(var_name, var_def, provided_variable, blueprint_name):
     return value
 
 
+def parse_user_data(variables, raw_user_data, blueprint_name):
+    """Translate a userdata file to into the file contents.
```
mhahn: this sentence doesn't make sense

stacker/blueprints/base.py
```diff
@@ -220,7 +222,48 @@ def resolve_variable(var_name, var_def, provided_variable, blueprint_name):
...
+        else:
+            variable_values[key] = variables[key]
+
+    template = StringTemplate(raw_user_data)
+
+    res = ""
+
+    try:
+        res = template.substitute(variable_values)
+    except Exception as e:
```
mhahn: we should be catching the exact exception that gets raised

stacker/blueprints/base.py
```diff
@@ -219,23 +220,28 @@ def resolve_variable(var_name, var_def, provided_variable, blueprint_name):
 
 
 def parse_user_data(variables, raw_user_data, blueprint_name):
-    """Translate a userdata file to into the file contents.
+    """Parses the user data file
```
phobologic: Maybe `Parse the given user data and render it as a template`

stacker/blueprints/base.py
```diff
@@ -246,14 +252,16 @@ def parse_user_data(variables, raw_user_data, blueprint_name):
...
-    template = StringTemplate(raw_user_data)
+    template = string.Template(raw_user_data)
 
     res = ""
 
     try:
         res = template.substitute(variable_values)
-    except Exception as e:
-        raise MissingVariable(blueprint_name, e)
+    except ValueError:
```
phobologic: how does this get raised?  Not sure I understand from the exception, seems similar to the MissingVariable?

stacker/exceptions.py
```diff
@@ -44,6 +54,14 @@ def __init__(self, blueprint_name, variable_name, *args, **kwargs):
         super(MissingVariable, self).__init__(message, *args, **kwargs)
 
 
+class InvalidUserdataParameter(Exception):
```
phobologic: Where is this used?

stacker/blueprints/base.py
ttaub: This gets raised when the placeholder name for the variable in the template is actually invalid. E.g ${100} raises a ValueError because the placeholder name (100) can not only be a number.

ttaub: This is also explained in the Raises section of the comments.

stacker/blueprints/base.py
```diff
@@ -216,7 +219,55 @@ def resolve_variable(var_name, var_def, provided_variable, blueprint_name):
...
+
+    Raises:
+        MissingVariable: Raised when a variable is in the user_data that
+                         is not given in the blueprint
+
+        InvalidUserdataPlaceholder: Raised when a placeholder name in user_data
+                                    is not valid. E.g ${100} would raise this.
+
+
+        """
```
phobologic: This indentation is still off, should be flush with the start of the doc string.

stacker/blueprints/base.py
```diff
@@ -216,7 +219,55 @@ def resolve_variable(var_name, var_def, provided_variable, blueprint_name):
...
+
+    Returns:
+        str: The parsed user data, with all the variables values and
+             refs replaced with their resolved values.
+
+    Raises:
+        MissingVariable: Raised when a variable is in the user_data that
+                         is not given in the blueprint
+
+        InvalidUserdataPlaceholder: Raised when a placeholder name in user_data
```
phobologic: `user_data` -> `raw_user_data` - this actually happens in multiple places in the doc string.

stacker/exceptions.py
```diff
@@ -17,6 +17,16 @@ def __init__(self, lookup, *args, **kwargs):
         super(UnknownLookupType, self).__init__(message, *args, **kwargs)
 
 
+class InvalidUserdataPlaceholder(Exception):
+
+    def __init__(self, blueprint_name, *args, **kwargs):
+        message = "Could not parse userdata in blueprint \"%s\". " % (
```
phobologic: I'd include the text from the ValueError that raises this - would make it easier to track down the issues:

```
>>> t.substitute({"100": "blue", "a": "apple"})
  File "<stdin>", line 1, in <module>
  File "/usr/local/Cellar/python/2.7.12_1/Frameworks/Python.framework/Versions/2.7/lib/python2.7/string.py", line 176, in substitute
    return self.pattern.sub(convert, self.template)
  File "/usr/local/Cellar/python/2.7.12_1/Frameworks/Python.framework/Versions/2.7/lib/python2.7/string.py", line 173, in convert
    self._invalid(mo)
  File "/usr/local/Cellar/python/2.7.12_1/Frameworks/Python.framework/Versions/2.7/lib/python2.7/string.py", line 146, in _invalid
    (lineno, colno))
ValueError: Invalid placeholder in string: line 2, col 1
```

The line # & col # seem useful.  You can get that from `.args` of the original error.

PR #307, opened 2017-01-26
Add the credential caching

ttaub:
This PR allows stacker to cache the temp credentials for the assumed role and use them until they expire without being required to use a OTP for every operation. 

How to use functionality:

1) Make sure you have an aws config file: [http://docs.aws.amazon.com/cli/latest/topic/config-vars.html](http://docs.aws.amazon.com/cli/latest/topic/config-vars.html)

2) export AWS_PROFILE=profile_you_want_to_use

3) Call any stacker command **without** using assume-role

The first time you use this, it will ask for an mfa code, it will cache the response of the assume-role call and only ask for the code when the credentials expire.

Caveats:

**Do not use this with assume-role**

Make sure **AWS_PROFILE** is the only AWS related environment variable that is set. Otherwise, it will not use the role you want


stacker/actions/base.py
```diff
@@ -61,8 +64,11 @@ def __init__(self, context, provider=None):
     def s3_conn(self):
         """The boto s3 connection object used for communication with S3."""
         if not hasattr(self, "_s3_conn"):
-            session = boto3.Session(region_name=self.provider.region)
-            self._s3_conn = session.client('s3')
+            # session = boto3.Session(region_name=self.provider.region)
```
phobologic: no need for these to stay in the code, I'd go ahead and remove them

stacker/hooks/aws_lambda.py
```diff
@@ -415,8 +416,10 @@ def create_template(self):
     else:
         logger.info('lambda: using custom bucket: %s', bucket)
 
-    session = boto3.Session(region_name=region)
-    s3_conn = session.client('s3')
+    # session = boto3.Session(region_name=region)
```
phobologic: Same here.

stacker/providers/session_cache.py
```diff
@@ -0,0 +1,64 @@
+import json
+import os
+import logging
+import botocore
+
+
+def get_session(region):
```
phobologic: Could use a docstring.

phobologic: Also probably worth throwing some debug logging statements in here.

stacker/providers/session_cache.py
```diff
@@ -0,0 +1,64 @@
...
+    session = botocore.session.get_session()
+    session.set_config_variable('region',  region)
+    c = session.get_component('credential_provider')
+    provider = c.get_provider('assume-role')
```
phobologic: How does this work?  What if someone isn't using an assumed role, will this cause issues?

stacker/providers/session_cache.py
```diff
@@ -0,0 +1,64 @@
...
+    provider.cache = CredentialCache()
+    return session
+
+
+class CredentialCache(object):
```
phobologic: Could you write some tests around this class?

stacker/providers/session_cache.py
```diff
@@ -0,0 +1,64 @@
...
+    """
+
+    CACHE_DIR = os.path.expanduser(
+        os.path.join('~', '.stacker', 'cache'))
+
+    def __init__(self, working_dir=CACHE_DIR):
+        self._working_dir = working_dir
+
+    def __contains__(self, cache_key):
+        print("Called Contains")
```
phobologic: I'm guessing these are just debug statements, but they should be pulled out.  Might be worth replacing them with logging statements at the debug level though.

stacker/providers/session_cache.py
```diff
@@ -0,0 +1,64 @@
+import json
```
phobologic: The providers directory is actually for service providers (like AWS - in theory we could have more in the future).  Might be worth giving this it's own module like `stacker/credential_cache.py`

stacker/providers/session_cache.py
```diff
@@ -0,0 +1,64 @@
...
+        print("Called Set Item")
+        full_key = self._convert_cache_key(cache_key)
+        try:
+            file_content = json.dumps(value, default=str)
+        except (TypeError, ValueError):
+            raise ValueError("Value cannot be cached, must be "
+                             "JSON serializable: %s" % value)
+        if not os.path.isdir(self._working_dir):
+            os.makedirs(self._working_dir)
+        with os.fdopen(os.open(full_key,
```
phobologic: why not just use `with open` here?

ttaub:
This is still a WIP, I'll be adding tests and making changes to it. I just committed it before I left work. Thanks for all the tips!

stacker/providers/session_cache.py
ttaub: There will be no issues, assume-role is just where botocore stores all the role info.

stacker/session_cache.py
```diff
@@ -23,18 +32,16 @@ class CredentialCache(object):
     """
 
     CACHE_DIR = os.path.expanduser(
-        os.path.join('~', '.stacker', 'cache'))
+        os.path.join('~', '.aws', 'cli', 'cache'))
```
phobologic: Is it worth putting some debug logging in here so that people can debug in case things start breaking?

stacker/session_cache.py
```diff
@@ -1,16 +1,25 @@
...
 
 
 def get_session(region):
+    """Creates a boto3 session with a cache
+
+    Args:
+        region (str): The region for the session
+
+    Returns:
+        Object: A  boto3 session with credential
```
phobologic: This actually returns a `boto3.session.Session` object, right?  See https://github.com/remind101/stacker/blob/52b4508ab2f030f5127774afdd50d2f27206ef7b/stacker/stack.py#L63 for an example of how you can document those.

PR #311, opened 2017-01-29
Convert booleans to strings for CFNTypes

### LINKED ISSUE #290: Convert booleans in YAML to strings for CFNParameter type vars
For example: 

`storage_encrypted: "true"`

should be able to be written as a boolean: 

`storage_encrypted: true` 

in the environment and config files.

phobologic:
This is old functionality that was lost with the conversion to Variables
& CFNTypes.

Fixes #290

### DIFF
--- a/stacker/blueprints/base.py
+++ b/stacker/blueprints/base.py
@@ -72,7 +72,6 @@ class CFNParameter(object):
                 "CFNParameter (%s) value must be one of %s got: %s" % (
                     name, "str, int, bool, or list", value))
 
-        print value
         self.name = name
         self.value = value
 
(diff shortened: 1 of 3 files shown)


stacker/blueprints/base.py
```diff
@@ -43,25 +43,33 @@ def __init__(self, name, value):
...
-                acceptable = True
+                    break
 
         if not acceptable:
             raise ValueError(
                 "CFNParameter (%s) value must be one of %s got: %s" % (
-                    name, "str, int, or list", value))
+                    name, "str, int, bool, or list", value))
 
+        print value
```
ttaub: Is this leftover debug code?

phobologic: Certainly is - good catch, I'll remove it

PR #314, opened 2017-01-31
Added ssmstore as a lookup type.

syphon7:


stacker/lookups/handlers/ssmstore.py
```diff
@@ -0,0 +1,50 @@
...
+
+    client = boto3.client("ssm", region_name=region)
+    response = client.get_parameters(
+        Names=[
+            value,
+        ],
+        WithDecryption=True
+    )
+
+    return response['Parameters'][0]['Value']
```
phobologic: This will throw an IndexError if the parameter doesn't exist, right?  I wonder if there might be a better error we can return that'll give folks a better idea of where this is breaking?  Maybe a new exception type?

stacker/tests/lookups/handlers/test_ssmstore.py
```diff
@@ -0,0 +1,51 @@
...
+
+        self.ssmkey = "ssmkey"
+        self.ssmvalue = 'ssmvalue'
+
+    @patch('stacker.lookups.handlers.ssmstore.boto3.client',
+           return_value=client)
+    def test_ssmstore_handler(self, mock_client):
+        with self.stubber:
+            value = handler(self.ssmkey)
+            print "Value: %s" % value
```
phobologic: leftover debug statement?  Can probably remove.

stacker/tests/lookups/handlers/test_ssmstore.py
```diff
@@ -0,0 +1,51 @@
...
+            self.assertEqual(value, self.ssmvalue)
+
+    @patch('stacker.lookups.handlers.ssmstore.boto3.client',
+           return_value=client)
+    def test_ssmstore_handler_with_region(self, mock_client):
+        region = "us-east-1"
+        temp_value = "%s@%s" % (region, self.ssmkey)
+        print temp_value
```
phobologic: Another debug statement

phobologic:
Looks great!  Thanks @syphon7 !

PR #315, opened 2017-02-03
Update ssmstore handler to use cred cache

phobologic:


phobologic:
@syphon7 @ttaub

phobologic:
This is to take into account the new credential cache for 1.0 for the new ssmstore lookup :)

stacker/tests/lookups/handlers/test_ssmstore.py
```diff
@@ -1,12 +1,21 @@
 import unittest
-import boto3
 import mock
 from botocore.stub import Stubber
 from stacker.lookups.handlers.ssmstore import handler
+import boto3
+
+
+class sessionStub:
```
phobologic: `SessionStub`

stacker/tests/lookups/handlers/test_ssmstore.py
```diff
@@ -1,12 +1,21 @@
...
+
+    def __init__(self, clientStub):
```
phobologic: `clientStub` -> `client_stub`

Issue #298, opened 2017-01-18
Remove explicit/default lookup pointing at `output`


phobologic:
Right now if you don't specify the lookup you're using, it'll default to `output`, as detailed here: http://stacker.readthedocs.io/en/latest/lookups.html

After a chat in the Slack, folks seemed to agree that this shortcut wasn't necessary, and could lead to confusion and/or other issues.  It also complicates the code a bit.

To do that, the plan is to remove this feature in 1.0.

phobologic:
Fixed in #310

PR #317, opened 2017-02-04
add default lookup

troyready:
**Updated**: See next comment instead.

I'd like the ability to reference environment variables that might not
be specified.

~~However, it doesn't appear to be working yet; specifying something like:~~
```
Variables:
  InstanceType: ${default ${nonpresent_env_var}::t2.micro}
```
~~in the config yaml results in a stacktrace:~~
```
  File "/Users/troyready/Library/Python/2.7/bin/stacker", line 9, in <module>
    args.run(args)
  File "/Users/troyready/Library/Python/2.7/lib/python/site-packages/stacker/commands/stacker/build.py", line 52, in run
    dump=options.dump)
  File "/Users/troyready/Library/Python/2.7/lib/python/site-packages/stacker/actions/base.py", line 122, in execute
    self.run(*args, **kwargs)
  File "/Users/troyready/Library/Python/2.7/lib/python/site-packages/stacker/actions/build.py", line 279, in run
    plan = self._generate_plan(tail=tail)
  File "/Users/troyready/Library/Python/2.7/lib/python/site-packages/stacker/actions/build.py", line 243, in _generate_plan
    dependencies = self._get_dependencies()
  File "/Users/troyready/Library/Python/2.7/lib/python/site-packages/stacker/actions/build.py", line 255, in _get_dependencies
    dependencies[stack.fqn] = stack.requires
  File "/Users/troyready/Library/Python/2.7/lib/python/site-packages/stacker/stack.py", line 87, in requires
    d = deconstruct(lookup.input)
  File "/Users/troyready/Library/Python/2.7/lib/python/site-packages/stacker/lookups/handlers/output.py", line 44, in deconstruct
    stack_name, output_name = value.split("::")
ValueError: need more than 1 value to unpack
```

~~I'm think this might be related to #298 , so this might 'just work' after it's resolved?~~

### DIFF
--- a/stacker/lookups/handlers/default.py
+++ b/stacker/lookups/handlers/default.py
@@ -1,0 +1,32 @@
+TYPE_NAME = "default"
+
+
+def handler(value, **kwargs):
+    """Use a value from the environment or fall back to a default if the
+       environment doesn't contain the variable.
+
+    Format of value:
+
+        <env_var>::<default value>
+
+    For example:
+
+        Groups: ${default app_security_groups::sg-12345,sg-67890}
+
+    If `app_security_groups` is defined in the environment, its defined value
+    will be returned. Otherwise, `sg-12345,sg-67890` will be the returned
+    value.
+
+    This allows defaults to be set at the config file level.
+    """
+
+    try:
+        env_var_name, default_val = value.split("::", 1)
+    except ValueError:
+        raise ValueError("Invalid value for default: %s. Must be in "
+                         "<env_var>::<default value> format." % value)
+
+    if env_var_name in kwargs['context'].environment:
+        return kwargs['context'].environment[env_var_name]
+    else:
+        return default_val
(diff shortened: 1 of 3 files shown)


troyready:
I've cleaned this up to actual function properly with working tests. I'm not married to the lookup name, but I like the functionality.

phobologic:
Cool - thanks @troyready I'm wondering if we should make this a default part of all lookups, rather than it's own lookup.  Not entirely sure how we'd work the syntax for that though.  Any thoughts?

troyready:
@phobologic yeah, that could be nice. Really, my ideal use case would be something like a CloudFormation `If ... else Ref('AWS::NoValue')` so that a missing environment value would have the blueprint variable's default value used instead.

I don't want to muck up the design too much for our edge-case use though; I like the simplicity of how this fits in with the existing lookup system.

phobologic:
Ok, yeah, that makes sense - this is great.  Thanks @troyready

PR #318, opened 2017-02-07
Remove default aws region

troyready:
This will still allow the region to be specified via '-r', but adds the
additional option to set it via `AWS_DEFAULT_ENVIRONMENT`.

A failure to provide either will generate a fairly clear stacktrace
like:
```
...
  File "/Users/troyready/Library/Python/2.7/lib/python/site-packages/botocore/regions.py", line 135, in _endpoint_for_partition
    raise NoRegionError()
botocore.exceptions.NoRegionError: You must specify a region.
```

This does introduce an additional unit testing requirement to set the environment variable; not sure if there's a better way to handle that for the kms lookup?

It's a fairly major change, but this follows alongside the logic of #298 and will help keep mistakes from happening (a NoRegionError is more clear than an xref failing with a output lookup stacktrace, or a bunch of stuff deployed to the wrong region).

### DIFF
--- a/stacker/commands/stacker/base.py
+++ b/stacker/commands/stacker/base.py
@@ -126,9 +126,8 @@ class BaseCommand(object):
                                  "the command line. Overrides your "
                                  "environment file settings. Can be specified "
                                  "more than once.")
-        parser.add_argument("-r", "--region", default="us-east-1",
-                            help="The AWS region to launch in. Default: "
-                                 "%(default)s")
+        parser.add_argument("-r", "--region",
+                            help="The AWS region to launch in.")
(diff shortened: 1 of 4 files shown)


ttaub:
Looks good to me, I like that it doesn't just automatically assume the default region. @troyready would you be able to write some tests for it?

troyready:
@ttaub sure thing, PR updated.

The only change to the connection code is logic around `None` being returned from the config, so I added a test for that config parsing.

phobologic:
Honestly, I think we should just do away with --region altogether, and use the standard (ie: either profile, the environment variable, or stack specific - when we have stack specific).  What do you think @troyready ?

PR #321, opened 2017-02-15
Make the new release-1.0 more clear.

ttaub:
The new error:

```
stacker only allows explicit lookup types now, and no longer uses `output` as the default if not specified. Please update your output lookups.
```

was preventing my old error which explicitly mentioned what variable it was failing on from being raised. So I just, added the new info the the error.

### DIFF
--- a/stacker/lookups/__init__.py
+++ b/stacker/lookups/__init__.py
@@ -38,10 +38,11 @@ def extract_lookups_from_string(value):
         raw = match.groups()[0]
         lookup_type = groupdict["type"]
         if not lookup_type:
-            raise ValueError("stacker only allows explicit lookup types "
+            raise ValueError("on value `%s`\n"
+                             "stacker only allows explicit lookup types "
                              "now, and no longer uses `output` as the "
                              "default if not specified. Please update your "
-                             "output lookups.")
+                             "output lookups." % groupdict["input"])
         lookup_input = groupdict["input"]
         lookups.add(Lookup(lookup_type, lookup_input, raw))
     return lookups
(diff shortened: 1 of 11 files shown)


stacker/lookups/__init__.py
```diff
@@ -38,10 +38,11 @@ def extract_lookups_from_string(value):
         raw = match.groups()[0]
         lookup_type = groupdict["type"]
         if not lookup_type:
-            raise ValueError("stacker only allows explicit lookup types "
+            raise ValueError("on value `%s`\n"
```
phobologic: I'd instead keep this error the way it is, but catch it at:

https://github.com/remind101/stacker/blob/release-1.0/stacker/variables.py#L149

Then we can raise an exception with the original message, but give both the variable name and value if necessary.

stacker/lookups/handlers/output.py
```diff
@@ -37,11 +37,5 @@ def handler(value, provider=None, context=None, fqn=False, **kwargs):
 
 
 def deconstruct(value):
-
-    # Probably an undefined environment variable
```
phobologic: I think this is still necessary, though now that I've looked at it an if statement was probably the wrong way to catch it.  Here's how it should probably be written (note, I think the `value.split` is wrong in and of itself as well:

```
try:
    stack_name, output_name = value.split("::", 1)
except ValueError:
    raise ValueError("output handler requires syntax of <stack>::<output>.  Got: %s" % value)
```

Not married to the message in the ValueError though, if you can think of something better

PR #328, opened 2017-02-17
added rxref lookup

discobean:
The new rxref lookup will search for stack names relative to the current `namespace` identified in the environment file.

This encourages reuse of stacks in multiple environments within the same AWS account.

Thanks for building some great software!

### DIFF
--- a/stacker/lookups/handlers/output.py
+++ b/stacker/lookups/handlers/output.py
@@ -4,8 +4,7 @@ TYPE_NAME = "output"
 
 Output = namedtuple("Output", ("stack_name", "output_name"))
 
-
-def handler(value, provider=None, context=None, fqn=False, **kwargs):
+def handler(value, provider=None, context=None, fqn=False, rfqn=False, **kwargs):
     """Fetch an output from the designated stack.
 
     Args:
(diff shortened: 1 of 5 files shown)


phobologic:
Hey @discobean - sorry for the delay at looking at this, it got lost in the mix up of work & vacation!  I'll try to get this looked at more today!  Thanks for the contribution!

stacker/lookups/handlers/output.py
```diff
@@ -17,11 +18,22 @@ def handler(value, provider=None, context=None, fqn=False, **kwargs):
...
+        rfqn (bool): boolean for whether or not the
+            :class:`stacker.context.Context` should resolve the `fqn` of the
+            stack prefixed by the namespace variable
 
     Returns:
         str: output from the specified stack
 
     """
+
+    if rfqn and context.environment.get('namespace'):
```
phobologic: Shouldn't need the `context.environment.get('namespace')` due to this check: https://github.com/remind101/stacker/blob/master/stacker/context.py#L49

stacker/lookups/handlers/output.py
```diff
@@ -17,11 +18,22 @@ def handler(value, provider=None, context=None, fqn=False, **kwargs):
...
+            value = "%s%s%s" % (
+                    context.environment['namespace'],
+                    context.environment.get('namespace_delimiter')
```
phobologic: There should never be a time that the delimiter isn't set as far as I know, but if you can think of a time when that would happen you would better express this as:

`context.environment.get('namespace_delimiter', '-')`

discobean:
I've updated adding those tests, let me know how it goes.

stacker/lookups/handlers/output.py
```diff
@@ -27,11 +27,11 @@ def handler(value, provider=None, context=None, fqn=False, rfqn=False,
 
     """
 
-    if rfqn and context.environment.get('namespace'):
+    if rfqn and context.environment['namespace']:
```
phobologic: Sorry, I realize I might have been a little confusing in my original comment.  You should be able to change this line to:

```
if rfqn:

```

Since we always ensure that we have a namespace, it's required in the context.

discobean:
Just pushed that updated now.  Thanks!

gservat:
Really keen on this feature. How long to go until its merged??

discobean:
🎉  thanks @phobologic

bwhaley:
This is just what I was looking for! Thanks for this patch.

PR #331, opened 2017-02-24
Convert stacker to use sqs

ttaub:
To Do:
- [x] Fix Broken Tests
- [x] Add try catch to command to cleanup sqs queues
- [x] Create benchmarks to measure performance improvements
- [ ] Exhaustive testing for edge cases
- [x] Better error messages 
- [x] Add error page for stacker destroy error
- [x] Gracefully fail if sns topic does not exist for destroy command


stacker/providers/aws/default.py
```diff
@@ -106,18 +111,43 @@ def cloudformation(self):
...
+        pid = os.getpid()
+        if pid != self._pid or not self._listener:
+            session = get_session(self.region)
+            self._listener = CloudListener(
+                LISTENER_NAME, session=session)
+            self._listener.start()
+
+        return self._listener
+
+    def pre_run(self):
```
phobologic: Lets call this `flush_events` or something similar to better represent what it's doing.

stacker/providers/aws/default.py
```diff
@@ -106,18 +111,43 @@ def cloudformation(self):
...
+        logger.debug("Clearing old sqs queue")
+        messages = self.listener.get_messages()
+        for message in messages:
```
phobologic: Use `DeleteMessageBatch` to speed this up/limit the # of calls ot the API.

stacker/providers/aws/default.py
```diff
@@ -106,18 +111,43 @@ def cloudformation(self):
...
+            message.delete()
+        logger.info("Done clearing queue")
+
+    def parse_stack_events(self):
+        messages = self.listener.get_messages()
+        for message in messages:
+            message_dict = message.__dict__
+            message_dict['fqn'] = message_dict['StackName']
+            self._stacks[message.StackName] = message_dict
+            message.delete()
```
phobologic: See above about `DeleteMessageBatch`

stacker/providers/aws/default.py
```diff
@@ -106,18 +111,43 @@ def cloudformation(self):
...
+        return self._listener
+
+    def pre_run(self):
+        logger.debug("Clearing old sqs queue")
+        messages = self.listener.get_messages()
+        for message in messages:
+            message.delete()
+        logger.info("Done clearing queue")
+
+    def parse_stack_events(self):
```
phobologic: Docstring would be good here.

stacker/providers/aws/default.py
```diff
@@ -106,18 +111,43 @@ def cloudformation(self):
...
     def get_stack(self, stack_name, **kwargs):
-        try:
-            return retry_on_throttling(
-                self.cloudformation.describe_stacks,
-                kwargs=dict(StackName=stack_name))['Stacks'][0]
-        except botocore.exceptions.ClientError as e:
-            if "does not exist" not in e.message:
-                raise
+        self.parse_stack_events()
+        if stack_name in self._stacks:
```
phobologic: Rather than testing if the name is in the stack, just try to get it - and deal with the exception as it raises:

```
try:
    return self._stacks[stack_name]
except KeyError:
    raise exceptions.StackDoesNotExist(stack_name)
```

stacker/providers/aws/default.py
```diff
@@ -125,22 +125,30 @@ def listener(self):
         return self._listener
 
     def pre_run(self):
-        logger.debug("Clearing old sqs queue")
-        messages = self.listener.get_messages()
-        for message in messages:
-            message.delete()
-        logger.info("Done clearing queue")
+        logger.info("Purging old queue")
```
phobologic: This makes more sense as debug I think.

benchmark/tmp.env
```diff
@@ -0,0 +1,26 @@
+namespace: stacker-benchmark
+
+# VPC settings
```
phobologic: Is this used anywhere?  Seems like the current method just creates buckets.

benchmark/tmp.yaml
```diff
@@ -0,0 +1,12 @@
+stacks:
+  - name: bucket
```
phobologic: maybe worth having more buckets, and making some buckets rely on other buckets (you can do this explicity with the requires keyword on stacks)

stacker/actions/build.py
```diff
@@ -270,6 +269,9 @@ def pre_run(self, outline=False, dump=False, *args, **kwargs):
                 provider=self.provider,
                 context=self.context)
 
+        # Clear old sns events
+        self.provider.pre_run()
```
phobologic: I wouldn't use `pre_run` for this, instead I'd name it what it is/what it actually does.

stacker/providers/aws/default.py
```diff
@@ -106,18 +111,54 @@ def cloudformation(self):
...
+
+    def flush_events(self):
+        logger.info("Purging old queue")
+        while 1:
+            messages = self.listener.get_messages()
+            if len(messages) == 0:
+                break
+
+            for message in messages:
+                message.delete()
```
phobologic: Should look at batch delete for this, rather than deleting each message one call at a time.

stacker/providers/aws/default.py
```diff
@@ -106,18 +111,54 @@ def cloudformation(self):
...
+        if pid != self._pid or not self._listener:
+            session = get_session(self.region)
+            self._listener = CloudListener(
+                LISTENER_NAME, session=session)
+            self._listener.start()
+
+        return self._listener
+
+    def flush_events(self):
+        logger.info("Purging old queue")
```
phobologic: This seems more like debug info, also could be a bit more descriptive.  Also what happened with using a new SQS queue per run?

stacker/providers/aws/default.py
```diff
@@ -106,18 +111,54 @@ def cloudformation(self):
...
+        while 1:
+            messages = self.listener.get_messages()
+            if len(messages) == 0:
+                break
+
+            for message in messages:
+                message.delete()
+        logger.info("Done purging queue")
```
phobologic: debug

stacker/providers/aws/default.py
```diff
@@ -106,18 +111,54 @@ def cloudformation(self):
...
+
+    def pre_run(self):
```
phobologic: This feels like an early optimization, should probably just have things call self.flush_events until we need more steps.

stacker/providers/aws/default.py
```diff
@@ -106,18 +111,54 @@ def cloudformation(self):
...
+            message_dict = message.__dict__
+            message_dict['fqn'] = message_dict['StackName']
+            self._stacks[message.StackName] = message_dict
+
+            if tail:
+                Provider._tail_print(message)
+
+        self.listener.delete_messages(messages)
+
+    def get_stack(self, stack_name, tail=False, **kwargs):
```
phobologic: This is kind of confusing - I know you were trying to avoid changing the actual methods that were called as much as possible, but I think this may just make it more confusing.  Also the fact we need to keep the dictionary of stacks seems like a complication that we might not necessarily need if this was refactored correctly.  Lets talk about this more.

stacker/providers/aws/default.py
```diff
@@ -129,13 +170,21 @@ def is_stack_destroyed(self, stack, **kwargs):
...
-        def log_func(e):
-            event_args = [e['ResourceStatus'], e['ResourceType'],
-                          e.get('ResourceStatusReason', None)]
-            # filter out any values that are empty
-            event_args = [arg for arg in event_args if arg]
-            template = " ".join(["[%s]"] + ["%s" for _ in event_args])
-            logger.info(template, *([stack.fqn] + event_args))
+        def log_func(message):
+            # Check that the message is for the correct stack
+            if message.StackName == stack.fqn:
```
phobologic: Is this actually necessary anymore?  I think you'd only check the stack name if you were tailing the stack the old way?  Maybe this is partially due to the confusion introduced by keeping the old methodology.

stacker/providers/aws/default.py
```diff
@@ -129,13 +170,21 @@ def is_stack_destroyed(self, stack, **kwargs):
...
 
     def tail_stack(self, stack, retries=0, **kwargs):
-        def log_func(e):
-            event_args = [e['ResourceStatus'], e['ResourceType'],
-                          e.get('ResourceStatusReason', None)]
-            # filter out any values that are empty
-            event_args = [arg for arg in event_args if arg]
-            template = " ".join(["[%s]"] + ["%s" for _ in event_args])
-            logger.info(template, *([stack.fqn] + event_args))
+        def log_func(message):
```
phobologic: Not sure where this is actually called from now.

stacker/exceptions.py
```diff
@@ -102,6 +102,34 @@ def __init__(self, stack_name, output, *args, **kwargs):
...
 
 
+class DestoryWithoutNotificationQueue(Exception):
+
+    def __init__(self, stack_name, *args, **kwargs):
+        self.stack_name = stack_name
+        message = "You are attempting to destroy `%s` " % stack_name
+        message += "which was created using an older version of "
+        message += "Stacker. Please first updated all your stacks using "
+        message += "`Stacker build` so that they can be adjusted "
```
phobologic: `Stacker` -> `stacker`

stacker/exceptions.py
```diff
@@ -102,6 +102,34 @@ def __init__(self, stack_name, output, *args, **kwargs):
...
+        message += "for the new version. More information on "
+        message += "this issue here: "
+        message += "https://github.com/remind101/stacker/blob/"
```
phobologic: This doesn't seem to exist.

stacker/actions/build.py
```diff
@@ -176,7 +138,7 @@ def _build_stack_tags(self, stack):
         return [
             {'Key': t[0], 'Value': t[1]} for t in self.context.tags.items()]
 
-    def _launch_stack(self, stack, **kwargs):
+    def _launch_stack(self, stack, tail=False, **kwargs):
```
phobologic: Why do we pass tail down this far?  Doesn't appear to be used.

stacker/plan.py
```diff
@@ -40,11 +40,12 @@ class Step(object):
 
     """
 
-    def __init__(self, stack, run_func, requires=None):
+    def __init__(self, stack, run_func, tail, requires=None):
```
phobologic: Does this actually need tail anymore?

stacker/plan.py
```diff
@@ -150,8 +152,16 @@ def add(self, stack, run_func, requires=None):
             stack=stack,
             run_func=run_func,
             requires=requires,
+            tail=self.tail
         )
 
+    def poll(self):
+        stack_dict = self._poll_func(self.tail)
+        for step_name, step in self.list_pending():
+            if step_name in stack_dict:
```
phobologic: I'd replace this with:

```
try:
    status = stack_dict[step_name]
except KeyError:
    continue
step.set_status(status)
```

Usually it's better to just try and use something that should be there, then fail gracefully if it doesn't exist, rather than testing to see if it's there and then doing the work.

stacker/plan.py
```diff
@@ -210,65 +220,40 @@ def _single_run(self):
...
+            if not step.submitted:
+                try:
+                    status = step.run()
+                except CancelExecution:
+                    status = SkippedStatus(reason="canceled execution")
 
-            # Terminate any watchers when step completes
-            if step.done and step_name in self._watchers:
-                self._terminate_watcher(self._watchers[step_name])
+                if not isinstance(status, Status):
```
phobologic: I wonder if this should just be moved into `Step.set_status`, what do you think?

stacker/actions/destroy.py
```diff
@@ -48,42 +50,36 @@ def _generate_plan(self, tail=False):
         plan_kwargs = {}
         if tail:
```
phobologic: Is this still necessary?

stacker/plan.py
```diff
@@ -210,65 +220,40 @@ def _single_run(self):
...
+
+        while not self.completed:
+            if (not attempts % self.check_point_interval or
+                    self.md5 != last_md5):
+                last_md5 = self.md5
+                self._check_point()
+
+            attempts += 1
+            if not self._single_run():
+                self._wait_func(self.sleep_time)
```
phobologic: Do we need the wait_func here anymore?  Since `_single_run` long polls, I think this is just going to slow things down unnecessarily.

stacker/providers/aws/default.py
```diff
@@ -1,19 +1,24 @@
...
+from stacker.status import (
+    SubmittedStatus,
+    CompleteStatus,
+)
 
 logger = logging.getLogger(__name__)
 
 MAX_TAIL_RETRIES = 5
 
+LISTENER_NAME = "StackerSNSListener"
```
phobologic: This ends up being the queue name, right?  I don't think this should be hard-coded - I think we need to do a review of cloudsns at this point, it was sort of developed with very few PRs.

stacker/providers/aws/default.py
```diff
@@ -106,105 +112,65 @@ def cloudformation(self):
...
+            self._listener = CloudListener(
+                LISTENER_NAME,
+                session=session,
+                existing_topic_arn=existing_topic_arn
+            )
+            self._listener.start()
+
+        return self._listener
+
+    def set_listener_topic_arn(self, existing_topic_arn):
```
phobologic: So some weirdness here - if `self.listener` has already been called, then calling this will never work, since you test if `self._listener` exists before allowing any changes.

stacker/providers/aws/default.py
```diff
@@ -106,105 +112,65 @@ def cloudformation(self):
...
-    def tail_stack(self, stack, retries=0, **kwargs):
-        def log_func(e):
-            event_args = [e['ResourceStatus'], e['ResourceType'],
-                          e.get('ResourceStatusReason', None)]
-            # filter out any values that are empty
-            event_args = [arg for arg in event_args if arg]
-            template = " ".join(["[%s]"] + ["%s" for _ in event_args])
-            logger.info(template, *([stack.fqn] + event_args))
+    def cleanup(self):
+        self.listener.close()
```
phobologic: probably worht calling this `self.listener.teardown` or `cleanup`

stacker/plan.py
ttaub: Yes it passes it down to the provider polling function.

stacker/plan.py
ttaub: Not going to change this because 99% of the time, not all the stacks will exist in the event polls so this behaviour better suits the expected stack_dict

stacker/plan.py
ttaub: Agreed

stacker/actions/build.py
phobologic: This is still in here - does get_stack actually use tail/need tail?

stacker/plan.py
phobologic: But doesn't the polling happen in the Plan, not the step?

docs/destory_error.rst
```diff
@@ -0,0 +1,5 @@
+==================
+Destroy Error Note
+==================
+
+This is the sample error page
```
phobologic: This needs to be filled in with actual info, also you'll need to put it in index.rst I believe.

stacker/actions/build.py
```diff
@@ -287,6 +230,10 @@ def run(self, outline=False, tail=False, dump=False, *args, **kwargs):
             if dump:
                 plan.dump(directory=dump, context=self.context)
 
+    def cleanup(self):
```
phobologic: Needs a docstring.

stacker/actions/build.py
```diff
@@ -296,3 +243,50 @@ def post_run(self, outline=False, dump=False, *args, **kwargs):
                 hooks=post_build,
                 provider=self.provider,
                 context=self.context)
+
+    def _handle_missing_parameters(self, params, required_params,
+                                   stack_name=None):
```
phobologic: lets change this variable to `existing_stack_name`

stacker/actions/destroy.py
```diff
@@ -48,43 +46,33 @@ def _generate_plan(self, tail=False):
...
```
phobologic: You dont' need this any longer do you? @ttaub

stacker/actions/destroy.py
```diff
@@ -48,43 +46,33 @@ def _generate_plan(self, tail=False):
...
-        )
-        if self.provider.is_stack_destroyed(provider_stack):
-            return DestroyedStatus
-        elif self.provider.is_stack_in_progress(provider_stack):
-            return DestroyingStatus
-        else:
-            logger.debug("Destroying stack: %s", stack.fqn)
-            self.provider.destroy_stack(provider_stack)
-        return DestroyingStatus
+            return SkippedStatus('stack does not exist')
```
phobologic: It looks like we have a `StackDoesNotExistStatus` status.

stacker/exceptions.py
phobologic: @ttaub  ^^

stacker/exceptions.py
```diff
@@ -102,6 +111,34 @@ def __init__(self, stack_name, output, *args, **kwargs):
         super(OutputDoesNotExist, self).__init__(message, *args, **kwargs)
 
 
+class DestroyWithoutNotificationQueue(Exception):
+
+    def __init__(self, stack_name, *args, **kwargs):
+        self.stack_name = stack_name
+        message = "You are attempting to destroy `%s` " % stack_name
```
phobologic: btw, rather than doing the += method, you can also represent this with paranthesis,and it tends to look a bit cleaner:

```
message = (
    "You are attempting to "
    "destroy `%s`"
)
```

etc

stacker/plan.py
```diff
@@ -80,6 +79,11 @@ def set_status(self, status):
             status (:class:`Status <Status>` object): The status to set the
                 step to.
         """
+        if not isinstance(status, Status):
+            raise ValueError(
+                "Step run_func must return a valid Status object. "
```
phobologic: This error text seems out of place with the new location of this code.

stacker/plan.py
phobologic: 👍

stacker/providers/aws/default.py
```diff
@@ -1,20 +1,185 @@
...
+    CompleteStatus,
+)
 
-logger = logging.getLogger(__name__)
 
+logger = logging.getLogger(__name__)
 MAX_TAIL_RETRIES = 5
 
 
+class Message(object):
```
phobologic: Docstrings, so many docstrings.

stacker/providers/aws/default.py
```diff
@@ -1,20 +1,185 @@
...
+
+        parsed_message = self.parse_message(metadata)
+
+        for key, value in parsed_message.iteritems():
+            setattr(self, key, value)
+
+        for key, value in metadata.iteritems():
+            setattr(self, key, value)
+
+    def parse_message(self, message):
```
phobologic: Not sure what the idea is behind this - should it be returning a Message object?  If so, should this be a `classmethod`?

stacker/providers/aws/default.py
```diff
@@ -1,20 +1,185 @@
...
+
+        self.sns.set_subscription_attributes(
+            SubscriptionArn=sub_arn,
+            AttributeName="RawMessageDelivery",
+            AttributeValue="true"
+        )
+        logging.debug('Done creating cloudsns resources')
+
+    @property
+    def TopicArn(self):
```
phobologic: should be `topic_arn`, this is python!

stacker/providers/aws/default.py
```diff
@@ -1,20 +1,185 @@
...
+        if not self.sqs or not self.sns:
+            raise NotInitialized('TopicArn')
+        if not self._topicArn:
+            topic = self.sns.create_topic(Name=self.topic_name)
+            self._topicArn = topic["TopicArn"]
+
+        return self._topicArn
+
+    @property
+    def QueueUrl(self):
```
phobologic: `queue_url`

stacker/providers/aws/default.py
```diff
@@ -1,20 +1,185 @@
...
+            queue = self.sqs.create_queue(
+                QueueName=self.queue_name,
+                Attributes={"MessageRetentionPeriod": "60"}
+            )
+            self._queueUrl = queue["QueueUrl"]
+
+        return self._queueUrl
+
+    @property
+    def QueueArn(self):
```
phobologic: `queue_arn`

stacker/providers/aws/default.py
```diff
@@ -1,20 +1,185 @@
...
+                    "ArnEquals":{
+                      "aws:SourceArn":"%s"
+                    }
+                  }
+                }
+              ]
+            }""" % (self.QueueArn, self.TopicArn)
+
+    def start(self):
+        logging.debug('Creating cloudsns resources')
```
phobologic: not really called cloudsns anymore, also I think `setup` is still better than `start`, or maybe even `initialize`

phobologic: You should also not use logging directly, and instead use the logger initialized at the top of this file.

phobologic: Also, since you have a `setup` function, I'm not sure it's worth doing the property caching magic with things like the QueueArn, etc.

stacker/providers/aws/default.py
```diff
@@ -1,20 +1,185 @@
...
+        )
+
+        sub_arn = result["SubscriptionArn"]
+
+        self.sns.set_subscription_attributes(
+            SubscriptionArn=sub_arn,
+            AttributeName="RawMessageDelivery",
+            AttributeValue="true"
+        )
+        logging.debug('Done creating cloudsns resources')
```
phobologic: cloudsns

stacker/providers/aws/default.py
```diff
@@ -1,20 +1,185 @@
...
+        if not self._queueArn:
+            attr = self.sqs.get_queue_attributes(
+                QueueUrl=self.QueueUrl,
+                AttributeNames=["QueueArn"]
+            )
+            self._queueArn = attr["Attributes"]["QueueArn"]
+
+        return self._queueArn
+
+    def poll(self, user_fn):
```
phobologic: can toss this right?

stacker/providers/aws/default.py
```diff
@@ -1,20 +1,185 @@
...
+        return data
+
+
+class CloudListener(object):
+
+    def __init__(self, queue_name, topic_name, session):
+        self.session = session
+        self.queue_name = queue_name
+        self.topic_name = topic_name
+        self._topicArn = None
```
phobologic: `_topic_arn`, `_queue_url`, `_queue_arn`

stacker/providers/aws/default.py
```diff
@@ -1,20 +1,185 @@
...
+                message.delete()
+
+    def get_messages(self):
+        updates = self.sqs.receive_message(
+            QueueUrl=self.QueueUrl,
+            AttributeNames=["All"],
+            WaitTimeSeconds=20
+        )
+
+        if "Messages" in updates:
```
phobologic: Could get rid of this check with:

```
return [Message(m, self) for m in updates.get("Messages", [])]
```

If the dictionary doesn't have any Messages, it will default to `[]`

stacker/providers/aws/default.py
```diff
@@ -1,20 +1,185 @@
...
+                'ReceiptHandle': m.ReceiptHandle
+            })
+
+        self.sqs.delete_message_batch(
+            QueueUrl=self.QueueUrl,
+            Entries=receipts
+        )
+
+    def cleanup(self, delete_topic=False):
+        logging.debug('Deleting cloudsns resources')
```
phobologic: cloudsns

stacker/providers/aws/default.py
```diff
@@ -1,20 +1,185 @@
...
+
+        self.sqs.delete_queue(
+            QueueUrl=self.QueueUrl
+        )
+
+        if delete_topic:
```
phobologic: Is this ever used?

stacker/providers/aws/default.py
```diff
@@ -1,20 +1,185 @@
...
+            self.sns.delete_topic(
+                TopicArn=self.TopicArn
+            )
+
+        logging.debug('Done deleting cloudsns resources')
```
phobologic: I'd also use double quotes around the log messages.  Again, use `logger`, not `logging`.  Make sure you update this throughout the file.

stacker/providers/aws/default.py
```diff
@@ -106,105 +273,71 @@ def cloudformation(self):
...
+    @property
+    def listener(self):
+        # deals w/ multiprocessing issues w/ sharing ssl conns
+        # see https://github.com/remind101/stacker/issues/196
+        pid = os.getpid()
+        if pid != self._pid or not self._listener:
+            session = get_session(self.region)
+
+            # This stays the same for stacker call in this namespace
+            topic_name = self.namespace
```
phobologic: maybe prepend `stacker-`  just in case someone else tries to create an sns topic w/ the namespace.  Maybe not worth worrying about, but just the namespace may make people unsure of what this topic is.

stacker/providers/aws/default.py
```diff
@@ -106,105 +273,71 @@ def cloudformation(self):
...
-        except botocore.exceptions.ClientError as e:
-            if "does not exist" not in e.message:
-                raise
-            raise exceptions.StackDoesNotExist(stack_name)
+    @property
+    def listener(self):
+        # deals w/ multiprocessing issues w/ sharing ssl conns
+        # see https://github.com/remind101/stacker/issues/196
+        pid = os.getpid()
+        if pid != self._pid or not self._listener:
```
phobologic: Hmm - this is going to cause issues when/if this goes multiprocess.  I wonder if we need this code anymore, if you're removing multiprocessing?

Anyway, since it checks if `pid` is not `self._pid` you could end up with multiple listeners, using different SQS queues, if we use multiprocessing.

stacker/providers/aws/default.py
```diff
@@ -106,105 +273,71 @@ def cloudformation(self):
...
+    def poll_events(self, tail):
+        messages = self.listener.get_messages()
+        status_dict = {}
 
-    def get_stack_status(self, stack, **kwargs):
-        return stack['StackStatus']
+        for message in messages:
+            status_dict[message.StackName] = self.get_status(message)
+            if tail:
+                Provider._tail_print(message)
```
phobologic: Curious why you went with a class method for this?

stacker/providers/aws/interactive.py
```diff
@@ -127,17 +129,23 @@ def _wait_till_change_set_complete(self, change_set_id):
...
+                    'Parameters': parameters,
+                    'Tags': tags,
+                    'Capabilities': ["CAPABILITY_NAMED_IAM"],
+                    'ChangeSetName': get_change_set_name(),
+                    'NotificationARNs': [self.listener.TopicArn]
+                },
+            )
+        except botocore.exceptions.ClientError as e:
+            if "does not exist" in e.message:
+                raise exceptions.StackDoesNotExist(fqn)
```
phobologic: Need to just `raise` after this, so any other types of ClientError's don't get missed.

stacker/tests/actions/test_build.py
```diff
@@ -7,7 +7,7 @@
 from stacker.actions import build
 from stacker.actions.build import (
     _resolve_parameters,
-    _handle_missing_parameters,
+    # _handle_missing_parameters,
```
phobologic: Just delete, dont' comment it out

stacker/plan.py
ttaub: Why do you say that?

stacker/actions/build.py
```diff
@@ -231,7 +231,14 @@ def run(self, outline=False, tail=False, dump=False, *args, **kwargs):
...
     def cleanup(self):
-        logger.debug('Cleaning up')
+        """Cleans up any resources that do not need to exist after stacker
+        has finished building all the stacks.
+
+        This is specifically used when deleting the sqs queues used for
+        polling events.
+
+        """
+        logger.debug('Cleaning up sqs queues')
```
phobologic: I'd leave this a bit more generic - since this isn't actually what cleans up the queues, but the provider call itself (so putting a log message in there like this makes more sense)

stacker/plan.py
```diff
@@ -81,7 +81,7 @@ def set_status(self, status):
         """
         if not isinstance(status, Status):
             raise ValueError(
-                "Step run_func must return a valid Status object. "
+                "Step status must return a valid Status object. "
```
phobologic: This still doesn't seem quite right - maybe `"Invalid status type: %s - must be subclass of stacker.status.Status class." % type(status)` or something similar?

stacker/providers/aws/default.py
```diff
@@ -19,36 +17,41 @@
 MAX_TAIL_RETRIES = 5
 
 
+def parse_message(message):
```
phobologic: docstring

stacker/providers/aws/default.py
```diff
@@ -67,23 +70,41 @@ def create_policy(self):
                   }
                 }
               ]
-            }""" % (self.QueueArn, self.TopicArn)
+            }""" % (self.queue_arn, self.topic_arn)
 
-    def start(self):
-        logging.debug('Creating cloudsns resources')
+    def setup(self):
+        logger.debug('Creating cloudformation listener')
```
phobologic: nit: I'd use double quotes here for consistency

stacker/providers/aws/default.py
```diff
@@ -19,36 +17,41 @@
...
     def __init__(self, queue_name, topic_name, session):
         self.session = session
         self.queue_name = queue_name
         self.topic_name = topic_name
-        self._topicArn = None
-        self._queueUrl = None
-        self._queueArn = None
+        self.topic_arn = None
+        self.queue_url = None
+        self.queue_arn = None
```
phobologic: Any reason to not call `self.setup()` here?

stacker/providers/aws/default.py
```diff
@@ -67,23 +70,41 @@ def create_policy(self):
                   }
                 }
               ]
-            }""" % (self.QueueArn, self.TopicArn)
+            }""" % (self.queue_arn, self.topic_arn)
 
-    def start(self):
-        logging.debug('Creating cloudsns resources')
+    def setup(self):
```
phobologic: docstring

stacker/providers/aws/default.py
```diff
@@ -161,23 +136,20 @@ def delete_messages(self, messages):
...
         self.sqs.delete_message_batch(
-            QueueUrl=self.QueueUrl,
+            QueueUrl=self.queue_url,
             Entries=receipts
         )
 
-    def cleanup(self, delete_topic=False):
-        logging.debug('Deleting cloudsns resources')
+    def cleanup(self):
+        logger.debug('Deleting cloudsns resources')
```
phobologic: double quotes

stacker/providers/aws/default.py
```diff
@@ -161,23 +136,20 @@ def delete_messages(self, messages):
...
 
-        self.sqs.delete_queue(
-            QueueUrl=self.QueueUrl
-        )
+        # print self.__dict__
```
phobologic: delete rather than comment out

phobologic: also shouldn't this actually delete?

stacker/providers/aws/default.py
```diff
@@ -391,12 +354,32 @@ def update_stack(self, fqn, template_url, parameters, tags, **kwargs):
     def get_stack_name(self, stack, **kwargs):
         return stack['StackName']
 
-    def get_outputs(self, stack_name, *args, **kwargs):
+    def try_get_outputs(self, stack_name):
+        print "TRY GET OUTPUTS"
```
phobologic: debug? no docstring, not sure if this is used

stacker/providers/aws/default.py
```diff
@@ -391,12 +354,32 @@ def update_stack(self, fqn, template_url, parameters, tags, **kwargs):
...
+        print self
+        print stack_name
         if stack_name not in self._outputs:
+
             stack = self.get_stack(stack_name)
+
+            if 'Outputs' not in stack:
+                raise KeyError('No Outputs in Stack')
```
phobologic: double quotes

stacker/providers/aws/default.py
```diff
@@ -391,12 +354,32 @@ def update_stack(self, fqn, template_url, parameters, tags, **kwargs):
...
+
             self._outputs[stack_name] = get_output_dict(stack)
+
         return self._outputs[stack_name]
 
+    def get_outputs(self, stack_name):
```
phobologic: docstring

stacker/providers/aws/default.py
```diff
@@ -12,12 +12,11 @@
     CompleteStatus,
 )
 
-
 logger = logging.getLogger(__name__)
-MAX_TAIL_RETRIES = 5
 
 
 def parse_message(message):
+    """Parses cloudformation SNS message to grab event metdata"""
```
phobologic: Missing Args/Returns part of docstring.

stacker/actions/destroy.py
```diff
@@ -83,6 +83,17 @@ def pre_run(self, outline=False, *args, **kwargs):
...
 
+    def cleanup(self):
+        """Cleans up any resources that do not need to exist after stacker
+        has finished building all the stacks.
+
+        This is specifically used when deleting the sqs queues used for
+        polling events.
+
+        """
+        logger.debug('Cleaning up sqs queues')
```
phobologic: This isn't actually doing anything - the provider is doing the cleaning up. At most I'd say `Cleaning up` in here, but that's it.

stacker/providers/aws/default.py
```diff
@@ -49,9 +48,9 @@ def __init__(self, queue_name, topic_name, session):
         self.session = session
         self.queue_name = queue_name
         self.topic_name = topic_name
-        self.topic_arn = None
-        self.queue_url = None
-        self.queue_arn = None
+
+        # Create all the necessary cloudfromation resources
```
phobologic: probably not necessary to comment

stacker/providers/aws/default.py
```diff
@@ -73,7 +72,10 @@ def create_policy(self):
             }""" % (self.queue_arn, self.topic_arn)
 
     def setup(self):
-        logger.debug('Creating cloudformation listener')
+        """Creates the initial SNS/SQS resources for listening to
```
phobologic: I don't see `self.topic_arn` and `self.queue_url` getting set anywhere anymore, is that ok/meant to be?

stacker/providers/aws/default.py
```diff
@@ -141,15 +143,17 @@ def delete_messages(self, messages):
         )
 
     def cleanup(self):
-        logger.debug('Deleting cloudsns resources')
+        """Delete the sqs queue after each run of stacker. However
+        the sns topic does not get deleted as it follows a consistent
+        naming scheme of the form stacker-{namespace}."""
 
-        # print self.__dict__
+        logger.debug('Deleting cloud listener resources')
```
phobologic: Now here I'd put what it's actually doing.

stacker/providers/aws/default.py
```diff
@@ -355,27 +357,32 @@ def get_stack_name(self, stack, **kwargs):
         return stack['StackName']
 
     def try_get_outputs(self, stack_name):
-        print "TRY GET OUTPUTS"
-        print self
-        print stack_name
-        if stack_name not in self._outputs:
+        """Attempt to get outputs from stack.
```
phobologic: Missing Args/Return

stacker/providers/aws/default.py
```diff
@@ -355,27 +357,32 @@ def get_stack_name(self, stack, **kwargs):
...
-                raise KeyError('No Outputs in Stack')
-
+            if "Outputs" not in stack:
+                raise KeyError("No Outputs in Stack")
             self._outputs[stack_name] = get_output_dict(stack)
 
         return self._outputs[stack_name]
 
     def get_outputs(self, stack_name):
+        """Gets a given stacks outputs
```
phobologic: Missing Args/Return

phobologic:
This has drifted for too long, and has man unanswered questions.  I think we'll take a step back, and re-address when we have more time.  Sorry @ttaub - thanks for all the research/work!

PR #332, opened 2017-02-28
add ability to load python classes (blueprints, etc) directly from git

troyready:
Provides the ability to separate blueprints from config code while still explicitly referencing the code location/revision (somewhat emulating the style of https://www.terraform.io/docs/modules/sources.html)

### DIFF
--- a/stacker/context.py
+++ b/stacker/context.py
@@ -2,6 +2,7 @@ import collections
 import logging
 import sys
 
+from stacker.util import SourceProcessor
 from .config import parse_config
 from .exceptions import MissingEnvironment
 from .stack import Stack
@@ -89,6 +90,12 @@ class Context(object):
         lookups = self.config.get("lookups", {})
         for key, handler in lookups.iteritems():
             register_lookup_handler(key, handler)
+        sources = self.config.get("package_sources")
+        if sources is not None:
+            processor = SourceProcessor(
+                stacker_cache_dir=self.config.get("stacker_cache_dir")
+            )
+            processor.get_package_sources(sources=sources)
 
     def _get_stack_definitions(self):
         if not self.stack_names:
--- a/stacker/tests/test_context.py
+++ b/stacker/tests/test_context.py
@@ -1,4 +1,7 @@
+import os
+import shutil
 import sys
+import tempfile
 import textwrap
 import unittest
 
@@ -161,6 +164,32 @@ class TestContext(unittest.TestCase):
         stage = "pre_build"
         handle_hooks(stage, context.config[stage], "mock-region-1", context)
         self.assertEqual("mockResult", context.hook_data["myHook"]["result"])
+
+    def test_package_sources_in_config(self):
+        tmp_dir = tempfile.mkdtemp(prefix='stackerunittest')
+        try:
+            context = Context({"namespace": "test"})
+            config = textwrap.dedent("""\
+                stacker_cache_dir: %s
+                package_sources:
+                  git:
+                    - uri: https://github.com/remind101/stacker_blueprints.git
+                      tag: 1.0.0
+                      paths:
+                        - stacker_blueprints
+            """ % tmp_dir)
+            context.load_config(config)
+            self.assertEqual(
+                sys.path[-1],
+                os.path.join(
+                    tmp_dir,
+                    'packages',
+                    'https___github.com_remind101_stacker_blueprints-1.0.0',
+                    'stacker_blueprints'
+                )
+            )
+        finally:
+            shutil.rmtree(tmp_dir)
 
 
 class TestFunctions(unittest.TestCase):
--- a/stacker/tests/test_context.py
+++ b/stacker/tests/test_context.py
@@ -1,7 +1,4 @@
-# import os
-# import shutil
 import sys
-# import tempfile
 import textwrap
 import unittest
 
@@ -164,32 +161,6 @@ class TestContext(unittest.TestCase):
         stage = "pre_build"
         handle_hooks(stage, context.config[stage], "mock-region-1", context)
         self.assertEqual("mockResult", context.hook_data["myHook"]["result"])
-
-#     def test_package_sources_in_config(self):
-#         tmp_dir = tempfile.mkdtemp(prefix='stackerunittest')
-#         try:
-#             context = Context({"namespace": "test"})
-#             config = textwrap.dedent("""\
-#                 stacker_cache_dir: %s
-#                 package_sources:
-#                   git:
-#                     - uri: git@github.com:remind101/stacker_blueprints.git
-#                       tag: 1.0.0
-#                       paths:
-#                         - stacker_blueprints
-#             """ % tmp_dir)
-#             context.load_config(config)
-#             self.assertEqual(
-#                 sys.path[-1],
-#                 os.path.join(
-#                     tmp_dir,
-#                     'packages',
-#                     'git_github.com_remind101_stacker_blueprints-1.0.0',
-#                     'stacker_blueprints'
-#                 )
-#             )
-#         finally:
-#             shutil.rmtree(tmp_dir)
 
 
 class TestFunctions(unittest.TestCase):
--- a/stacker/tests/test_context.py
+++ b/stacker/tests/test_context.py
@@ -1,7 +1,7 @@
-import os
-import shutil
+# import os
+# import shutil
 import sys
-import tempfile
+# import tempfile
 import textwrap
 import unittest
 
@@ -165,31 +165,31 @@ class TestContext(unittest.TestCase):
         handle_hooks(stage, context.config[stage], "mock-region-1", context)
         self.assertEqual("mockResult", context.hook_data["myHook"]["result"])
 
-    def test_package_sources_in_config(self):
-        tmp_dir = tempfile.mkdtemp(prefix='stackerunittest')
-        try:
-            context = Context({"namespace": "test"})
-            config = textwrap.dedent("""\
-                stacker_cache_dir: %s
-                package_sources:
-                  git:
-                    - uri: https://github.com/remind101/stacker_blueprints.git
-                      tag: 1.0.0
-                      paths:
-                        - stacker_blueprints
-            """ % tmp_dir)
-            context.load_config(config)
-            self.assertEqual(
-                sys.path[-1],
-                os.path.join(
-                    tmp_dir,
-                    'packages',
-                    'https___github.com_remind101_stacker_blueprints-1.0.0',
-                    'stacker_blueprints'
-                )
-            )
-        finally:
-            shutil.rmtree(tmp_dir)
+#     def test_package_sources_in_config(self):
+#         tmp_dir = tempfile.mkdtemp(prefix='stackerunittest')
+#         try:
+#             context = Context({"namespace": "test"})
+#             config = textwrap.dedent("""\
+#                 stacker_cache_dir: %s
+#                 package_sources:
+#                   git:
+#                     - uri: git@github.com:remind101/stacker_blueprints.git
+#                       tag: 1.0.0
+#                       paths:
+#                         - stacker_blueprints
+#             """ % tmp_dir)
+#             context.load_config(config)
+#             self.assertEqual(
+#                 sys.path[-1],
+#                 os.path.join(
+#                     tmp_dir,
+#                     'packages',
+#                     'git_github.com_remind101_stacker_blueprints-1.0.0',
+#                     'stacker_blueprints'
+#                 )
+#             )
+#         finally:
+#             shutil.rmtree(tmp_dir)
 
 
 class TestFunctions(unittest.TestCase):
--- a/stacker/util.py
+++ b/stacker/util.py
@@ -474,14 +474,11 @@ class SourceProcessor():
         self.create_cache_directories()
 
     def create_cache_directories(self):
-        """Ensures that SourceProcessor cache directories exist.
-
-           Returns True to indicate the cache directories are present."""
+        """Ensures that SourceProcessor cache directories exist."""
         if not os.path.isdir(self.package_cache_dir):
             if not os.path.isdir(self.stacker_cache_dir):
                 os.mkdir(self.stacker_cache_dir)
             os.mkdir(self.package_cache_dir)
-        return True
 
     def get_package_sources(self, sources):
         """Makes remote python packages available for local use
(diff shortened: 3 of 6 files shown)


phobologic:
A couple of questions that arises from this:

1. Do you (& others) think it's worth allowing different stack configs to pull blueprints from the same git repo, but at different SHAs?
2. If so, it might be worth putting some effort into caching git clones that are of the same git repo at the same git SHA, just so we don't download a ton of the same thing over and over.

If those two are true, it might be worth gathering up all git repos in stack configs, and doing the clones at the very beginning, de-duping as it goes.

I think this looks pretty good though, and most of that is just implementation.

troyready:
@phobologic that does sounds reasonable. I was afraid of making it smarter to avoid having to think through any of the edge cases but faster is better.

syphon7:
@phobologic I think those are great points - both would be required for it to work with large stacks. The idea is great though, we would use that functionality extensively!

troyready:
@phobologic How about moving the repositories to the top level of the config and handle it when the context object is instantiated, somewhat similarly to the way [lookups are done](https://github.com/remind101/stacker/blob/1.0.0a6/stacker/context.py#L84-L86).

```
stacker_bucket: foobar

blueprint_sources:
  git:
    - uri: git@github.com:remind101/stacker_blueprints.git
       ref: v1.0
...
```

phobologic:
@troyready I like it!  Maybe add a `blueprint_sources.directory` (with a default) or something so you could keep a cache and not have to download the blueprints each time?

troyready:
@phobologic make sense. So the config yaml options would look something like:
```
stacker_bucket: foobar

blueprint_sources:
  directory: /tmp/stackerrepos # optional
  git:
    - uri: git@github.com:remind101/stacker_blueprints.git
       git_directory: remind101_stacker_blueprints # optional
       ref: v1.0 # optional
    - uri: git@github.com:acmecorp/stacker_blueprints.git
...
```

So the following logic should cover most use cases I think:

- If no top level directory is specified stacker will grab a random one (this means every command invocation would do a fresh clone for good and bad).
- If a directory is specified, and a subdirectory for that repo exists (e.g `/tmp/stackerrepos/stacker_blueprints` in the above example's second git entry), stacker will do a git fetch and checkout the appropriate ref in it

Any big workflow pieces I missed or any part of the design that should change?

phobologic:
That sounds good to me!  In the example above, how would I reference the first git repo vs the second in a `class_path`?

troyready:
@phobologic I think that part would be transparent - both directories would have `sys.path.append()` run on them, so the class_path would be relative to the folders in the repos.

I forget that I had one additional option in the original commit; let me see if I can mock out the whole thing now. I don't imagine it'd be particularly common to reference this many remote repos, but there shouldn't be any reason it couldn't work. I tried to make the elasticache example actually match up to the real folder structure:
```
stacker_bucket: foobar

blueprint_sources:
  directory: /tmp/stackerrepos # optional
  git:
    - uri: git@github.com:remind101/stacker_blueprints.git
       git_directory: remind101_stacker_blueprints # optional
       ref: 1.0.0 # optional
       rel: stacker_blueprints # optional
    - uri: git@github.com:acmecorp/stacker_blueprints.git
    - uri: git@github.com:contoso/webapp.git
       rel: stacker # optional

stacks:
  - name: elasticache-base
    class_path: elasticache.base.BaseReplicationGroup
  - name: some-acme-stack
    class_path: fileintheirrepo.Class
  - name: contoso-webapp
    class_path: somedirintheirepo.somefile.SomeClass
```

phobologic:
Ok, so in that case, which of those git repos would the elasticache blueprint be loaded from @troyready ?  Also not sure what `rel` is used for?

troyready:
@phobologic I was copying the rel option from [Berkshelf](https://docs.chef.io/berkshelf.html) (see `rel option` under `Git Location`).

Regarding the blueprint loading, for the example above the following will be executed when blueprint_sources is evaluated:
```
sys.path.append(os.path.join('/tmp/stackerrepos', 'remind101_stacker_blueprints'))
sys.path.append(os.path.join('/tmp/stackerrepos', 'stacker_blueprints'))
sys.path.append(os.path.join('/tmp/stackerrepos', 'webapp'))
``` 

So the elasticache blueprint should come from the first path with a elasticache folder and base file in it.

phobologic:
Ok, so in that case if I wanted a blueprint from the existing stacker_blueprints (using your example), I'd have to use a class_path like:

`class_path: stacker_blueprints.rds.postgres.Postgres` right?

That seems like it should work - I could see some collisions, but in that case the first one imported would be chosen first, which is probably fine - it's how the python path works anyway.

Though it would mean that if you did have a collision, you'd end up only being able to use the first one, right?

troyready:
@phobologic based on feedback here and further development I've updated the PoC to be mostly code complete (without tests or documentation and I'm sure some other rough edges). I've changed some of the wording to be more properly general and adopted single cache directory to keep downloads to a minimum.

Would appreciate feedback when you get a chance.

stacker/context.py
```diff
@@ -84,6 +85,9 @@ def load_config(self, conf_string):
         lookups = self.config.get("lookups", {})
         for key, handler in lookups.iteritems():
             register_lookup_handler(key, handler)
+        sources = self.config.get("package_sources", {})
```
phobologic: I'd change this block to:

```
sources = self.config.get("package_sources")
if not sources:
    get_package_sources(sources)
```

stacker/util.py
```diff
@@ -366,3 +369,88 @@ def read_value_from_path(value):
...
     return value
+
+
+def sanitize_git_path(uri, ref='HEAD'):
+    """Takes a git URI and ref and converts it to a directory safe path
+
+    Args:
+        uri (string): git URI (e.g. git@github.com:remind101/stacker_blueprints.git)
+        ref (string): optional git ref to be appended to the path
+
```
phobologic: Needs a `Returns:` docstring.

stacker/util.py
```diff
@@ -366,3 +369,88 @@ def read_value_from_path(value):
         with open(relative_path) as read_file:
             value = read_file.read()
     return value
+
+
+def sanitize_git_path(uri, ref='HEAD'):
```
phobologic: I'd change `ref` to `ref=None`, and then in the function do:

`ref = ref or "HEAD"` to set the default

stacker/util.py
```diff
@@ -366,3 +369,88 @@ def read_value_from_path(value):
...
+def fetch_git_package(config, package_cache):
+    """Makes a remote git repository available for local use
+
+    Args:
+        config (dict): Dictionary of git repo configuration
+        package_cache (string): Directory in which to store cached packages
+
+    """
+    # Sanitize directory name
+    if 'ref' in config:
```
phobologic: If you do the stuff above, then you could change this entire block to:

```
dir_name = sanitize_git_path(uri=config["uri"], ref=config.get("ref"))
```

stacker/util.py
```diff
@@ -366,3 +369,88 @@ def read_value_from_path(value):
...
+        dir_name = sanitize_git_path(uri=config['uri'], ref=config['ref'])
+    else:
+        dir_name = sanitize_git_path(uri=config['uri'])
+    cached_dir_path = os.path.join(package_cache, dir_name)
+
+    # Clone the repo if it doesn't already exist; otherwise refresh it
+    if not os.path.isdir(cached_dir_path):
+        tmp_dir = tempfile.mkdtemp(prefix='stacker')
+        tmp_repo_path = os.path.join(tmp_dir, dir_name)
+        repo = Repo.clone_from(config['uri'], tmp_repo_path)
```
phobologic: May want to throw all of this in a `try/finally` block to clean up the `tmp_repo_path`

stacker/util.py
```diff
@@ -366,3 +369,88 @@ def read_value_from_path(value):
...
+          {'git': [{'uri': 'git@github.com:remind101/stacker_blueprints.git',
+                    'ref': '1.0.0',
+                    'paths': ['stacker_blueprints']},
+                   {'uri': 'git@github.com:acmecorp/stacker_blueprints.git'},
+                   {'uri': 'git@github.com:contoso/webapp.git',
+                    'ref': 'staging'}]}
+
+    """
+    # First, ensure the cache directory exists
+    stacker_cache = os.path.join(os.environ['HOME'], '.stacker')
```
phobologic: Probably worth making this configurable in config, and then default to this in the case that it's not given.

ejholmes:
Just to play devils advocate. What advantage does this provide over just adding packages to a setup.py? Seems that setuptools [can already pull directly from github](http://stackoverflow.com/questions/3472430/how-can-i-make-setuptools-install-a-package-thats-not-on-pypi).

troyready:
@ejholmes I think this will open up a significantly better workflow, where specific commits/tags are referenced per environment. Then we can have a promotion process, confidence in what version is being referenced, etc

troyready:
@phobologic thanks again for the feedback. All specific points should be addressed now.

I'm not 100% sure I've picked the correct name for the now-configurable stacker cache directory option, but that's obviously easy to change.

Unless I get other feedback, I'll proceed to docs/tests/squashing when I get a chance.

phobologic:
Sorry @troyready , meant to get to this this weekend, but then weekend happened.  I'll try to get to this in the next day or so.

phobologic:
@troyready oh jeez, I'm really sorry - I thought I'd already approved this PR - do you want to go ahead and write docs/tests, and I'll pull this in as soon as you are done?  Sorry about that.

troyready:
@phobologic appreciate it, no apology needed. The core ideas are good here but there's still some additional rework I want to do so it's very much still waiting for me to move it forward.

phobologic:
Cool - no worries - I'll probably do a PR later today (if I can clean up the in-flight PRs and get them merged), but if it doesn't make it into this release it can definitely be in the next :)

troyready:
@phobologic I've rebased this against master and finished my preliminary testing/docs. The caching should be significantly better optimized now (e.g. users that specify a tag or commit will skip right by any remote repo access after the initial download).

In addition to normal review/feedback, I could use some help with the test development now. I know there's gotta be a simple way to handle:

1) Invoking the new `SourceProcessor` class while overriding its `create_cache_directories` method (so directories don't get created on the real filesystem)
2) Mocking out the subprocess calls to `git ls-remote`

...but I'm not sure how to do it. If any has any directions on them it'd be much appreciated.

troyready:
I've updated the PR with some basic testing of the helper methods. I tried adding an actual execution of the retrieval process as well (still in the commit but commented out) but the shelling out to git on CircleCI was failing.

Not sure if it makes sense to pursue more testing along those lines, or if it's too fair outside the lines of unit testing to pursue anyway.

stacker/util.py
```diff
@@ -448,3 +452,168 @@ def ensure_s3_bucket(s3_client, bucket_name, bucket_region):
...
+        package_cache_dir = os.path.join(stacker_cache_dir, 'packages')
+        self.stacker_cache_dir = stacker_cache_dir
+        self.package_cache_dir = package_cache_dir
+        self.create_cache_directories()
+
+    def create_cache_directories(self):
+        """Ensures that SourceProcessor cache directories exist.
+
+           Returns True to indicate the cache directories are present."""
+        if not os.path.isdir(self.package_cache_dir):
```
phobologic: Using `os.isdir` will only check that the path is a directory - if it's a file, or link, it'll still try and create the directory (and fail, so maybe this is fine) - just wanted to make sure you've thought of this.

stacker/util.py
```diff
@@ -448,3 +452,168 @@ def ensure_s3_bucket(s3_client, bucket_name, bucket_region):
...
+            if not os.path.isdir(self.stacker_cache_dir):
+                os.mkdir(self.stacker_cache_dir)
+            os.mkdir(self.package_cache_dir)
+        return True
```
phobologic: Shouldn't be necessary.

stacker/util.py
```diff
@@ -448,3 +452,168 @@ def ensure_s3_bucket(s3_client, bucket_name, bucket_region):
...
+            raise ValueError("Ref \"%s\" not found for repo %d." % (ref, uri))
+
+    def determine_git_ls_remote_ref(self, config):
+        """Takes a dict describing a git repo and determines the ref to be used
+           with the "git ls-remote" command
+
+        Args:
+            config (dict): git config dictionary; 'branch' key is optional
+
+        Returns: string (a branch reference or "HEAD")
```
phobologic: Can you split this up, and make it more like other docs?  This would be:

```
Returns:
    str: A branch reference or "HEAD"
```

This should happen in a few places in this PR.

stacker/tests/test_context.py
```diff
@@ -162,6 +165,32 @@ def test_hook_with_sys_path(self):
         handle_hooks(stage, context.config[stage], "mock-region-1", context)
         self.assertEqual("mockResult", context.hook_data["myHook"]["result"])
 
+#     def test_package_sources_in_config(self):
```
phobologic: Go ahead and remove all these commented out bits please.

stacker/util.py
```diff
@@ -448,3 +452,168 @@ def ensure_s3_bucket(s3_client, bucket_name, bucket_region):
...
+    def fetch_git_package(self, config):
+        """Makes a remote git repository available for local use
+
+        Args:
+            config (dict): Dictionary of git repo configuration
+
+        """
+        # If a specific commit or tag is defined, first check to see if it is
+        # already cached.
+        if config.get('commit'):
```
phobologic: Worth turning this whole block into it's own function/method and writing some tests around it?

Also, do you think it's worth catching if someone specifies both a commit and tag?

stacker/util.py
```diff
@@ -448,3 +452,168 @@ def ensure_s3_bucket(s3_client, bucket_name, bucket_region):
...
+                config['uri'],
+                self.determine_git_ls_remote_ref(config)
+            )
+            dir_name = self.sanitize_git_path(uri=config['uri'],
+                                              ref=ref)
+
+        cached_dir_path = os.path.join(self.package_cache_dir, dir_name)
+
+        # Clone the repo if it doesn't already exist
+        if not os.path.isdir(cached_dir_path):
```
phobologic: What happens here if I change the tag in the config?  Will the cached dir be updated?  It doesn't seem like it here, but I may be missing something.

troyready:
@phobologic good thoughts - I think I've updated each requested item. Happy to adjust further or squash the commits.

stacker/util.py
troyready: I think that's fine here -- seems like it won't come up unless someone does something very exceptional and then they should know how to fix it based on the raised exception

stacker/util.py
troyready: A changed tag will be treated as an entirely new package source (via a new directory).

This is possibly too cautious (at the expense of disk space), but I think the current logic should be very predictable & stable (even in situations like the remote repo's history changing)

phobologic: Awesome, sounds good to me.

PR #341, opened 2017-03-09
Adding sturdy to docs/organizations_using_stacker.rst

GarisonLotus:
I didn't know how you wanted to keep this list going, so I just duplicated what you had for Remind. We will ask some of our clients if we can include them as well create more PRs to add them later as permission is obtained.


phobologic:
Awesome, thanks!

PR #359, opened 2017-03-30
Better handling of rollback errors with stacks

koblas:
In the process of trying to get stacker working, I found that it didn't handle ROLLBACK states very well. When a stack fails to create you end up having to go to the console to delete it, rather than having stacker understand that the stack is "not-real".

Also, this covers more states in CF when displaying update progress.

### DIFF
--- a/stacker/actions/base.py
+++ b/stacker/actions/base.py
@@ -70,6 +70,7 @@ class BaseAction(object):
     def ensure_cfn_bucket(self):
         """The CloudFormation bucket where templates will be stored."""
         try:
+            logger.debug("Head on bucket %s", self.bucket_name)
             self.s3_conn.head_bucket(Bucket=self.bucket_name)
         except botocore.exceptions.ClientError as e:
             if e.response['Error']['Message'] == "Not Found":
@@ -79,8 +80,13 @@ class BaseAction(object):
                 logger.exception("Access denied for bucket %s.",
                                  self.bucket_name)
                 raise
+            elif e.response['Error']['Message'] == "Bad Request":
+                logger.exception(
+                    "Bucket name in used by another AWS customer bucket=%s.",
+                    self.bucket_name)
+                raise
             else:
-                logger.exception("Error creating bucket %s. Error %s",
+                logger.exception("Error checking bucket %s. Error %s",
                                  self.bucket_name, e.response)
                 raise
 
--- a/stacker/actions/build.py
+++ b/stacker/actions/build.py
@@ -15,6 +15,7 @@ from ..status import (
     DidNotChangeStatus,
     SubmittedStatus,
     CompleteStatus,
+    PENDING,
     SUBMITTED
 )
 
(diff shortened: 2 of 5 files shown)


stacker/actions/base.py
```diff
@@ -79,8 +80,13 @@ def ensure_cfn_bucket(self):
                 logger.exception("Access denied for bucket %s.",
                                  self.bucket_name)
                 raise
+            elif e.response['Error']['Message'] == "Bad Request":
```
phobologic: Is this the correct case?  When another account is using a bucket name, I get the following:

```
>>> c.head_bucket(Bucket=USED_BUCKET)
  File "<stdin>", line 1, in <module>
  File "/Users/mike/venv/boto3/lib/python2.7/site-packages/botocore/client.py", line 251, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/Users/mike/venv/boto3/lib/python2.7/site-packages/botocore/client.py", line 537, in _make_api_call
    raise ClientError(parsed_response, operation_name)
botocore.exceptions.ClientError: An error occurred (403) when calling the HeadBucket operation: Forbidden
```

stacker/actions/build.py
```diff
@@ -192,18 +193,37 @@ def _launch_stack(self, stack, **kwargs):
...
-                submit_reason = getattr(old_status, "reason", None)
-                return CompleteStatus(submit_reason)
-            elif self.provider.is_stack_in_progress(provider_stack):
-                logger.debug("Stack %s in progress.", stack.fqn)
-                return old_status
+
+        logger.debug("Stack %s old_status=%r  provider=%r",
+                     stack.fqn, old_status, provider_stack)
+
+        if provider_stack:
```
phobologic: This logic seems scary - especially using `CompleteStatus` to mean that a stack is in rollback, since it would continue  executing other stacks while the stack was in a rollback state, and in most cases I think people would want to stop immediately rather than continue.

stacker/actions/build.py
```diff
@@ -192,18 +193,37 @@ def _launch_stack(self, stack, **kwargs):
...
-                self.provider.get_stack_status(provider_stack),
-            )
-            if self.provider.is_stack_completed(provider_stack):
-                submit_reason = getattr(old_status, "reason", None)
-                return CompleteStatus(submit_reason)
-            elif self.provider.is_stack_in_progress(provider_stack):
-                logger.debug("Stack %s in progress.", stack.fqn)
-                return old_status
+
+        logger.debug("Stack %s old_status=%r  provider=%r",
```
phobologic: I wouldn't call this `provider=%r` rather `existing_stack` maybe?

stacker/actions/build.py
```diff
@@ -192,18 +193,37 @@ def _launch_stack(self, stack, **kwargs):
...
+                    return CompleteStatus("Failed to create/update stack")
+            elif old_status == PENDING:
+                if self.provider.is_stack_rollback(provider_stack):
+                    logger.info("Stack %s in rollback - can't start",
+                                stack.fqn)
+                    return CompleteStatus("Failed to create/update stack")
+                elif self.provider.is_stack_rollback_complete(provider_stack):
+                    logger.info("Stack %s failed to create, deleting",
+                                stack.fqn)
+                    self.provider.destroy_stack(provider_stack)
```
phobologic: I *really* don't feel comfortable with this logic - assuming that the right thing to do is to destroy the stack out of hand, without at least confirmation/etc is really dangerous here.  When a stack goes into a rollback, the usual use case is to examine why the stack broke so you can fix it.

phobologic: It's extra strange here because since the `old_status` is in `PENDING`, that means that this instantiation of `stacker` hasn't yet tried to execute the stack - so some other thing must have tried to launch that stack, and it seems dangerous to have 2 tools working on this at the same time.

phobologic:
Ping @koblas

Issue #350, opened 2017-03-17
New Lookup: ami


discobean:
It would be great to have a new lookup like `{ami filter=xyz}` or something similar.

The filter will fetch the lastest AMI ID with a specific search.

For continuity it would be good to build it so the filter is similar that to the one available already in terraform.

phobologic:
Sounds like a useful lookup - not sure we'll have time ourselves to try and get this in for 1.1, but if anyone else is up for it I'd definitely be up for the PR!

discobean:
Sure, I can have a look when we get around to automating this part.

danielkza:
I would love to have this, since I'm working on a [packer](https://packer.io) build hook to generate AMIs.

phobologic:
Closing since #360 was merged!

PR #360, opened 2017-04-04
Feature/ami

discobean:
This adds an `${ami xxx}` feature that resolves Issue #350 

It still needs to be:
1. documented
2. tests completed
3. reviewed

Apart from that the lookup is working for us in prod.


phobologic:
Hey @discobean - it looks like the change to the regexp actually caused an issue with nested regular expressions (it causes it to hang forever in the test: 

```
test_variable_resolve_nested_lookup (stacker.tests.test_variables.TestVariables
```

Doing nested regexps is a pain, really need to find a better way to do this.

discobean:
Hi mate, I've just added [] to the regex, which supports our use case for the AMI lookup

phobologic:
@discobean I went ahead and added some tests & docs for this lookup.  Can you take a look and let me know if that looks right to you?  Thanks!

stacker/lookups/__init__.py
```diff
@@ -15,7 +15,7 @@
                                        # space
 ?\s*                                   # any number of spaces separating the
                                        # type from the input
-(?P<input>[@\+\/,\._\-a-zA-Z0-9\:\s=]+) # the input value to the lookup
+(?P<input>[@\+\/,\._\-a-zA-Z0-9\:\s=\[\]\*]+) # the input value to the lookup
```
ralph-tice: i think this deserves a little more explanation or to be broken up further

phobologic: Honestly, @ralph-tice, I'd like to do away with that regex someday.  I've attempted to refactor the way we parse lookups in configs a few times now, and it's surprisingly hard for a few reasons:

- we allow nested lookups
- we allow multiple lookups per variable
- we allow lookups to return non-text values (basically any python object)
- lookups can actually affect the way the external code works (only the `output` lookup at this point, which is used to determine dependencies)

This results in it being shockingly complex to use one of the existing parsers out there.  At some point I'll give it another shot and hopefully be more successful.

phobologic:
ping @discobean

discobean:
Thanks @phobologic been away and didn't realised you brought this in.  Much appreciated.

phobologic:
NP @discobean - thank you for the PR :)

PR #368, opened 2017-04-25
Allow piping a config to stacker build/destroy/diff

ejholmes:
A small UNIX friendly change.

If you want to do some pre-processing, or use something to generate a config, this can make that easier:

```console
$ ./gen_config.py | stacker build prod.env -
```


stacker/commands/stacker/base.py
```diff
@@ -141,8 +141,10 @@ def add_arguments(self, parser):
                  "\"namespace\".")
         parser.add_argument(
             "config", type=argparse.FileType(),
+            nargs='?',
```
phobologic: Why is this necessary?  Usually I've seen this if a `default` is set?

phobologic: Ping!

ejholmes: From my limited understanding of argparse, if `type` is `FileType` and the argument is `-` then it uses `sys.stdin`. [From the docs](https://docs.python.org/3/library/argparse.html):

> One of the more common uses of nargs='?' is to allow optional input and output files:

```
>>>
>>> parser = argparse.ArgumentParser()
>>> parser.add_argument('infile', nargs='?', type=argparse.FileType('r'),
...                     default=sys.stdin)
>>> parser.add_argument('outfile', nargs='?', type=argparse.FileType('w'),
...                     default=sys.stdout)
>>> parser.parse_args(['input.txt', 'output.txt'])
Namespace(infile=<_io.TextIOWrapper name='input.txt' encoding='UTF-8'>,
          outfile=<_io.TextIOWrapper name='output.txt' encoding='UTF-8'>)
>>> parser.parse_args([])
Namespace(infile=<_io.TextIOWrapper name='<stdin>' encoding='UTF-8'>,
          outfile=<_io.TextIOWrapper name='<stdout>' encoding='UTF-8'>)
```

> FileType objects understand the pseudo-argument '-' and automatically convert this into sys.stdin for readable FileType objects and sys.stdout for writable FileType objects:

phobologic:
I think these are two separate things - without the nargs change, `FileType` already accepts `-` to say "accept from stdin".  With nargs, you're saying that the stacker.yaml is optional, and if it's not provided then it should read from stdin.  I'm less comfortable with that - I think at some point we'll make the stack config optional, but if it's not defined we'll use a default like `stacker.yaml`, rather than stdin, since that's the much more common use case.

phobologic:
Just verified the `-` already works in master:
```
# cat ./stacker.yml | stacker build --region us-west-1 stacker.env -
[2017-07-06T13:40:00] Using Default AWS Provider
[2017-07-06T13:40:34] Plan Status:
[2017-07-06T13:40:34]     bucket-us-east-stack-us-west-bucket: complete (creating new stack)
# cat ./stacker.yml | stacker destroy --force --region us-west-1 stacker.env -
[2017-07-06T13:41:06] Using Default AWS Provider
[2017-07-06T13:41:32] Plan Status:
[2017-07-06T13:41:32]     bucket-us-east-stack-us-west-bucket: complete (stack destroyed)
```

I think your update to the docs is useful though, and I'm going to go ahead and include that.

PR #372, opened 2017-04-30
Make lambda hook deterministic

### LINKED ISSUE #370: Content hash in aws_lambda hook non-deterministic?
Need to investigate this further, but we've noticed that the hash seems to change periodically, or across machines. I think this _might_ be because we hash the zip file itself, which may be including some extra metadata that changes.

Maybe we should hash the included files themselves? Basically:

1. Filter/Exclude files
2. Sort the file names.
3. Hash the file content.
4. Hash the combination of the file hash, and file name

We use this algorithm in another internal build system to produce hashes of a file system that's deterministic across machines. In bash:

```console
find . -L -type f | sort | xargs sha1sum | sha1sum | cut -d ' ' -f 1
```

phobologic:
The old method would cause issues if the mtime of the files was
different, since it was using the hash of the zipfile (and the zipfile
includes headers for mtime for each file).

This relies on generating a hash of all the files to be uploaded, then
hashes all those hashes. As long as the content hasn't changed, it
should return the same hash each time.

This also stops relying on the ETAGs from s3, and instead just uses the
calculated hash to create the key name, and doesn't re-upload if the key
already exists.

Fixes #370

### DIFF
--- a/stacker/hooks/aws_lambda.py
+++ b/stacker/hooks/aws_lambda.py
@@ -78,14 +78,16 @@ def _calculate_hash(files, root):
     Returns:
         str: A hash of the hashes of the given files.
     """
-    file_hashes = []
+    file_hash = hashlib.md5()
     for fname in files:
         f = os.path.join(root, fname)
+        file_hash.update(fname + "\0")
         with open(f, "rb") as fd:
-            file_hashes.append("%s:%s" % (fname,
-                                          hashlib.md5(fd.read()).hexdigest()))
+            for chunk in iter(lambda: fd.read(4096), ""):
+                file_hash.update(chunk)
+            file_hash.update("\0")
 
-    return hashlib.md5(' '.join(file_hashes)).hexdigest()
+    return file_hash.hexdigest()
 
 
 def _find_files(root, includes, excludes):
--- a/stacker/hooks/aws_lambda.py
+++ b/stacker/hooks/aws_lambda.py
@@ -36,6 +36,7 @@ def _zip_files(files, root):
 
     Returns:
         str: content of the ZIP file as a byte string.
+        str: A calculated hash of all the files.
 
     """
     zip_data = StringIO()
@@ -61,8 +62,29 @@ def _zip_files(files, root):
 
     contents = zip_data.getvalue()
     zip_data.close()
+    content_hash = _calculate_hash(files, root)
 
-    return contents
+    return contents, content_hash
+
+
+def _calculate_hash(files, root):
+    """ Returns a hash of all of the given files at the given root.
+
+    Args:
+        files (list[str]): file names to include in the hash calculation,
+            relative to ``root``.
+        root (str): base directory to analyze files in.
+
+    Returns:
+        str: A hash of the hashes of the given files.
+    """
+    file_hashes = []
+    for fname in files:
+        f = os.path.join(root, fname)
+        with open(f) as fd:
+            file_hashes.append(hashlib.md5(fd.read()).hexdigest())
+
+    return hashlib.md5(' '.join(file_hashes)).hexdigest()
 
 
 def _find_files(root, includes, excludes):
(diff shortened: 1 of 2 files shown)


phobologic:
@danielkza for some reason it's not letting me add you as a reviewer, maybe after mentioning you it will?

stacker/hooks/aws_lambda.py
```diff
@@ -61,8 +62,29 @@ def _zip_files(files, root):
...
+        root (str): base directory to analyze files in.
+
+    Returns:
+        str: A hash of the hashes of the given files.
+    """
+    file_hashes = []
+    for fname in files:
+        f = os.path.join(root, fname)
+        with open(f) as fd:
+            file_hashes.append(hashlib.md5(fd.read()).hexdigest())
```
danielkza: a) It is probably a good idea to open the file with `rb` mode to avoid newline shenanigans.
b) This might not catch file renames if the files are still produced in the same order.

It might be necessary to sort files by their full path in the ZIP file, and prepend the path to the content before hashing.

stacker/tests/hooks/test_lambda.py
```diff
@@ -290,11 +288,29 @@ def test_idempotence(self):
...
+        with self.temp_directory_with_files() as d2:
+            root = d2.path
+            hash2 = _calculate_hash(ALL_FILES, root)
+
+        with self.temp_directory_with_files() as d3:
+            root = d3.path
+            with open(os.path.join(root, ALL_FILES[0]), "w") as fd:
+                fd.write("modified file data")
+            hash3 = _calculate_hash(ALL_FILES, root)
+
```
danielkza: It would be nice to add a test verifying that the same set of files with the same content, but different file names generate a different hash.

phobologic:
Thanks @danielkza  - updated to respond to your comments!

stacker/hooks/aws_lambda.py
```diff
@@ -81,8 +81,9 @@ def _calculate_hash(files, root):
     file_hashes = []
     for fname in files:
         f = os.path.join(root, fname)
-        with open(f) as fd:
-            file_hashes.append(hashlib.md5(fd.read()).hexdigest())
+        with open(f, "rb") as fd:
+            file_hashes.append("%s:%s" % (fname,
+                                          hashlib.md5(fd.read()).hexdigest()))
```
danielkza: One last thing (sorry that I forgot to mention earlier): it's probably a bit more efficient to use a single hash object and incrementally append to it. That makes it easy to use `\0` to separate path and content, which is better as it is almost always disallowed in filenames, and not read the whole file to memory.

```python
hash = hashlib.md5()
for fname in files:
    f = os.path.join(root, fname)
    with open(f, "rb") as fd:
        hash.update(fname + "\0")
        for chunk in iter(lambda: f.read(4096), ''):
            hash.update(chunk)

return hash.hexdigest()

**Comment by danielkza** [comment]
LGTM, thanks for bearing with my nitpicks :)

File: stacker/hooks/aws_lambda.py [hunk 4]
```diff
@@ -61,8 +62,32 @@ def _zip_files(files, root):
...
+    Args:
+        files (list[str]): file names to include in the hash calculation,
+            relative to ``root``.
+        root (str): base directory to analyze files in.
+
+    Returns:
+        str: A hash of the hashes of the given files.
+    """
+    file_hash = hashlib.md5()
+    for fname in files:
```
ejholmes: Do you know if [_find_files](https://github.com/remind101/stacker/blob/ef299e02aaf9e770d55b102092a401c812568fe9/stacker/hooks/aws_lambda.py#L93) returns a sorted list? Should probably perform a sort on the files before iterating through them, since non-deterministic ordering will screw up the hash.

ejholmes: And maybe add a test for this as well.

phobologic:
@ejholmes updated to sort the file lists before calculating them and wrote a test to verify that if calculate gets a file list that is in different order the hashes still end up the same.

Issue #374, opened 2017-05-02
Non us-east regions have issue with create_bucket


evanjaaahn:
Looks like this is a regression back to #17 - ensure_cfn_bucket() in stacker.actions.base doesn't specify the location if the provider region isn't us-east-1.

phobologic:
Hey @evanjaaahn - sorry for the delay, but I just merged a PR that should fix this for you.  Thanks for the Issue, and sorry for taking so long!

phobologic:
BTW, planning to do a new release with this fix today/tomorrow.

PR #378, opened 2017-05-03
Update hooks to use the provider for aws interactions

phobologic:
Rather than setting up their own sessions, we already pass the provider
in.  By standardizing on using the provider for all interactions, we can
share logic between all the connections, which I think is beneficial.

In doing this I discovered that the mock_provider really wasn't
necessary any longer, and we could use an actual Provider class for all
our tests.

This is all in preparation for trying to get to #374, and make it less
likely that we'll regress once again.

### DIFF
--- a/stacker/actions/base.py
+++ b/stacker/actions/base.py
@@ -2,7 +2,6 @@ import copy
 import logging
 
 import botocore.exceptions
-from stacker.session_cache import get_session
 
 logger = logging.getLogger(__name__)
 
@@ -62,8 +61,7 @@ class BaseAction(object):
     def s3_conn(self):
         """The boto s3 connection object used for communication with S3."""
         if not hasattr(self, "_s3_conn"):
-            session = get_session(self.provider.region)
-            self._s3_conn = session.client('s3')
+            self._s3_conn = self.provider.s3
 
         return self._s3_conn
 
--- a/stacker/actions/base.py
+++ b/stacker/actions/base.py
@@ -69,20 +69,7 @@ class BaseAction(object):
 
     def ensure_cfn_bucket(self):
         """The CloudFormation bucket where templates will be stored."""
-        try:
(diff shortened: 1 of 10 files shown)


GarisonLotus:
@phobologic - Mike, Any specific reason (besides the conflicts) why this PR is stuck?

I ran into the issue where in 1.0.2 I was using the aws_lambda.upload_lambda_functions prehook and deploying everything to us-east-2, but the S3 bucket for the lambda artifact ended up being created in us-east-1. This caused a rollback in my lambda stack. It's not blocking or anything (easy fix to just create a bucket in the correct region) but I was going to create an issue when I saw you already had one and a PR already existed.

... I don't mind helping with it. Let me know.

phobologic:
Hey @GarisonLotus - I think the big hurdle I hit was trying to be forward thinking.  Like, in the future, if stacker is multi-region, how would this look?  It made me think that maybe it's worth just getting away from these cached AWS API connections, and instead have each use request a new one when it's necessary - that way the region, etc, could be specified each time.  Any thoughts on that?

GarisonLotus:
@phobologic - I think your forward thinking approach / idea is neat (and would set Stacker apart from some other IaC tools).  Multi-Region would be amazing and not using cached connections would solve the problem I ran into.  Is that what this PR is for? (I didn't actually check the commits you made, since I was in the middle of a deployment when I ran into this.)

phobologic:
This started as a PR to fix the issues we have in #374, though when I started to think about the future I got derailed, and started to think that maybe the method in this PR wasn't the right approach.  That's basically where I left it - meaning to get back to it, but other tasks have taken higher priority (not to mention tons of visiting guests on the weekend - ready to have a weekend where I can sit at home and geek out on stacker code, it's been a while :))

phobologic:
Closing for now - I'm going to have to rethink this with multi-region in mind.

PR #379, opened 2017-05-04
Add support for Property objects in TroposphereType variables

danielkza:
Previously TroposphereType only worked for complete resources, not
property objects meant to be used as parameters for other resources.

Change that by inspecting the type to determine if it represents a property
object, and interpreting the variable value differently. A dictionary is
used directly as parameters, instead as title/parameters pairs, and a list
can be provided to generate multiple objects (if the type definition
allows).

Also add an `optional` boolean parameter to the constructor (mutually
exclusive to `many`), indicating that a single object should be created,
but that a `None` (or ommited) value is acceptable, in which case it is
passed through to the template.

### DIFF
--- a/stacker/blueprints/variables/types.py
+++ b/stacker/blueprints/variables/types.py
@@ -8,10 +8,28 @@ class TroposphereType(object):
         :class:`Troposphere` will convert the value provided to the variable to
         the specified Troposphere type.
 
+        Both resource and parameter classes (which are just used to configure
+        other resources) are acceptable as configuration values.
+
+        Complete resource definitions must be dictionaries, with the keys
+        identifying the resource titles, and the values being used as the
+        constructor parameters.
+
+        Parameter classes can be defined as dictionariy or a list of
+        dictionaries. In either case, the keys and values will be used directly
+        as constructor parameters.
+
         Args:
             defined_type (type): Troposphere type
-            many (Optional[bool]): Boolean indicating whether or not the type
-                accepts a single object or a list of objects.
+            many (Optional[bool]): Boolean indicating whether or not multiple
+                resources can be constructed. If the defined type is a
+                resource, multiple resources can be passed as a dictionary of
+                dictionaries
+                If it is a parameter class, multiple resources are passed as
+                a list.
+            optional (Optional[bool]): Boolean indicating if an undefined/null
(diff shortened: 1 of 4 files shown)


stacker/blueprints/variables/types.py
```diff
@@ -35,17 +45,37 @@ def create(self, value):
...
+                raise ValueError("Resources must be specified as a dict of "
+                                 "title to parameters")
+            if not self._many and len(value) > 1:
+                raise ValueError("Only one resource can be provided for this "
+                                 "TroposphereType variable")
+
+            result = [self._type.from_dict(title, v) for title, v in
+                      value.items()]
+        else:
+            # Our type is for properties, not a resource, so don't use
```
danielkza: Yes, I couldn't think of a better way to handle it. Having different types of input is confusing, but having them differ from what you would pass to the actual resource would be even more confusing IMO. I'll update the docs and try to explain things as best as I can.

danielkza:
Updated with more descriptive docs.

danielkza:
PEP8 issues fixed with an amend.

danielkza:
Rebased to the latest master and added a bit more documentation about the new options.

stacker/blueprints/variables/types.py
```diff
@@ -2,50 +2,100 @@
...
-            raise ValueError(message)
-
-        output = [self._type.from_dict(title, values) for title, values in
-                  value.items()]
-
-        return output if self._many else output[0]
+        if self._optional and not value:
+            return None
+
+        if hasattr(self._type, 'resource_type'):
```
phobologic: This should probably be:

`if isinstance(self._type, troposphere.AWSObject):`

danielkza: Making this change seems to break some tests. Unfortunately Troposphere seems not to be very consistent with the usage of it's base classes.

danielkza:
Updated user docs.

stacker/blueprints/variables/types.py
danielkza: This seems to break some tests. Unfortunately it seems there are some inconsistencies in troposphere.

phobologic:
Awesome, thanks for all your work on this, @danielkza!  Sorry for the delay in merging!

PR #385, opened 2017-05-15
add debugging statements to sys.path appending

troyready:
When module loading goes awry, it's nice to be able to easily see this data on what was supposed to happen vs what actually happened.

### DIFF
--- a/stacker/context.py
+++ b/stacker/context.py
@@ -1,10 +1,13 @@
 import collections
+import logging
 import sys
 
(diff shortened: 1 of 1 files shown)


phobologic:
@troyready looks good, small flake8 error.  Thanks!

troyready:
@phobologic thanks for the heads-up -- corrected.

phobologic:
Thanks!

Issue #387, opened 2017-05-16
stacker build without defining a variable referenced in yaml produces `UnboundLocalError` error without much info


GarisonLotus:
The `UnboundLocalError` is actually the intended behavior here and most of us would never run into this... However this standard error means nothing to someone who doesn't know python, not familiar with standard python errors, or is not able to dig into the code. A nicer error would be l nice. This issue is here so we can make this error a little nicer for someone launching a blueprint and yaml, but forgetting to set a variable. 

Actual message received: 
15:21:55 UnboundLocalError: local variable 'value' referenced before assignment

To reproduce:  Just do a stacker build and forget to set a variable referenced. 
Stack Trace to help find where to catch this: Attached... but here is a good start: base.py, line 214, in resolve_variable

[UnboundLocalError_trace.txt](https://github.com/remind101/stacker/files/1006100/UnboundLocalError_trace.txt)

phobologic:
```
Traceback (most recent call last):
  File "/app/jenkins/workspace/OPS_Provision-hosts-stacker-c3-equivalent_Build/venv/bin/stacker", line 9, in <module>
    args.run(args)
  File "/app/jenkins/workspace/OPS_Provision-hosts-stacker-c3-equivalent_Build/venv/lib/python2.7/site-packages/stacker/commands/stacker/build.py", line 52, in run
    dump=options.dump)
  File "/app/jenkins/workspace/OPS_Provision-hosts-stacker-c3-equivalent_Build/venv/lib/python2.7/site-packages/stacker/actions/base.py", line 124, in execute
    self.run(*args, **kwargs)
  File "/app/jenkins/workspace/OPS_Provision-hosts-stacker-c3-equivalent_Build/venv/lib/python2.7/site-packages/stacker/actions/build.py", line 283, in run
    plan.execute()
  File "/app/jenkins/workspace/OPS_Provision-hosts-stacker-c3-equivalent_Build/venv/lib/python2.7/site-packages/stacker/plan.py", line 267, in execute
    if not self._single_run():
  File "/app/jenkins/workspace/OPS_Provision-hosts-stacker-c3-equivalent_Build/venv/lib/python2.7/site-packages/stacker/plan.py", line 227, in _single_run
    status = step.run()
  File "/app/jenkins/workspace/OPS_Provision-hosts-stacker-c3-equivalent_Build/venv/lib/python2.7/site-packages/stacker/plan.py", line 74, in run
    return self._run_func(self.stack, status=self.status)
[...]

PR #388, opened 2017-05-17
Attempting to catch undefined variable value - Issue #387

GarisonLotus:
and passing that to the ValidatorError function.

This resolves the specific scenario where a user defines the key, but not the value of a variable. This was causing an "UnboundLocalError" when attempting to pass in the "value" to the ValidatorError function.  

This resolves issue #387

### DIFF
--- a/stacker/blueprints/base.py
+++ b/stacker/blueprints/base.py
@@ -211,6 +211,10 @@ def resolve_variable(var_name, var_def, provided_variable, blueprint_name):
     try:
         value = validator(value)
     except Exception as exc:
+        try:
+            value
+        except UnboundLocalError:
+            value = "mising or undefined"
         raise ValidatorError(var_name, validator.__name__, value, exc)
 
     # Ensure that the resulting value is the correct type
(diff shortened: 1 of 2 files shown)


troyready:
@GarisonLotus as you correctly pointed out in our in-person conversation, it would be good to add a test here to ensure regressions on this can't pop up on the future

phobologic:
Looks good - basically only request is what @troyready said (and it sounds like you pointed out in person, so awesome) - be good to test this new logic.

Thanks for this, and for tracking down where the actual error was!

GarisonLotus:
@phobologic , @troyready - Tests added and passing

GarisonLotus:
Just an FYI on this, I tested removing my change locally in base.py, and it did indeed fail with my test:
```
======================================================================
ERROR: test_variable_exists_but_value_is_none (stacker.tests.blueprints.test_base.TestVariables)
----------------------------------------------------------------------
  File "/Users/garison/Development/GarisonLotus/stacker/stacker/tests/blueprints/test_base.py", line 547, in test_variable_exists_but_value_is_none
    blueprint_name)
  File "/Users/garison/Development/GarisonLotus/stacker/stacker/blueprints/base.py", line 214, in resolve_variable
    raise ValidatorError(var_name, validator.__name__, value, exc)
UnboundLocalError: local variable 'value' referenced before assignment
```


This is expected.  The test does work. :-)

phobologic:
Thanks @GarisonLotus !

PR #389, opened 2017-05-17
Added exponential backoff waiting for AWS changeset to stabilize

smverleye:
We appear to be the only stacker users experiencing long wait periods on CFN changeset generation, but we've seen some stacks esp. our VPN take up to a minute to stabilize. This backoff is meant to function very similarly to the `retry_on_throttling`

### DIFF
--- a/stacker/providers/aws/interactive.py
+++ b/stacker/providers/aws/interactive.py
@@ -143,6 +143,11 @@ def wait_till_change_set_complete(cfn_client, change_set_id, try_count=25,
         complete = response["Status"] in ("FAILED", "CREATE_COMPLETE")
         if complete:
             break
+        if sleep_time == max_sleep:
+            logger.debug(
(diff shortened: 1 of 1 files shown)


phobologic:
This looks good, thanks - one thing I think it needs is more logging, especially if it's going to hang for up to a minute.  Maybe post a `logger.debug` message saying we're waiting on the change set still every 10 retries?

smverleye:
Yep that seems reasonable. I switched it to logging every time it's waiting the maximum amount; that'll give frequent feedback once it's looping longer than should be expected. What do you think?

troyready:
Just hit this for the first time today. By the time I went into the console to look at the error it had stabilized, so I suspect this would be helpful for us too.

phobologic:
Man, sorry - I've been swamped and meant to merge this.  It's in now.  Thanks for the ping @troyready, and thanks @smverleye for the PR!

PR #397, opened 2017-06-09
add stacker_bucket_endpoint config option

troyready:
Allows for use with S3 buckets in isolated regions like cn-north-1

Example endpoint retrieved from
http://docs.amazonaws.cn/en_us/general/latest/gr/rande.html#cnnorth_region

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -42,9 +42,9 @@ in the config.
 
 S3 Bucket Endpoint
 ------------------
-Isolated AWS regions don't permit access to S3 buckets via the default
-'s3.amazonaws.com' endpoint. This endpoint can be overriden via the
-**stacker_bucket_endpoint** top level keyword in the config, e.g.::
+The correct S3 endpoint should be automatically chosen, but it can also be
+overriden via the **stacker_bucket_endpoint** top level keyword in the config,
+e.g.::
 
   stacker_bucket_endpoint: s3.cn-north-1.amazonaws.com.cn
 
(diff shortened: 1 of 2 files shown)


phobologic:
@troyready is there no way to determine this in a more automated fashion?  Like can we automatically choose the right end point based on the region, or do you think it's better to make this more manual?

troyready:
@phobologic good thought. I've added the automatic detection to the pull request.

I tend to over-value config knobs, so I've left the ability to manually override the endpoint as well. I wouldn't be opposed to dropping it though.

phobologic:
So I was wondering how boto handles this, and I think this can be totally automatic with botocore/boto3:

```
>>> from botocore import session
>>> s = session.get_session()
>>> c = s.create_client("s3", "us-west-1")
>>> c._endpoint.host
u'https://s3-us-west-1.amazonaws.com'
>>> c = s.create_client("s3", "us-east-1")
>>> c._endpoint.host
u'https://s3.amazonaws.com'
>>> c = s.create_client("ec2", "cn-north-1")
>>> c._endpoint.host
u'https://ec2.cn-north-1.amazonaws.com.cn'
```

Since the `provider` should have the s3 connection on it, you should be able to do something similar to above with it I believe.  What do you think?

troyready:
@phobologic yeah, that seems much better. Based on my quick reading, it looks like you could just close this and add the region to https://github.com/remind101/stacker/blob/1.0.2/stacker/actions/base.py#L66 ?

troyready:
Closing in favor of #413

PR #401, opened 2017-06-15
add a hint when forbidden bucket access

ralph-tice:
In case you make a newb mistake like me and try to use a namespace called "dev" or something similar, a bucket named "stacker-dev" or similar is prolly gonna be taken.


phobologic:
Thanks!

PR #404, opened 2017-06-16
Get region from provider to the environment for interpolation

smverleye:
When working with stacker in multiple regions I've found it essential to be able to use variable interpolation for the region. For example, the stacker template bucket needs to be region specific because CloudFormation does not allow manipulating objects in other regions.

This change is meant to read the region back from the provider into the context, if it's set in the provider. This way we get a variable ${region} available to configuration templates, and it is set correctly whether passed in or using the default boto credentials chain.

### DIFF
--- a/stacker/providers/aws/default.py
+++ b/stacker/providers/aws/default.py
@@ -94,6 +94,9 @@ class Provider(BaseProvider):
         # Necessary to deal w/ multiprocessing issues w/ sharing ssl conns
         # see: https://github.com/remind101/stacker/issues/196
         self._pid = os.getpid()
+        # If region is not set explicitly, fetch it from the boto session
+        if not self.region:
+            self.region = get_session().region_name
 
     @property
     def cloudformation(self):
--- a/stacker/commands/stacker/__init__.py
+++ b/stacker/commands/stacker/__init__.py
@@ -31,6 +31,11 @@ class Stacker(BaseCommand):
         else:
             logger.info('Using Default AWS Provider')
             options.provider = default.Provider(region=options.region)
+
+        # Copy the region from the provider, if set, to the environment
+        if options.provider.region:
+            options.environment.update({'region': options.provider.region})
+
         options.context = Context(
             environment=options.environment,
             logger_type=self.logger_type,
--- a/stacker/context.py
+++ b/stacker/context.py
@@ -30,7 +30,6 @@ class Context(object):
     the command line and specified in the config to `Stack` objects.
 
     Args:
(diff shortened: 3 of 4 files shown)


phobologic:
Hey @smverleye - this looks good, can you write some tests? The get_session change - not sure what None for the region actually does, so just a little nervous there.

phobologic:
Oh, and I think someone in the channel actually had this suggestion - if there's already a `region` in the environment, maybe we should avoid clobbering it.  Sound good?

stacker/providers/aws/default.py
```diff
@@ -94,6 +94,9 @@ def __init__(self, region, **kwargs):
         # Necessary to deal w/ multiprocessing issues w/ sharing ssl conns
         # see: https://github.com/remind101/stacker/issues/196
         self._pid = os.getpid()
+        # If region is not set explicitly, fetch it from the boto session
+        if not self.region:
```
phobologic: I think you could combine this with the region line by using:

```
self.region = region or get_session().region_name
```

That said, after digging into this, I'm not really sure how get_session without any arguments is going to return an actual region_name?

smverleye: I think typical usage of get_session() was without any region being passed in; the only exception is if a user passes the `--region` argument. Otherwise, the region is set by the credential chain; that is, it uses boto default session logic to resolve the region, which amounts to AWS_REGION env variable, or the contents of the profile (`~/.aws/credentials | config`).

I did see the comment about "not clobbering existing region" but that's not a good idea. The only way someone would already have region in the environment is if they passed it in. For example, someone could: `stacker build -r us-east-1 -e region=us-west-2`. This will establish a boto session to `us-east-1` but use `us-west-2` in the template - Stacker shouldn't allow that to happen.

I'll look at adding a test case specific to this change.

phobologic: You make a good point about not allowing changing of the region - but I think we should be careful of squashing any potential settings people have/rely on.  It seems unlikely that people have a `region` variable, and that it'd be different than what they'd want when using `--region`, etc, but it might.

How about we throw a helpful exception if someone has `region` set manually?

Also, can you add some documentation that `region` is a reserved environment variable, and how it's set?

phobologic:
Hey @smverleye - havne't heard back from you on this, and it looks like we have some code coverage changes that might need to be looked at.  I'm planning to release a new verison of stacker today, so this likely won't make it into that release.  Let me know when you'll have a chance to look at this.

PR #405, opened 2017-06-20
Fix issues with "none" as variable values

phobologic:
This was introduced by the merging of two different PRs (#379, #388),
each with their own ideas about "None" as a value.  This should give
us the best of both worlds (hopefully).


phobologic:
@danielkza @GarisonLotus I'd appreciate it if you could both take a look at this and :+1: if it still works for you.

GarisonLotus:
I know I was late to the party here - but I tested it and it works. :-)

Issue #384, opened 2017-05-12
Cannot destroy individual stack with requires


sdechgan:
I cannot destroy an individual stack that is required by another stack

For example:

Here is my stacker.yml
```
stacks:
  - name: stack1
    class_path: blueprints.application.Application
    enabled: true
  - name: stack2
    class_path: blueprints.s3.Event
    enabled: true
    requires:
      - stack1
  - name: stack3
    class_path: blueprints.api.OpenApi
    enabled: true
    requires:
      - stack1
      - stack2
```

Here is my command
```
stacker destroy --stacks stack3  environments/dev.env stack.yml
```

I get the following error:
```
[2017-05-12T14:53:29] Using Default AWS Provider
Traceback (most recent call last):
  File "~/.pyenv/bin/stacker", line 9, in <module>
    args.run(args)
  File "~/.pyenv/lib/python2.7/site-packages/stacker/commands/stacker/destroy.py", line 35, in run
    action.execute(force=options.force, tail=options.tail)
  File "~/.pyenv/lib/python2.7/site-packages/stacker/actions/base.py", line 124, in execute
    self.run(*args, **kwargs)
  File "~/.pyenv/lib/python2.7/site-packages/stacker/actions/destroy.py", line 100, in run
    plan = self._generate_plan(tail=tail)
  File "~/.pyenv/lib/python2.7/site-packages/stacker/actions/destroy.py", line 56, in _generate_plan
    stacks_dict[stack_name],
KeyError: 'dev-stack2'
```

Isn't that the correct command to delete an individual stack?

phobologic:
I think we could definitely have a better error there, but I think the logic is pretty sound.  If stack3 requires stack2, then you shouldn't be able to destroy stack2 without destroying stack3 as well.

sdechgan:
Shouldn't you be able to delete `stack3` without deleting `stack2`?  Logically, `stack2` doesn't need `stack3` to work properly because `stack3` was the last stack created.

That or I don't understanding is usage of the `--stacks` parameter.

ejholmes:
This looks like a legit bug when specifying `--stacks` with `destroy`, no? It errors out [here](https://github.com/remind101/stacker/blob/master/stacker/actions/destroy.py#L56) when trying to generate the plan.

`stack3` should only depend on `stack2` during a build. During a `destroy` the graph should be transposed, so that `stack2` depends on `stack3`. It's _likely_ that this is already fixed in the DAG branch.

[...]

Issue #300, opened 2017-01-20
handle dependency trees within stacker's purview


ajk8:
**Problem:**
If you have a stacker configuration defining several stacks which pass variables between them at the configuration level, you cannot specify a stack in the middle using the `--stacks` option. Instead, you must explicitly specify all of the upstream stacks using separate `--stacks` arguments. 

**Desired Outcome:**
I should be able to specify a single stack, and stacker should be able to build the dependency tree and submit everything upstream for building.

**Example:**

* I have 5 stacks, A, B, C, D, and E
* All stacks depend upon A
* Stack D depends upon B and C
* Stack E depends upon E

In order to build this currently, I have to specify:

```
stacker build <path/to/file.env> <path/to/file.yml> --stacks A --stacks B --stacks C --stacks D
```

A preferred invocation would be:

```
stacker build <path/to/file.env> <path/to/file.yml> --stacks D
```

These two things should be functionally equivalent.

**Addendum:**
I could see a reasonable argument that this could lead to confusing scenarios when folks are using direct cross-stack references inside of their troposphere code. These would then fail to be satisfied at runtime. One way to solve for this would be to introduce an additional flag to the effect of `--detect-stack-dependencies` which requires the user to knowingly automagically handle dependencies.

phobologic:
Thanks for the bug, @ajk8 - fortunately we already have a PR to do just this: #282 - it's delayed a bit because the author is on leave, but hopefully it should be in soon!

ejholmes:
This will be fixed in the next release. The `—stacks` flag now builds any required dependencies.

PR #406, opened 2017-06-21
DAG

ejholmes:
This PR is an internal refactor to make stacker build a Directed Acyclic Graph when planning execution.

Fixes https://github.com/remind101/stacker/issues/300
Fixes https://github.com/remind101/stacker/issues/186
Fixes https://github.com/remind101/stacker/issues/279
Fixes https://github.com/remind101/stacker/issues/384

### Performance

The DAG simplifies [parallel scheduling](https://ocw.mit.edu/courses/electrical-engineering-and-computer-science/6-042j-mathematics-for-computer-science-spring-2015/readings/MIT6_042JS15_Session17.pdf), so that we can execute mutually exclusive stacks at the same time.

In this PR, each stack is executed in it's own thread. The thread blocks until all of the stacks dependencies have been executed, and then immediately starts. This means that stacks at the same depth, won't block each other unnecessarily.

The worst case `stacker build` time, is now the total time it takes for the longest critical chain (the longest chain of dependencies). Previously, the worst case `stacker build` time was a combination of the longest critical chain, in addition to the widest point in the graph.

For a concrete example, using the `benchmark.yaml` config, this PR dropped NOOP execution time from ~1min+ to ~15 seconds.

In our internal infrastructure repo, it drops NOOP execution time from ~7 minutes down to ~2.5 minutes (with my shitty RTT in Thailand).

### Automatic dependency resolution for `--stacks`

`stacker build` supports a `--stacks` flag, however, in practice, it's difficult to use, because it won't automatically resolve dependencies. The DAG makes this easy, so when you provide a `--stacks` flag now, any dependencies are automatically visited.

### Simplified destroy

`stacker destroy` is just a `stacker build` with a transposed graph.

### Graceful SIGINT/SIGTERM

SIGINT and SIGTERM are handled more gracefully now. When received, stacker will wait until all _current_ stack operations complete before shutting down.

### Cyclic Dependency Checks

If there are any circular dependencies, it's detected and an error is shown. Previously, stacker would just hang forever.

### Duplicate Node Detection

Previously, if you accidentally added a node with the same name, you wouldn't know about it, until it overwrote your existing stack with the new template (I've done this before with an RDS stack...). With the DAG, duplicate nodes are automatically detected.

### Logging Improvements

The log output is updated whenever a step changes status, which fixes a number of issues where it would previously look like stacker wasn't doing anything (e.g. a noop run would never update the statuses).

Currently, logging is pretty jiterry because of this. I'll try to fix these issues before this gets merged.

https://asciinema.org/a/ilOwudPADDXQBnmfqa57JaC4Y

### Thread Safety

stacker is now multi-threaded, so we need to be careful to ensure that everything is thread safe. Not sure if there are any static analysis tools in the Python world for this.

### Rate limiting

Because we're executing more API calls in parallel, it's possible for stacker to hit rate limits more easily, although, I've yet to hit any rate limits in all of my tests.

---

### Tested

* [x] Interactive mode
* [ ] Tailing
* [x] Assuming role with MFA
* [x] Diff

### TODO

* [ ] Add a `--j/jobs` flag for specifying concurrency
* [x] Figure out how to show stack traces without getting overwritten by checkpointing
* [ ] Fix jitter in checkpointing
* [ ] Fix SIGINT catch breaking CTL-C from MFA prompt.
* [ ] Better handling of retrieving stack status updates to avoid throttling on DescribeStacks.

### DIFF
--- a/stacker/actions/base.py
+++ b/stacker/actions/base.py
@@ -7,6 +7,9 @@ from ..plan import Step
 from ..status import (
     SUBMITTED,
     COMPLETE,
+    ERRORED,
+    CANCELLED,
+    INTERRUPTED,
 )
 
 import botocore.exceptions
@@ -63,6 +66,8 @@ def _check_point(plan):
     status_to_color = {
         SUBMITTED.code: Fore.YELLOW,
         COMPLETE.code: Fore.GREEN,
+        CANCELLED.code: Fore.MAGENTA,
+        ERRORED.code: Fore.RED,
     }
     logger.info("Plan Status:", extra={"reset": True, "loop": plan.id})
 
@@ -147,14 +152,21 @@ class BaseAction(object):
         self._conn = None
         self.cancel = cancel or threading.Event()
 
-    def _action(self):
+    def _action(self, *args, **kwargs):
         raise NotImplementedError
+
+    def _run_action(self, *args, **kwargs):
+        # Cancel execution if flag is set.
+        if self.cancel.wait(0):
+            return INTERRUPTED
+
+        return self._action(*args, **kwargs)
 
     @property
     def steps(self):
         if not hasattr(self, "_steps"):
             self._steps = [
-                Step(stack, fn=self._action)
+                Step(stack, fn=self._run_action)
                 for stack in self.context.get_stacks()]
         return self._steps
 
--- a/stacker/actions/base.py
+++ b/stacker/actions/base.py
@@ -1,10 +1,98 @@
-import copy
+import threading
 import logging
+
+from colorama.ansi import Fore
+
+from ..plan import Step
+from ..status import (
+    SUBMITTED,
+    COMPLETE,
+)
 
 import botocore.exceptions
 from stacker.session_cache import get_session
 
 logger = logging.getLogger(__name__)
+
+
+def outline_plan(plan, level=logging.INFO, message=""):
+    """Print an outline of the actions the plan is going to take.
+    The outline will represent the rough ordering of the steps that will be
+    taken.
+    Args:
+        level (int, optional): a valid log level that should be used to log
+            the outline
+        message (str, optional): a message that will be logged to
+            the user after the outline has been logged.
+    """
+    steps = 1
+    logger.log(level, "Plan \"%s\":", plan.description)
+
+    nodes = plan.dag.topological_sort()
+    nodes.reverse()
+    for step_name in nodes:
+        step = plan.steps[step_name]
+        logger.log(
+            level,
+            "  - step: %s: target: \"%s\", action: \"%s\"",
+            steps,
+            step_name,
+            step.fn.__name__,
+        )
+        steps += 1
+
+    if message:
+        logger.log(level, message)
+
+
+def check_point_fn():
+    """Adds a check_point function to each of the given steps."""
+
+    lock = threading.Lock()
+
+    def _fn(plan):
+        lock.acquire()
+        _check_point(plan)
+        lock.release()
+
+    return _fn
+
+
+def _check_point(plan):
+    """Outputs the current status of all steps in the plan."""
+    status_to_color = {
+        SUBMITTED.code: Fore.YELLOW,
+        COMPLETE.code: Fore.GREEN,
+    }
+    logger.info("Plan Status:", extra={"reset": True, "loop": plan.id})
+
+    longest = 0
+    messages = []
+
+    nodes = plan.dag.topological_sort()
+    nodes.reverse()
+    for step_name in nodes:
+        step = plan.steps[step_name]
+
+        length = len(step.name)
+        if length > longest:
+            longest = length
+
+        msg = "%s: %s" % (step.name, step.status.name)
+        if step.status.reason:
+            msg += " (%s)" % (step.status.reason)
+
+        messages.append((msg, step))
+
+    for msg, step in messages:
+        parts = msg.split(' ', 1)
+        fmt = "\t{0: <%d}{1}" % (longest + 2,)
+        color = status_to_color.get(step.status.code, Fore.WHITE)
+        logger.info(fmt.format(*parts), extra={
+            'loop': plan.id,
+            'color': color,
+            'last_updated': step.last_updated,
+        })
 
 
 def stack_template_key_name(blueprint):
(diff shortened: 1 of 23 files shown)


stacker/actions/base.py
```diff
@@ -1,12 +1,100 @@
...
+        steps += 1
+
+    if message:
+        logger.log(level, message)
+
+
+def check_point_fn():
+    """Adds a check_point function to each of the given steps."""
+
+    lock = threading.Lock()
```
ejholmes: I think the jitter comes from the lock here. Checkpointing should be throttled, not locked, so that multiple status changes in short succession don't spam the terminal.

stacker/dag/__init__.py
```diff
@@ -0,0 +1,313 @@
...
+        transposed = DAG()
+        for node, edges in graph.items():
+            transposed.add_node(node)
+        for node, edges in graph.items():
+            # for each edge A -> B, transpose it so that B -> A
+            for edge in edges:
+                transposed.add_edge(edge, node)
+        return transposed
+
+    def walk(self, walk_func, graph=None):
```
ejholmes: This probably deserves some special attention when reviewed. It's the core algorithm for implementing a parallel walk using threads that block on each other.

stacker/dag/__init__.py
```diff
@@ -0,0 +1,313 @@
...
+    from collections import OrderedDict
+except:
+    from ordereddict import OrderedDict
+
+
+class DAGValidationError(Exception):
+    pass
+
+
+class DAG(object):
```
ejholmes: FYI, this implementation is pulled from https://github.com/thieman/py-dag with a few changes (adds `walk` and `transpose` methods). Since it's small, I just pulled it in directly so we don't need to add an extra dependency.

stacker/commands/stacker/base.py
```diff
@@ -101,6 +117,11 @@ def run(self, options, **kwargs):
     def configure(self, options, **kwargs):
         self.logger_type = setup_logging(options.verbose, options.interactive)
 
+    def semaphore(self, options):
+        if options.interactive:
+            return threading.Semaphore(1)
```
ejholmes: Currently, dropping down to interactive mode removes parallelism, which isn't actually necessary. It would be better to add locks around the change set prompts.

stacker/dag/__init__.py
russellballestrini: Even though we plan to pull this in, do you think a PR to the origin would be worth while with the additional `walk` and `transpose` methods?

ejholmes: I think `transpose` is generic enough to contribute back. `walk` is somewhat specific, being threaded and all (in reality, I should probably pull it off of `DAG` entirely and just make it a function that takes a `DAG` as an argument).

ejholmes:
One annoying thing about going to threading, is that if a thread throws an exception, the stack trace can get overwritten by checkpointing:

https://asciinema.org/a/EUd6jUK8ogTqjsm0CgjMgtXDB

stacker/actions/build.py
```diff
@@ -229,31 +234,20 @@ def _launch_stack(self, stack, **kwargs):
...
             except StackDidNotChange:
                 return DidNotChangeStatus()
+            except CancelExecution as e:
+                return CancelledStatus(reason=e.message)
 
         return new_status
 
-    def _generate_plan(self, tail=False):
-        plan_kwargs = {}
-        if tail:
```
ejholmes: I'm not really sure what to do about the `--tail` flag. From playing with it on `master`, it doesn't seem to really work very well (it constantly gets overwritten by checkpoint'ing, see https://github.com/remind101/stacker/issues/386).

Since the DAG will likely go into 1.1, maybe we should pull `--tail` out? 

If we want to keep it in, we can make it work, but we should probably fix it in 1.0.x first.

danielkza:
Apologies if it is too late given the seemly advanced stage of this PR, but were any other approaches considered other than explicitly building a DAG and walking it in multiple threads?

I've recently had a similar use case, and chose to implement it using Python 3's `asyncio` with "implicit" dependencies defined by waiting on futures corresponding to parents ' builds instead of explicitly walking the graph (which has the secondary advantage of making data propagation to children easier). Concurrency is limited by waiting on a semaphore before the first stage of an actual build (but only after waiting for dependencies).

While `asyncio` is not an option with Python 2, and probably a bit too large of a change for stacker as a whole, a very similar approach can be implemented with threads and events.

The interesting part of that implementation can be seen [here](https://github.com/Cobliteam/shelver/blob/master/shelver/build/coordinator.py) (apologies for the less-than-prod-quality code).

ejholmes:
@danielkza thanks for the feedback!

I think there's probably two things here that are orthogonal.

1. Building a graph before execution.
2. Whether or not to use threads for parallelization. The current implementation here uses threads to walk the graph, using an algorithm very similar to what you described above, but could be implemented using asyncio I think (if we ditched python2 support).

There's a couple of advantages that I see of building an explicit graph vs using implicitly waiting on outputs:

1. When we build a graph, it makes introspection of dependencies really easy. We can output a `DOT` formatted graph so you can see how all the stacks relate to each other.
2. It makes the `destroy` action really simple, since we just need to take the graph and then transpose it.
3. We can detect cyclic dependencies.

So, with that in place, then the question comes down to whether or not to use threading to walk the graph.

I'm definitely not married to the idea of using threads for parallelization here, since it's pretty resource inefficient (we're just waiting on IO to complete). I'm not sure what other options there are for python2 that would give us the same level of performance though.

danielkza:
@ejholmes Yeah, I figure the explicit DAC has some clear advantages. But I still think using futures for "flow control" would be preferable to checkpointing and explicit state monitoring. They make waiting relationships much more clear, likely easier to debug an test, and can more easily work with a fixed pool of resources instead of creating an arbitrary number of threads.

I will play with the idea a bit during this week and see if it actually makes sense. At worse I'll learn why it *doesn't* work and maybe get some insights.

phobologic:
Looks like someone ported asyncio to 2.7 - though YMMV: https://pypi.python.org/pypi/trollius

ejholmes:
Closing this in favor of https://github.com/remind101/stacker/pull/523.

Issue #277, opened 2016-11-29
Configure region outside of the CLI?


rouge8:
It would be great if we could specify the region when configuring stacks, e.g.:

```yaml
stacks:
  - name: app-us-west
    class_path: my_blueprints.App
    region: us-west
  - name: app-us-east
    class_path: my_blueprints.App
    region: us-east
```

phobologic:
I wonder if this would be a good idea to combine w/ #263.  Maybe make a profile a mix of account id + region?

phobologic:
@ejholmes ping ^^

ejholmes:
Yep, I think it'd make sense to just add support for specifying what profile to use when creating a stack. That'd easily allow for cross account/region/etc use cases, as well as all the other stuff that profiles allow, like assuming rules + mfa.

For example, in the case above, you can just point stacker at an `~/.aws/config` file like:

```
[profile "infra-east"]
region = us-east-1

[profile "infra-west"]
region = us-west-1
```

And then specify the profile in the stacks definition:

```yml
stacks:
  - name: app-us-west
    class_path: my_blueprints.App
    profile: infra-west
  - name: app-us-east
    class_path: my_blueprints.App
    profile: infra-east
```

phobologic:
(pinging @rouge8 since he opened this, and @troyready since he mentioned he was looking at this in #318):

I'm going to mark this as post-1.0.  I'm working on trying to get 1.0 out, and I think this can wait till after that (though I don't think it'll take long).  Let me know if you disagree!

rouge8:
Fine with me, I ended up going with a wrapper script around the Stacker CLI anyways. :)

phobologic:
@rouge8 Cool - hopefully we can pull this in soon so you can do away with the wrapper :)

kappataumu:
@phobologic This would really simplify some stuff, like creating buckets for cross-region replication. Is it on the roadmap?

phobologic:
@kappataumu It's in theory on the roadmap, but not currently scheduled in any way.  It's definitely something we want to do though - or, if someone wants to give us a PR, we'd be happy to take a look!

ejholmes:
This is something I want to tackle in the next ~month or so. I put together a little RFC for how I think this should work if anyone else has thoughts: https://github.com/remind101/stacker/wiki/RFC:-Profiles

troyready:
[...]

Issue #263, opened 2016-11-08
Feature: Cross account management


ejholmes:
Just jotting this down here to make note of it. It'd be pretty cool if stacker could link cross account stacks. For example, if most of your resources are managed in account A, you may want to add an account B that's 100% isolated from account A, then give account A access to AssumeRole into B. Being able to manage resources via stacker in both A and B would be pretty rad.

I'm not sure the best way to do it, but it could integrate with AWS profiles, so you could tag certain stacks with a certain profile. e.g.

```yaml
stacks:
  - name: vpc
     profile: A
  - name: emr
     profile: B
  - name: app
     profile: A
     variables:
       EMRAssumeRole: ${emr::Role}
```

ttaub:
This would actually be pretty easy to do using the new way we do sessions, you could just have an optional profile field when creating the session object.

E.g
```
session = get_session(region_name='us-east-1', profile='dev')
```

I'll submit a PR for a demo of this functionality when I get the chance.

ejholmes:
Put together a little RFC for how I think this should work: https://github.com/remind101/stacker/wiki/RFC:-Profiles

PR #407, opened 2017-06-21
Support for profiles (cross region/account support)

ejholmes:
This is based on the RFC in https://github.com/remind101/stacker/wiki/RFC:-Profiles

Closes https://github.com/remind101/stacker/issues/263
Fixes https://github.com/remind101/stacker/issues/277

With this change, you can specify a boto3 profile to use for a given stack. This can be used for cross account/region provisioning and linking of stacks. For example, say you wanted to provision a hot and cold version of an application in multiple regions, you could do something like this:

```yaml
stacks:
- name: vpc-hot
  profile: hot
- name: vpc-cold
  profile: cold
- name: app
  profile: hot
  variables:
    VpcId: ${vpc-hot::VpcId}
- name: app-cold
  profile: cold
  variables:
    VpcId: ${vpc-cold::VpcId}
```

```console
$ echo <<EOF > .aws/config.prod
[profile hot]
region = us-east-1

[profile prod/cold]
region = us-west-1
EOF
```

```console
$ AWS_CONFIG_FILE=.aws/config.prod stacker build prod.env stacker.yaml
```

## TODO

* [ ] Docs around using boto3 profiles with stacker.
* [x] Hooks. Add support for specifying profile?

### DIFF
--- a/stacker/actions/base.py
+++ b/stacker/actions/base.py
@@ -2,7 +2,6 @@ import copy
 import logging
 
 import botocore.exceptions
-from stacker.session_cache import get_session
 
 logger = logging.getLogger(__name__)
 
@@ -62,8 +61,7 @@ class BaseAction(object):
     def s3_conn(self):
         """The boto s3 connection object used for communication with S3."""
         if not hasattr(self, "_s3_conn"):
-            session = get_session(self.provider.region)
-            self._s3_conn = session.client('s3')
+            self._s3_conn = self.provider.session.client('s3')
 
         return self._s3_conn
 
--- a/stacker/actions/build.py
+++ b/stacker/actions/build.py
@@ -186,8 +186,12 @@ class Action(BaseAction):
         if not should_submit(stack):
             return NotSubmittedStatus()
 
+        provider = self.provider
+        if stack.profile:
+            provider = self.provider.with_profile(stack.profile)
+
         try:
-            provider_stack = self.provider.get_stack(stack.fqn)
+            provider_stack = provider.get_stack(stack.fqn)
         except StackDoesNotExist:
             provider_stack = None
 
@@ -196,17 +200,17 @@ class Action(BaseAction):
             logger.debug(
                 "Stack %s provider status: %s",
                 stack.fqn,
-                self.provider.get_stack_status(provider_stack),
+                provider.get_stack_status(provider_stack),
             )
-            if self.provider.is_stack_completed(provider_stack):
+            if provider.is_stack_completed(provider_stack):
                 submit_reason = getattr(old_status, "reason", None)
                 return CompleteStatus(submit_reason)
-            elif self.provider.is_stack_in_progress(provider_stack):
+            elif provider.is_stack_in_progress(provider_stack):
                 logger.debug("Stack %s in progress.", stack.fqn)
                 return old_status
 
         logger.debug("Resolving stack %s", stack.fqn)
-        stack.resolve(self.context, self.provider)
+        stack.resolve(self.context, provider)
 
         logger.debug("Launching stack %s now.", stack.fqn)
         template_url = self.s3_stack_push(stack.blueprint)
@@ -217,15 +221,15 @@ class Action(BaseAction):
         if not provider_stack:
             new_status = SubmittedStatus("creating new stack")
             logger.debug("Creating new stack: %s", stack.fqn)
-            self.provider.create_stack(stack.fqn, template_url, parameters,
-                                       tags)
+            provider.create_stack(stack.fqn, template_url, parameters,
+                                  tags)
         else:
             if not should_update(stack):
                 return NotUpdatedStatus()
             try:
                 new_status = SubmittedStatus("updating existing stack")
-                self.provider.update_stack(stack.fqn, template_url, parameters,
-                                           tags)
+                provider.update_stack(stack.fqn, template_url, parameters,
+                                      tags)
                 logger.debug("Updating existing stack: %s", stack.fqn)
             except StackDidNotChange:
                 return DidNotChangeStatus()
--- a/stacker/actions/destroy.py
+++ b/stacker/actions/destroy.py
@@ -60,8 +60,12 @@ class Action(BaseAction):
         return plan
 
     def _destroy_stack(self, stack, **kwargs):
+        provider = self.provider
+        if stack.profile:
+            provider = self.provider.with_profile(stack.profile)
+
         try:
-            provider_stack = self.provider.get_stack(stack.fqn)
+            provider_stack = provider.get_stack(stack.fqn)
         except StackDoesNotExist:
             logger.debug("Stack %s does not exist.", stack.fqn)
             # Once the stack has been destroyed, it doesn't exist. If the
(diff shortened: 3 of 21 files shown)


ralph-tice:
The RFC uses dashes for delimiters in profile names instead of slashes, which I think is more consistent and uniform -- slash implies a hierarchy to me rather than joining like dash and underscore indicate.

Also the `profile ` prefix specified in the RFC isn't necessary so should be excluded from the example IMO

ejholmes:
> The RFC uses dashes for delimiters in profile names instead of slashes, which I think is more consistent and uniform -- slash implies a hierarchy to me rather than joining like dash and underscore indicate.

Sorry, I guess the RFC is a little ambiguous about this. The RFC isn't really trying to force a convention around how people name their profiles. You can name them however you want, then reference them with `profile: <profile name>`.

> Also the profile prefix specified in the RFC isn't necessary so should be excluded from the example IMO.

This is true, although, if you have a stacker project shared amongst team members, you'll probably want to also share the config file with the named profiles (if you decide to use profiles). But again, this PR doesn't make any assumptions about that, people can use boto3 profiles however they want, based on http://boto3.readthedocs.io/en/latest/guide/configuration.html.

I'll make a note to update the docs on how to use boto3 profiles with stacker, and maybe some recommendations on sharing a common profile config.

smverleye:
> This is true, although, if you have a stacker project shared amongst team members, you'll probably want to also share the config file with the named profiles (if you decide to use profiles).

That just meant you can leave off the actual `profile` from the profile definition, and it's now recommended to do so:
```
[prod-cold]
region: us-west-2
```

FWIW I could definitely get behind this approach, as it would be compatible with our usage. Currently we're doing multiple regions, but with separate stacker executions for each. This would allow, as you say, for usage of xref across regions which would be valuable for many situations.

troyready:
@ejholmes love it!

Is there a way to extend this to minimize the need for setting the `AWS_CONFIG_FILE` environment variable? Maybe another config option to specify its location?

As is this would work well for us, but it'd be nice to minimize the amount of `botocore.exceptions.ProfileNotFound` exceptions that users encounter when checking out a stacker repo.

ralph-tice:
Sorry, the ~/.aws/config file requires the `[profile ` prefix but ~/.aws/credentials doesn't.  I've never really used the ~/.aws/config file.

I don't see how you would share these profiles without also getting into secrets management...

ejholmes:
> Is there a way to extend this to minimize the need for setting the AWS_CONFIG_FILE environment variable? Maybe another config option to specify its location?

@troyready I thought about it, but not sure it should really be the responsibility of stacker. I think it makes more sense to just use the boto3 defaults, and compose stacker with scripts. But I can be persuaded otherwise.

For example, in our stacker repos, there are just `make prod/stage/dev` targets that you can use immediately after checkout that sets everything up correctly.

In our setup, we'd end up doing this:

```make
stage prod dev: %:
	env AWS_CONFIG_FILE=.aws/config stacker build conf/$@.env stacker.yaml
```

> I've never really used the ~/.aws/config file.
> I don't see how you would share these profiles without also getting into secrets management...

@ralph-tice highly suggest reading up on http://boto3.readthedocs.io/en/latest/guide/configuration.html, and [assumable roles](http://docs.aws.amazon.com/cli/latest/userguide/cli-roles.html). They're pretty great :).

There's absolutely no need to share credentials, and we're not suggesting that people do this.

troyready:
@ejholmes appreciate the thoughts. I'd really like to avoid the need for a Makefile-type solution, as it doesn't really feel like it solves #277 for me, but I can appreciate the desire for simplicity.

ejholmes:
@troyready just for my own curiosity, why the aversion to wrapping `stacker build` with a script? Presumably, you're already specifying some flags/arguments on the command line (e.g. `.env` file and `stacker.yaml`), so not sure how `AWS_CONFIG_FILE` ends up being any different.

Any suggestions on where you would like this to be configured (e.g. `stacker.yaml` or `.env` file)?

troyready:
@ejholmes it's definitely just a bad "feeling" to me, where all practical uses of stacker need to be wrapped in a script. I like the idea of the config.yml being the language of what to do and where to get information from it.

As you've said, the current state of things don't allow that so what you're providing is all upside. I'm just trying to continuously nudge improvements to ensure feature superiority against Terraform.

ejholmes:
Added a commit to this that adds profile support to hooks, so you can do stuff like this:

```yaml
pre_build:
    required: true
    profile: cold
    args:
      domain: ${basedomain}
    required: true
    profile: hot
    args:
      domain: ${basedomain}
```

smverleye:
> As you've said, the current state of things don't allow that so what you're providing is all upside. I'm just trying to continuously nudge improvements to ensure feature superiority against Terraform.

So long as you make no attempt to maintain state outside of the provider itself, Stacker will always be superior to Terraform.

> Sorry, the ~/.aws/config file requires the [profile prefix but ~/.aws/credentials doesn't. I've never really used the ~/.aws/config file.

Was not aware of the difference, but personally I just recommend using only the credentials file. Boto has support for the `~/.aws/config` file but at least the Java SDK does not. From https://boto3.readthedocs.io/en/latest/guide/configuration.html

> If you want to interoperate with multiple AWS SDKs (e.g Java, Javascript, Ruby, PHP, .NET, AWS CLI, Go, C++), use the shared credentials file (~/.aws/credentials). By using the shared credentials file, you can use a single file for credentials that will work in all the AWS SDKs.

> I don't see how you would share these profiles without also getting into secrets management...

Indeed, assumed roles is great. We have shared profiles for different accounts and different access levels:

```
[dev]
region         = us-east-1
output         = json
role_arn       = arn:aws:iam::111111111111:role/NormalAccess
source_profile = identity

[dev-admin]
region         = us-east-1
output         = json
role_arn       = arn:aws:iam::111111111111:role/Admin
source_profile = identity

[prod]
region         = us-east-1
output         = json
role_arn       = arn:aws:iam::22222222222:role/Admin
source_profile = identity
```

Then you just have users populate the "identity" profile with actual IAM access keys. Or for bonus security points have a tool that uses Federation/SAML to fetch temporary STS tokens. Regardless, you can then use shared profiles to define appropriate permission level, account and potentially region without publishing any creds in the file.

This is why being able to specify a profile associated with each stack could be nice. If it requires access to IAM or other sensitive resources you could specify that directly on the stack metadata, by configuring it as requiring an elevated privilege role.

bwhaley:
This feature looks like exactly what I need for xref across multiple AWS accounts using assumerole. Looks like it has been stalled out for a bit and I think @ejholmes mention in Slack he's just not gotten back to it. Can I do something to help move it forward? Does it need additional test coverage?

ejholmes:
@bwhaley I think additional functional test coverage with real world use cases is really the only blocker here. This change isn't high on my list of priorities right now, so feel free to pick it up if you want. Would love to get this in, since it could clean up some none-stacker stuff that we have internally.

mchristen:
This branch piqued my interest because I think I have a use case that would benefit from being able to do multi-region/account deployments from a single config file.

I started to rebase this locally and most of the major conflicts are due to the changes brought in to use a global S3 client from the us-east-1 region.

Is there still interest in getting some mileage on this code?

ejholmes:
> Is there still interest in getting some mileage on this code?

I'm definitely planning on picking this up at some time again in the future, it just dropped down on my list of priorities. 

I think the only thing that really prevents this from happening is challenges around the S3 bucket that stacker uses for storing templates, and permissions for cross account stuff. I'm sure that can be worked out though.

ejholmes:
Gonna close this for now. I'll probably pick this up again in the future.

PR #409, opened 2017-06-21
fixes #408

### LINKED ISSUE #408: BlueprintTestCase adds extra `/`
The `BlueprintTestCase` helper class is adding an extra `/` in between the OUTPUT_path and `blueprint.name`, as you can see in this example output:

```
IOError: [Errno 2] No such file or directory: 'tests/fixtures/blueprints//kms_key_a.json'
```

**stacker/blueprints/testutil.py**

```
class BlueprintTestCase(unittest.TestCase):
    OUTPUT_PATH = "tests/fixtures/blueprints/"

    def assertEqualsDiff(self, a, b):  # noqa: N802
        self.assertEquals(a, b, diff(a, b))

    def assertRenderedBlueprint(self, blueprint):  # noqa: N802
        expected_output = "%s/%s.json" % (self.OUTPUT_PATH, blueprint.name)
        rendered = blueprint.template.to_json()
```

The fix is simple, remove the trailing slash from OUTPUT_PATH

russellballestrini:
modified:   stacker/blueprints/testutil.py

### DIFF
--- a/stacker/blueprints/testutil.py
+++ b/stacker/blueprints/testutil.py
@@ -15,7 +15,7 @@ def diff(a, b):
 
 
 class BlueprintTestCase(unittest.TestCase):
-    OUTPUT_PATH = "tests/fixtures/blueprints/"
+    OUTPUT_PATH = "tests/fixtures/blueprints"
 
     def assertEqualsDiff(self, a, b):  # noqa: N802
         self.assertEquals(a, b, diff(a, b))


russellballestrini:
The test that is failing appears to be related to something in troposphere S3, and appears to be unrelated to my change because it fails locally when I revert it.

phobologic:
I'll merge this once tests are passing in master and get merged here.

PR #410, opened 2017-06-21
Add provider to the dump method for lookups that require it

smverleye:
Currently attempting to dump (`stacker build -d <dir>`) a template that contains lookups like `${output}` fail when attempting to resolve those variables, as they now have a hard requirement on provider. This passes the provider from the `build` command so they can successfully resolve template variables.

### DIFF
--- a/stacker/actions/build.py
+++ b/stacker/actions/build.py
@@ -285,7 +285,7 @@ class Action(BaseAction):
             if outline:
                 plan.outline()
             if dump:
-                plan.dump(directory=dump, context=self.context)
+                plan.dump(directory=dump, context=self.context, provider=self.provider)
 
     def post_run(self, outline=False, dump=False, *args, **kwargs):
         """Any steps that need to be taken after running the action."""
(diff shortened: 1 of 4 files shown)


phobologic:
Pushed a small pep8 change, and some refactoring to cleanup the way the hooks were handled (there was a lot of duplicate code there).  Thanks!

stacker/actions/build.py
```diff
@@ -257,18 +280,15 @@ def _get_dependencies(self):
...
     def pre_run(self, outline=False, dump=False, *args, **kwargs):
         """Any steps that need to be taken prior to running the action."""
-        pre_build = self.context.config.get("pre_build")
-        should_run_hooks = (
-            not outline and
-            not dump and
-            pre_build
+        hooks = self.context.config.get("pre_build")
+        handle_hooks(
+            "post_build",
```
ejholmes: Shouldn't this be `pre_build`? I don't think this is a breaking change since it's only used in logging.

PR #411, opened 2017-06-22
add ssmstore docs

troyready:
Docs for #314

### DIFF
--- a/docs/lookups.rst
+++ b/docs/lookups.rst
@@ -261,7 +261,9 @@ For example::
 
 Encrypted values ("SecureStrings") can also be used, which will be
 automatically decrypted (assuming the Stacker user has access to the
-associated KMS key).
+associated KMS key). Care should be taken when using this with encrypted
+values (i.e. a safe policy is to only use it with ``no_echo`` CFNString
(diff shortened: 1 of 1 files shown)


docs/lookups.rst
```diff
@@ -235,6 +235,37 @@ and then assign UserData in a LaunchConfiguration or Instance to self.get_variab
...
+  $ aws ssm put-parameter --name "MyDBUser" --type "String" \
+      --value "root"
+
+  # In stacker we would reference the value like:
+  DBUser: ${ssmstore us-east-1@MyDBUser}
+
+  # The above would resolve to
+  DBUser: root
+
+Encrypted values ("SecureStrings") can also be used, which will be
```
ejholmes: It might be worth mentioning here that if you use the `ssmstore` lookup for encrypted secrets, the value will be stored in plaintext across AWS services (CloudFormation, S3, etc), so use it wisely (e.g. ensuring that you use CFNString with `no_echo` when passing it down, etc).

troyready: Good thought; I've added a sentence about it.

PR #415, opened 2017-06-23
aws_lambda hook now allows region override, defaults to session region.

GarisonLotus:
This fix by default will use provider.region when the aws_labda pre-hook is called.  There is an optional bucketregion arg allowing the user to override the region for bucket creation of the lambda artifact deployment giving us the ability to go multi-region in the future.


GarisonLotus:
#417 includes fixes that solve the issue this PR was fixing. (and in a better way.)  Keeping this open for now until that gets merged.

Issue #352, opened 2017-03-21
New lookup idea: env


phobologic:
This seems pretty obvious, so not sure why we haven't done it before, but it might be useful to have a lookup that lets you look at environment variables in the shell.

phobologic:
Resolved with #418

PR #418, opened 2017-06-25
Added a lookup for system environment variables.

syphon7:
Goes with issue #352

Adds the ability to use environment variables as a lookup.


phobologic:
Thanks @syphon7  - did some minor cleanup, and threw a doc in the PR.  This is great!

PR #419, opened 2017-06-26
adding self.name documentation to blueprints.rst...

GarisonLotus:
since the question has come up more than once.

I didn't really know where this would go in the blueprints documentation, or if it should be there at all.  So far now, I just threw it at the bottom with an example of both.  Figured since this has been asked more than once in slack, might as well put it in the docs.

### DIFF
--- a/docs/blueprints.rst
+++ b/docs/blueprints.rst
@@ -308,7 +308,7 @@ Below is an annotated example:
 
 
 Utilizing Stack name within your Blueprint
-==================
+==========================================
 
 Sometimes your blueprint might want to utilize the already existing stack name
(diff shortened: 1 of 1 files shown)


docs/blueprints.rst
```diff
@@ -307,6 +307,84 @@ Below is an annotated example:
                                 variables["CloudFormationSpecificType"].ref))
 
 
+Utilizing Stack name within your Blueprint
+==================
```
russellballestrini: "The underline/overline must be at least as long as the title text."

reference: http://docutils.sourceforge.net/docs/user/rst/quickstart.html#sections

GarisonLotus: Updated 👍

PR #433, opened 2017-07-07
Fixes AMI lookup (and adds optional region specification)

dgarbus:
The AMI lookup is currently broken as it attempts to call `get_session()` without passing any arguments, leading to the following error: `get_session() takes exactly 1 argument (0 given)`

This PR modifies the AMI lookup to accept an optional region (similar to some of the other lookups) and falls back to the provider to get the region if it's not explicitly set.

### DIFF
--- a/docs/lookups.rst
+++ b/docs/lookups.rst
@@ -327,6 +327,9 @@ matches the given filters.
 
 Valid arguments::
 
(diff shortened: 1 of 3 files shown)


phobologic:
Thanks @dgarbus !

PR #436, opened 2017-07-11
Make environment file optional

ejholmes:
This makes the `environment` argument optional, meaning you can invoke stacker `build/destroy` with only a stacker.yml file:

```console
$ echo "namespace: my-namespace" | stacker build -
```

Previously, you were required to pass both an "environment" file _and_ a stacker config file. The only reason the environment file was required was to determine the `namespace`. However, we already parse and interpolate the stacker config using the environment file, so the `namespace` might as well just be part of the stacker config.

In other words, instead of doing this:

```yaml
# dev.env
namespace: my-namespace
```

```yaml
# stacker.yaml
stacks:
- name: vpc
   ...
```

The recommended approach would be to do this:

```yaml
# dev.env
namespace: my-namespace
```

```yaml
# stacker.yaml
namespace: ${namespace}
stacks:
- name: vpc
   ...
```

Which I think is more consistent with the other keys that are already in the stack config (e.g. `namespace_delimiter`, `tags`, etc).

This change should be 100% backwards compatible. If the `namespace` key is not provided in the stack config, then it'll fallback to pulling it from the environment file, with a deprecation warning.

**TODO**

* [x] Update docs
* [x] Deprecation warning about moving `namespace` from environment to config.

### DIFF
--- a/stacker/context.py
+++ b/stacker/context.py
@@ -3,12 +3,15 @@ import logging
 import sys
 
 from stacker.util import SourceProcessor
+from .exceptions import MissingConfig
 from .config import parse_config
-from .exceptions import MissingEnvironment
 from .stack import Stack
 from .lookups import register_lookup_handler
 
 logger = logging.getLogger(__name__)
+
+
+DEFAULT_NAMESPACE_DELIMITER = "-"
 
 
 def get_fqn(base_fqn, delimiter, name=None):
@@ -31,12 +34,10 @@ class Context(object):
     the command line and specified in the config to `Stack` objects.
 
     Args:
-        namespace (str): A unique namespace for the stacks being built.
         environment (dict): A dictionary used to pass in information about
             the environment. Useful for templating.
         stack_names (list): A list of stack_names to operate on. If not passed,
             usually all stacks defined in the config will be operated on.
-        mappings (dict): Used as Cloudformation mappings for the blueprint.
         config (dict): The configuration being operated on, containing the
             stack definitions.
         force_stacks (list): A list of stacks to force work on. Used to work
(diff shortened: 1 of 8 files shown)


stacker/context.py
```diff
@@ -30,62 +32,80 @@ class Context(object):
...
+                or "stacker-%s" % (self.get_fqn(),)
+
+    @property
+    def tags(self):
+        tags = self.config.get("tags", None)
+        if tags is not None:
+            return dict([(str(tag_key), str(tag_value)) for tag_key,
+                         tag_value in tags.items()])
+        else:
+            return {'stacker_namespace': self.namespace}
```
phobologic: You know, I think this logic was actually a bug. I can't imagine a time when you wouldn't want to include the `stacker_namespace` tag, and it doesn't seem obvious that when you provide custom tags it wouldn't provide this one as well.

stacker/context.py
```diff
@@ -30,62 +32,80 @@ class Context(object):
...
+                DEFAULT_NAMESPACE_DELIMITER)
+
+    @property
+    def bucket_name(self):
+        return self.config.get("stacker_bucket") \
+                or "stacker-%s" % (self.get_fqn(),)
+
+    @property
+    def tags(self):
+        tags = self.config.get("tags", None)
```
phobologic: I think this logic could be cleaned up to:

```
tags = {}
for key, value in self.config.get("tags", {}):
    tags[str(key)] = str(value)

if self.namespace:
    tags["stacker_namespace"] = self.namespace
return tags
```

stacker/context.py
```diff
@@ -30,62 +32,80 @@ class Context(object):
...
+        # For backwards compatibility, we fallback to attempting to
+        # fetch the namespace from the environment file, if provided.
+        namespace = self.environment.get("namespace")
+        if namespace:
+            logger.warn("specifying namespace in the environment is "
+                        "deprecated, and should be moved to the config")
+            return namespace
+
+        # No namespace
+        return ""
```
phobologic: This feels like it will have far ranging consequences - have you looked at all the places where this might cause issues?  Why would we want to allow a blank namespace?  Seems like an easy way for folks to shoot themselves in the foot.

ejholmes: So, this was previously supported before this change by doing this in the env file:

```
namespace:
```

And I think it's definitely something we should continue supporting.

The only real change her is falling back to no namespace if the key isn't provided in the stacker config. We could keep namespace as a required key (e.g. if you don't want to have a namespace, then you have to pass `namespace: ""`), which would be more consistent with the existing behavior.

stacker/context.py
ejholmes: I figured as much, but didn't want to make the change here, since it seems potentially dangerous. If someone has a stacker config out in the wild to specifies `tags:`, and they update stacker, every single stack is gonna get an update on it with the new `stacker_namespace` tag.

If we wanna make that change, we should probably do it in a separate PR so the consequences can be documented. People can always "opt in" to the existing behavior by doing this in their config:

```yaml
tags:
  stacker_namespace: ${namespace}
  other_tag: ...
```

stacker/context.py
phobologic: Yeah, I think that's the right thing to do.  Good call.

stacker/context.py
phobologic: So, we could just skip setting `stacker_namespace` if it's set manually also.  But I'm fine w/ having a new PR/issue for this.

stacker/context.py
ejholmes: See the comment below.

ejholmes:
@phobologic ok, I addressed the comments here in https://github.com/remind101/stacker/pull/436/commits/235ed59d288efa7c0e71ce5620ef826064792e72. If this looks good to you, lemme know and I'll squash this.

Issue #437, opened 2017-07-11
Better functional testing suite


ejholmes:
We've talked about this internally a bit. We definitely need a better functional testing suite that actually exercises all of the stacker commands. Even just some "sanity" tests would provide a lot of value, and prevent breakages in commands that don't get run manually very frequently.

phobologic:
I agree - I'd love to discuss this more.  It'd be good to come up with an initial implementation list.

ejholmes:
We now have a functional test suite.

PR #439, opened 2017-07-13
Functional test suite

ejholmes:
This is a quick spike into https://github.com/remind101/stacker/issues/437.

This adds some _really_ basic end-to-end sanity tests that exercise a `stacker build/destroy` using [bats](https://github.com/sstephenson/bats).

Assuming you have bats installed, and some AWS access keys set, you can run this with:

```console
$ brew install bats
$ source venv/bin/activate
$ STACKER_NAMESPACE=<my-stacker-tests-namespace> bats tests
 ✓ stacker build - no config
 ✓ stacker build - empty config
 ✓ stacker build - config with no stacks
 ✓ stacker build - config with no namespace
 ✓ stacker build - missing environment key
 ✓ stacker build - missing variable
 ✓ stacker build - simple build destroy
 ✓ stacker build - simple build with environment
 ✓ stacker build - overriden environment key with -e

9 tests, 0 failures
```

We'd likely want to run this in CI using an isolated AWS account with access keys that are super tight (only allows cloudformation actions, and the ability to create a WaitCondtionHandle).

### DIFF
--- a/tests/test_helper.bash
+++ b/tests/test_helper.bash
@@ -1,0 +1,34 @@
+#!/usr/bin/env bash
+
+if [ -z "$STACKER_NAMESPACE" ]; then
+  >&2 echo "To run these tests, you must set a STACKER_NAMESPACE environment variable"
+  exit 1
+fi
+
+# Simple wrapper around the builtin bash `test` command.
+assert() {
+  builtin test "$@"
+}
+
+# Checks that the given line is in $output.
+assert_has_line() {
+  echo "$output" | grep "$@" 1>/dev/null
+}
+
+# This helper wraps "stacker" with bats' "run" and also outputs debug
+# information. If you need to execute the stacker binary _without_ calling
+# "run", you can use "command stacker".
+stacker() {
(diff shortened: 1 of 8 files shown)


tests/test_helper.bash
```diff
@@ -0,0 +1,34 @@
+#!/usr/bin/env bash
+
+if [ -z "$STACKER_NAMESPACE" ]; then
```
phobologic: maybe make this a random string if not specified?  it's pretty easy to generate one

ejholmes: Good idea. I think in CI we'll have a set prefix, and then append the build ID on it so that concurrent builds can't effect one another.

phobologic: Cool, sounds right!

ejholmes:
Alright, I spun up a new isolated AWS account, and created some access keys using the `FunctionalTests` blueprint in here.

The functional tests will now run automatically on the `master` branch only. PR's from fork's won't have access to the AWS access keys here, so I think I've covered any potential security/abuse concerns.

PR #440, opened 2017-07-18
update package_sources to use USERPROFILE variable on Windows

troyready:
~~This adds a conditional check for the correct environment variable name
to use.~~ Updated to handle this automatically without a condition.

### DIFF
--- a/stacker/util.py
+++ b/stacker/util.py
@@ -467,7 +467,7 @@ class SourceProcessor():
                                         Default to $HOME/.stacker
         """
         stacker_cache_dir = (stacker_cache_dir or
-                             os.path.join(os.environ['HOME'], '.stacker'))
+                             os.path.expanduser("~/.stacker"))
         package_cache_dir = os.path.join(stacker_cache_dir, 'packages')
         self.stacker_cache_dir = stacker_cache_dir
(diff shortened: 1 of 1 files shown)


troyready:
FWIW, I did some basic sanity checking and found everything to work properly on Windows except:

1) This (of course the one thing that broke was something I wrote)

and

2) #441 (which has a decent workaround until someone is motivated to fix it)

stacker/util.py
```diff
@@ -466,8 +466,13 @@ def __init__(self, stacker_cache_dir=None):
             stacker_cache_dir (string): Directory of stacker local cache.
                                         Default to $HOME/.stacker
         """
+        if os.name == 'nt':  # if windows
+            home_env_name = 'USERPROFILE'
+        else:
+            home_env_name = 'HOME'
         stacker_cache_dir = (stacker_cache_dir or
-                             os.path.join(os.environ['HOME'], '.stacker'))
```
phobologic: I think `os.path.expanduser` is what you want here.  It should be cross-platform ready, so you should be able to do:

```
os.path.expanduser("~/.stacker")
```

And I believe it should just work.

troyready:
@phobologic ah, TIL -- thanks! PR updated.

PR #443, opened 2017-07-20
Structured Config object

ejholmes:
While I was working on adding the [functional test suite](https://github.com/remind101/stacker/pull/439), I encountered a lot of scenarios where you could provide stacker with an invalid config file and get back ugly and confusing errors.

This implements a structured `stacker.config.Config` object, which would replace the unstructed dict returned from `parse_config` that we pass to `stacker.Context`. `stacker.config.Config` is a first class object that represents the stacker config format in Python, and can be pre-validated for sanity (e.g. `namespace` key is provided, `stacks` isn't empty, each stack has a `name` and `class_path`, etc).

You could also use this as a consumer of stacker to generate valid stacker config files (almost like troposphere for stacker). For example:

```python
#!/usr/bin/env python

from stacker.config import dump, Config, Stack

vpc = Stack({
    "name": "vpc",
    "class_path": "blueprints.VPC"})

bastion = Stack({
    "name": "bastion",
    "class_path": "blueprints.Bastion",
    "variables": {
        "VpcId": "${output vpc::VpcId}"}})

config = Config()
config.namespace = "prod"
config.stacks = [vpc, bastion]

config.validate()

print dump(config)
```

```console
$ ./gen.py | stacker build -
```

### DIFF
--- a/stacker/config/__init__.py
+++ b/stacker/config/__init__.py
@@ -1,16 +1,69 @@
+import sys
+import logging
+
 from string import Template
 from StringIO import StringIO
 
+from schematics import Model
+from schematics.exceptions import ValidationError
+from schematics.exceptions import BaseError as SchematicsError
+from schematics.types import (
+    ModelType,
+    ListType,
+    StringType,
+    BooleanType,
+    DictType,
+    BaseType
+)
+
 import yaml
 
+from ..lookups import register_lookup_handler
+from ..util import SourceProcessor
 from .. import exceptions
 
 # register translators (yaml constructors)
 from .translators import *  # NOQA
 
+logger = logging.getLogger(__name__)
 
-def parse_config(raw_config, environment=None):
-    """Parse a config, using it as a template with the environment.
+
+def render_parse_load(raw_config, environment=None, validate=True):
+    """Encapsulates the render -> parse -> validate -> load process.
+
+    Args:
+        raw_config (str): the raw stacker configuration string.
+        environment (dict, optional): any environment values that should be
+            passed to the config
+        validate (bool): if provided, the config is validated before being
+            loaded.
+
+    Returns:
+        :class:`Config`: the parsed stacker config.
+
+    """
+
+    rendered = render(raw_config, environment)
+
+    config = parse(rendered)
+
+    # For backwards compatibility, if the config doesn't specify a namespace,
+    # we fall back to fetching it from the environment, if provided.
+    if config.namespace is None:
+        namespace = environment.get("namespace")
+        if namespace:
+            logger.warn("specifying namespace in the environment is "
+                        "deprecated, and should be moved to the config")
+            config.namespace = namespace
+
+    if validate:
+        config.validate()
+
+    return load(config)
+
+
+def render(raw_config, environment=None):
+    """Renders a config, using it as a template with the environment.
 
     Args:
         raw_config (str): the raw stacker configuration string.
@@ -18,10 +71,11 @@ def parse_config(raw_config, environment=None):
             passed to the config
 
     Returns:
-        dict: the stacker configuration populated with any values passed from
+        str: the stacker configuration populated with any values passed from
             the environment
 
     """
+
     t = Template(raw_config)
     buff = StringIO()
     if not environment:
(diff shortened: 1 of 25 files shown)


stacker/config/__init__.py
```diff
@@ -37,3 +37,164 @@ def parse_config(raw_config, environment=None):
...
+
+    Returns:
+        :class:`Config`: the parse (and validated) stacker config.
+
+    """
+
+    config = yaml.load(raw, Loader=Loader) or Config()
+
+    if validate:
+        config.validate()
```
ejholmes: I think it might be better if this returned a list of errors, rather than raising an exception, so we can print all of the errors at once.

stacker/config/__init__.py
```diff
@@ -37,3 +37,164 @@ def parse_config(raw_config, environment=None):
...
+class Stack(Object):
+    yaml_tag = "!Stack"
+
+    def __setstate__(self, state):
+        self.name = state.get("name")
+        self.class_path = state.get("class_path")
+        self.requires = state.get("requires")
+        self.locked = state.get("locked", False)
+        self.enabled = state.get("enabled", True)
+        self.variables = state.get("variables")
```
ejholmes: This could probably use more structure, so that lookups are parsed out at this point in time. That'd help a lot with cleaning up the crazy regex's we currently have for doing this, making a lot of the internals for resolving lookups simpler.

ejholmes:
@phobologic I've updated this to base the config off of schematics, which makes it really clean, and will make it really easy to render a pretty list of validation errors. Good call.

If you like the overall `Config` type, I can add another commit to this that swaps out the internals to use the new structured `Config` type.

ejholmes:
Alrighty, added a second commit to this that swaps the internals over to the structured config object. This commit is basically the smallest change that can be made to make both test suites pass (basically just changing `parse_config` to return `Config` instead if `dict` and fixing any failing tests).

I'll follow up with a third commit that cleans up unit tests so that everything is passing a structured `Config` to `Context` instead of a `dict`.

ejholmes:
Added a third commit to this that cleans up unit tests to ensure that we're always passing in structured Config objects instead of `dict`'s.

@phobologic if you wanna give this a round of review, I'm pretty happy with it.

ejholmes:
Also, I haven't done much to try and make the `InvalidConfig` exception pretty, since we don't have a very good existing method for showing errors that don't involve stack traces. Right now, `InvalidConfig` just gets raised.

For example, here's the error when my config specifies a stack with no name:

```console
   $ stacker build /dev/fd/63
   [2017-07-21T19:16:48] Using Default AWS Provider
     File "/Users/ejholmes/src/github.com/remind101/stacker/venv/bin/stacker", line 6, in <module>
       exec(compile(open(__file__).read(), __file__, 'exec'))
     File "/Users/ejholmes/src/github.com/remind101/stacker/scripts/stacker", line 8, in <module>
       stacker.configure(args)
     File "/Users/ejholmes/src/github.com/remind101/stacker/stacker/commands/stacker/__init__.py", line 39, in configure
       validate=True)
     File "/Users/ejholmes/src/github.com/remind101/stacker/stacker/config/__init__.py", line 51, in render_parse_load
       config.validate()
     File "/Users/ejholmes/src/github.com/remind101/stacker/stacker/config/__init__.py", line 281, in validate
       raise exceptions.InvalidConfig(e.errors)
   stacker.exceptions.InvalidConfig: {'stacks': {0: {'name': ConversionError([ErrorMessage(u'This field is required.', None)])}}}
```

It's still better than before, since the error is a little clearer (stacker at index 0 has an error on `name`).

stacker/config/__init__.py
```diff
@@ -35,5 +80,202 @@ def parse_config(raw_config, environment=None):
...
+    post_destroy = ListType(ModelType(Hook), serialize_when_none=False)
+
+    tags = DictType(StringType, serialize_when_none=False)
+
+    mappings = DictType(
+        DictType(DictType(StringType)), serialize_when_none=False)
+
+    lookups = DictType(StringType, serialize_when_none=False)
+
+    stacks = ListType(ModelType(Stack), default=[], validators=[more_than_one])
```
ejholmes: Would probably be good to add a validator on here that ensures no stacks have the same name (e.g. you accidentally overwrite an existing stack, I've done this).

ejholmes: Added a commit that adds this validation.

stacker/config/__init__.py
```diff
@@ -35,5 +89,210 @@ def parse_config(raw_config, environment=None):
...
+    """
+
+    return yaml.safe_dump(
+        config.to_primitive(),
+        default_flow_style=False,
+        encoding='utf-8',
+        allow_unicode=True)
+
+
+def more_than_one(value):
```
phobologic: The current logic is `one_or_more` rather than `more_than_one`

stacker/config/__init__.py
```diff
@@ -35,5 +89,210 @@ def parse_config(raw_config, environment=None):
...
+
+    def validate(self):
+        try:
+            super(Config, self).validate()
+        except SchematicsError as e:
+            raise exceptions.InvalidConfig(e.errors)
+
+    def validate_stacks(self, data, value):
+        if value:
+            names = {}
```
phobologic: You can use a `set()` here, which is basically just a dict behind the scenes, but makes more sense for your use here.

ejholmes: Good call. Fixed in https://github.com/remind101/stacker/pull/443/commits/0d79c40cc8528ff280ff1b1e3e33cf51b67d77cc

stacker/config/__init__.py
ejholmes: Derp. Fixed in https://github.com/remind101/stacker/pull/443/commits/fe9c8cf771a8f887148572090f4dd6c316005124

ejholmes:
Fixed both comments. I'll go ahead and squash this then merge.

Issue #288, opened 2017-01-03
Make it easier to get started with stacker


mhahn:
I used stacker for a small project over break and it was painful to get up and running with it. A part of this was some unmerged PRs like `stacker blueprint info` as well as the dynamodb blueprint, but it was also annoying to have to remember the following:

- create an env file
- create a config file
- invoke stacker
- install private blueprints

`stacker init`
- something similar to `npm init` where we ask some questions with defaults that sets up the bare minimum file system that stacker depends on:
* conf directory
* yaml config
* dev env file
- i think this would make it easier to get stacker running within new projects faster

I think this would have taken care of my last two points (invoking and installing private blueprints):
https://github.com/remind101/stacker/blob/master/scripts/docker-stacker

but the stacker version within the docker image isn't release-1.0

`stacker blueprint info` was really useful when getting back into stacker after being away from it for awhile. I think adding first class support for explaining `troposphere` types as well as maybe even discovering community blueprints would make this even better.

troyready:
After you mentioned coming back to stacker in Slack, I started thinking about the on-boarding experience and came to many of the same conclusions.

Strongly agree about the `init` subcommand in particular. Hardest part will probably be the color of the bike...I mean the exact directory layout to do.

For the stacker-docker, you should be able to reference the `release-1.0` tag to use the current version (maybe that was something fixed after your post here)

The only thing I'd like to do differently is handling private stacks. Installing them as python packages seems fine enough to me in docker, but handling the different versions of an blueprints via a python install in general isn't appealing to me. I'd rather have the ability to reference them directly (#332) or something akin to it.

russellballestrini:
I'm currently on-boarding with stacker myself and have been running in similar feels.  I plan to look into developing the `init` command over the next few weeks.

russellballestrini:
Another update: I converted my `stacker-helloworld` into a `Cookiecutter` project template:
 https://github.com/remind101/stacker_cookiecutter

Assuming `Cookiecutter` is installed the usage is really simple:

```
cookiecutter gh:remind101/stacker_cookiecutter
```
A new project directory will be created in your present working directory.

dkondratovych:
`init` would help a lot, it's not clear that `python setup.py install` should be executed

PR #451, opened 2017-07-25
Allow directly uploading templates to CloudFormation

ejholmes:
I think this would help with https://github.com/remind101/stacker/issues/288, and make the "first run" experience with stacker a little more pleasant. In most cases, your templates will eventually get large enough that you want to always upload them to S3, but it's nice to be able to play around with stacker without leaving zombie buckets around, or worrying about namespace collisions.

If `stacker_bucket` is explicitly set to `""`, then templates won't be uploaded to S3. Also, if there's no namespace set, and no `stacker_bucket` set, then it will default to this behavior, since we can't auto generate a bucket name. This makes the simplest working stacker config/command:

```console
$ cat <<-EOF | stacker build -
namespace: ''
stacks:
- name: dummy-vpc
  class_path: stacker.tests.fixtures.mock_blueprints.Dummy
EOF
```

Which will work for anyone (no bucket/namespace clashing).

### DIFF
--- a/stacker/actions/build.py
+++ b/stacker/actions/build.py
@@ -1,6 +1,7 @@
 import logging
 
 from .base import BaseAction
+from ..providers.base import Template
 from .. import util
 from ..exceptions import (
     MissingParameterException,
@@ -233,7 +234,7 @@ class Action(BaseAction):
         stack.resolve(self.context, self.provider)
 
         logger.debug("Launching stack %s now.", stack.fqn)
-        template_url = self.s3_stack_push(stack.blueprint)
+        template = self._template(stack.blueprint)
         tags = build_stack_tags(stack)
         parameters = self.build_parameters(stack, provider_stack)
 
@@ -241,7 +242,7 @@ class Action(BaseAction):
         if not provider_stack:
             new_status = SubmittedStatus("creating new stack")
             logger.debug("Creating new stack: %s", stack.fqn)
-            self.provider.create_stack(stack.fqn, template_url, parameters,
+            self.provider.create_stack(stack.fqn, template, parameters,
                                        tags)
         else:
             if not should_update(stack):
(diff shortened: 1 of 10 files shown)


stacker/context.py
```diff
@@ -66,8 +66,26 @@ def namespace_delimiter(self):
...
+
         return self.config.stacker_bucket \
-                or "stacker-%s" % (self.get_fqn(),)
+            or "stacker-%s" % (self.get_fqn(),)
+
+    @property
+    def upload_templates_to_s3(self):
+        # Don't upload stack templates to S3 if `stacker_bucket` is explicitly
+        # set to an empty string.
+        if self.config.stacker_bucket == '':
```
phobologic: Can you throw some debug statements into this logic, just so it's easy to debug?

stacker/providers/aws/default.py
```diff
@@ -15,6 +15,13 @@
 MAX_TAIL_RETRIES = 5
 
 
+def template_args(template):
```
phobologic: Needs a docstring.

stacker/actions/build.py
```diff
@@ -232,30 +233,36 @@ def _launch_stack(self, stack, **kwargs):
...
-                self.provider.update_stack(stack.fqn, template_url,
+                self.provider.update_stack(stack.fqn, template,
                                            existing_params, parameters, tags)
                 logger.debug("Updating existing stack: %s", stack.fqn)
             except StackDidNotChange:
                 return DidNotChangeStatus()
 
         return new_status
 
+    def _template(self, blueprint):
```
phobologic: needs a docstring

stacker/providers/base.py
```diff
@@ -31,3 +31,9 @@ def get_outputs(self, stack_name, *args, **kwargs):
     def get_output(self, stack_name, output):
         # pylint: disable=unused-argument
         return self.get_outputs(stack_name)[output]
+
+
+class Template(object):
```
phobologic: docstring

stacker/providers/aws/default.py
phobologic: Also - any reason to not make this a method of the Template object itself?

phobologic:
@ejholmes ping

stacker/providers/aws/default.py
ejholmes: > Also - any reason to not make this a method of the Template object itself?

Mainly because it's specific to the AWS implementation of the provider, whereas `stacker.providers.base` isn't implementation specific.

ejholmes:
@phobologic updated this with doc strings and debug statements.

PR #452, opened 2017-07-26
Uniform deprecation warnings w/ URL references to readthedocs

GarisonLotus:
Hopefully I caught them all! 👍 

It doesn't look like we have a test for the blueprint deprecation warning. To be honest, I doubt there are many out there on stacker 0.6.

### DIFF
--- a/stacker/config/__init__.py
+++ b/stacker/config/__init__.py
@@ -226,7 +226,7 @@ class Stack(Model):
             raise ValidationError(
                     "DEPRECATION: Stack definition %s contains "
                     "deprecated 'parameters', rather than 'variables'. You are"
-                    " Required to update your config. See https://stacker.rea"
+                    " required to update your config. See https://stacker.rea"
                     "dthedocs.io/en/latest/config.html#variables for "
                     "additional information."
                     % stack_name)
--- a/stacker/config/__init__.py
+++ b/stacker/config/__init__.py
@@ -52,8 +52,11 @@ def render_parse_load(raw_config, environment=None, validate=True):
     if config.namespace is None:
         namespace = environment.get("namespace")
         if namespace:
-            logger.warn("specifying namespace in the environment is "
-                        "deprecated, and should be moved to the config")
+            logger.warn("DEPRECATION WARNING: specifying namespace in the "
+                        "environment is deprecated, see "
+                        "https://stacker.readthedocs.io/en/latest/config.html"
+                        "#namespace "
+                        "for more info.")
             config.namespace = namespace
 
     if validate:
--- a/stacker/config/__init__.py
+++ b/stacker/config/__init__.py
@@ -224,10 +224,11 @@ class Stack(Model):
         if value:
             stack_name = data['name']
             raise ValidationError(
-                    "DEPRECATION WARNING: Stack definition %s contains "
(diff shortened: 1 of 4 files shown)


stacker/config/__init__.py
```diff
@@ -221,9 +224,11 @@ def validate_parameters(self, data, value):
         if value:
             stack_name = data['name']
             raise ValidationError(
-                    "Stack definition %s contains deprecated "
-                    "'parameters', rather than 'variables'. Please "
-                    "update your config." % stack_name)
+                    "DEPRECATION WARNING: Stack definition %s contains "
+                    "deprecated 'parameters', rather than 'variables'. Please "
+                    "update your config. See https://stacker.readthedocs.io/en"
```
ejholmes: I think you're missing a `for` here:


> See https://stacker.readthedocs.io/en/latest/config.html#variables for additional information.

stacker/config/__init__.py
```diff
@@ -221,9 +224,11 @@ def validate_parameters(self, data, value):
         if value:
             stack_name = data['name']
             raise ValidationError(
-                    "Stack definition %s contains deprecated "
-                    "'parameters', rather than 'variables'. Please "
-                    "update your config." % stack_name)
+                    "DEPRECATION WARNING: Stack definition %s contains "
```
ejholmes: Hmm, technically these were only deprecated in the 0.8 branch. In 1.0, they're not technically deprecated, since they were removed. I think we still have this in here for safety. Is that right @phobologic?

Big 👍 on including a link to the docs though.

GarisonLotus:
Just noticed one more outdated reference to parameters instead of variables.  Figured I'd change it since this PR was still open.

stacker/config/__init__.py
```diff
@@ -221,9 +224,12 @@ def validate_parameters(self, data, value):
         if value:
             stack_name = data['name']
             raise ValidationError(
-                    "Stack definition %s contains deprecated "
-                    "'parameters', rather than 'variables'. Please "
-                    "update your config." % stack_name)
+                    "DEPRECATION: Stack definition %s contains "
```
phobologic: Should this be `DEPRECATION WARNING` to be consistent with the others, or did you mean for this to be different on purpose?

stacker/tests/test_config.py
```diff
@@ -157,8 +157,10 @@ def test_parse_with_deprecated_parameters(self):
         error = ex.exception.errors['stacks'][0]['parameters'][0]
         self.assertEquals(
             error.__str__(),
-            "Stack definition vpc contains deprecated 'parameters', rather "
-            "than 'variables'. Please update your config.")
+            "DEPRECATION: Stack definition vpc contains deprecated "
```
phobologic: Same question here.

phobologic:
@GarisonLotus ping

GarisonLotus:
@phobologic - I replied in line, did you not see it.. or did you not agree with it?  Eric brought up one of the existing duplication warming was not a warning, but now a completely removed feature. So in that case, I wanted it to be more of a deprecation notice, with instructions on how to fix and a URL. 

Let me know your thoughts on that or if you want me to change it.

phobologic:
Thanks @GarisonLotus

PR #453, opened 2017-07-26
importing stacker version so that readthedocs will get the version info

GarisonLotus:
no more permanent 0.5.0 in read the docs.

Do we want the version to match the release?


GarisonLotus:
Included in PR 455

PR #459, opened 2017-08-01
add blueprint `to_json` method for standalone rendering

troyready:
Enables blueprint files to use a few simple lines to output their template:
```
if __name__ == "__main__":
    from stacker.context import Context
    print BPClassName('test', Context({"namespace": "test"})).to_json()
```

### DIFF
--- a/stacker/blueprints/base.py
+++ b/stacker/blueprints/base.py
@@ -3,6 +3,7 @@ import hashlib
 import logging
 import string
 from stacker.util import read_value_from_path
+from stacker.variables import Variable
 
 from troposphere import (
     Parameter,
@@ -468,6 +469,16 @@ class Blueprint(object):
         rendered = self.template.to_json()
         version = hashlib.md5(rendered).hexdigest()[:8]
         return (version, rendered)
+
+    def to_json(self):
+        """Render the blueprint and return the template in json form."""
+
+        variables = []
+        for k in self.get_parameter_definitions():
+            variables.append(Variable(k, 'unused_value'))
+        self.resolve_variables(variables)
+
+        return self.render_template()[1]
 
     def read_user_data(self, user_data_path):
         """Reads and parses a user_data file.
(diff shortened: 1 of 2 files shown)


stacker/blueprints/base.py
```diff
@@ -462,6 +463,16 @@ def render_template(self):
         version = hashlib.md5(rendered).hexdigest()[:8]
         return (version, rendered)
 
+    def to_json(self):
+        """Render the blueprint and return the template in json form."""
+
+        variables = []
+        for k in self.get_parameter_definitions():
+            variables.append(Variable(k, 'unused_value'))
```
phobologic: Doesn't this break in the case of the variable not accepting a string type?

phobologic: Maybe we could do something a little more intelligent here to cover the most basic cases - the basic python types are probably the easiest.

troyready:
@phobologic All object types should be fine as long as they have a default value assigned to them; it's only CFN parameters in my experience that trip this up because they don't need really need default values for the template to be generated.

I'm not sure if it's too lazy to say that non-CFN parameters will require a default to use this method? Or maybe the method should extended to take something like a list Variable objects as a parameter to be passed to `resolve_variables`?

phobologic:
@troyready Oh, I like the idea of this taking a dictionary of `variable type -> variable value`.  We could even provide a default set for most python types.  Another option is to have it call the value type directly, since they all should provide a default value:


```
>>> str()
''
>>> int()
0
>>> dict()
{}
>>> list()
[]
```

Mix that with the ability to provide a map for default values, and I think we have something pretty awesome here.  This would make the `dump` command possible again as well possibly :)

stacker/blueprints/base.py
troyready: `get_parameter_definitions` exclusively returns CFN parameters, so they will always take a string.

troyready:
@phobologic I've finally gotten around to updating this and think it's in a good place now.

I've added the override dictionary, but couldn't find a simple way to set default value types. There's probably a simple method I'm missing, but I don't think it unreasonable to leave out anyway -- requiring that defaults be specified in the VARIABLES dictionary itself is nice for keeping things explicit.

Issue #463, opened 2017-08-17
`KeyError: 'Replacement'` thrown by interactive.py when stacker build used with `--replacements-only` on simple SG addition


GarisonLotus:
tested with Stacker 1.0.4

Our automated builds get triggered based on changes in a git repo. To be safe, we execute the stacker build with the `--replacements-only` flag in interactive mode. I ran into an issue and I have narrowed down a repeatable test cause where I could cause stacker to crash when I simply tried to add a single additional security group rule to an existing ruleset in a CFN stack. All changed was a new single security group rule added in a quick loop I created.    Below is the output of my stacker run in jenkins:

```
15:50:30 --------------------------------------------------------------------------------
15:50:30 building the following environment...
15:50:30 --------------------------------------------------------------------------------
15:50:30 The environment is: cgm
15:50:30 The stacker-config is: z_garison_test-core.yaml
15:50:30 The region is: us-east-1
15:50:30 
15:50:30 tags:
15:50:30   "team": Operations
15:50:30   "project": garison-test
15:50:30   "component": garison-test
15:50:30   "stacker_namespace": ${namespace}
15:50:30 
15:50:30 stacks:
15:50:30   - name: garison-test
15:50:30     class_path: blueprints.SecurityGroup.create
15:50:30     variables:
15:50:30       VpcId: ${corevpc}
15:50:30       SecurityGroupNameTag: garison-test
15:50:30       SecurityGroupDesc: garison-test Security Group
15:50:30 
15:50:30   - name: garison-test-adding-single-new-rule-to-SG
15:50:30     class_path: blueprints.SecurityGroup.addIngressRules
15:50:30     variables:
15:50:30       rules:
15:50:30         - GroupId=${output garison-test::GroupID},IpProtocol=tcp,FromPort=8080,ToPort=8080,CidrIp=${ipblock_private1}
15:50:30         - GroupId=${output garison-test::GroupID},IpProtocol=tcp,FromPort=8080,ToPort=8080,CidrIp=${ipblock_private2}
15:50:32 --------------------------------------------------------------------------------
15:50:32 Assuming Rights in AWS account ...
15:50:32 --------------------------------------------------------------------------------
15:50:33 Note: resources requiring replacement will fail this jenkins job. To allow resource replacement, check Authorize_exisiting_resource_replacement checkbox ...
15:50:34 --------------------------------------------------------------------------------
15:50:34 running stacker build using assumed rights...
15:50:34 --------------------------------------------------------------------------------
[...]

PR #464, opened 2017-08-18
Potential resolution of Issue #463 `--replacements-only` set, but no `Replacement` key in the ResourceChange back from CFN due to addition only

GarisonLotus:
I dug around this a bit, and it seems like this fix would work.  I don't see how this code path would ever not be executed when the --replacmenets-only flag is set.  Perhaps I missed something and this seemed way too easy to resolve (and I don't know what side effects this change will make)

If you were curious, this is what we got back from CFN when I printed it out to the shell in my test:

```
[{u'ResourceChange': {u'Action': 'Add', u'ResourceType': 'AWS::EC2::SecurityGroupIngress', u'Scope': [], u'Details': [], u'LogicalResourceId': 'RuleNumber1'}, u'Type': 'Resource'}]
```
No "Replacements" key existed, so yeah... this nich issue would always happen in this specific scenario.  

Is there another way you would prefer doing this?

### DIFF
--- a/stacker/providers/aws/interactive.py
+++ b/stacker/providers/aws/interactive.py
@@ -39,9 +39,11 @@ def requires_replacement(changeset):
         list: A list of changes that require replacement, if any.
 
     """
-    return [r for r in changeset if r["ResourceChange"]["Replacement"] ==
-            'True']
-
+    try:
+        return [r for r in changeset if r["ResourceChange"]["Replacement"] ==
+                'True']
+    except KeyError:
+        pass
 
 def get_raw_input(message):
     """ Just a wrapper for raw_input for testing purposes. """
--- a/stacker/providers/aws/interactive.py
+++ b/stacker/providers/aws/interactive.py
@@ -39,11 +39,8 @@ def requires_replacement(changeset):
         list: A list of changes that require replacement, if any.
 
     """
-    try:
-        return [r for r in changeset if r["ResourceChange"]["Replacement"] ==
-                'True']
-    except KeyError:
-        pass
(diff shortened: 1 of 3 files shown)


GarisonLotus:
Resolves issue 463

GarisonLotus:
@troyready, @jkalscheuer-sturdy and I spoke about options for this today at Sturdy. I will be working on a better/safer way to implement this.

Also going to add a test so we don't regress on this.  I suspect that not many people use the --replacements_only option, which is why this went unnoticed for so long.

GarisonLotus:
ok - no longer a try/catch... negative side effects of that was if you have any non "replacement" keys, it would just pass. That could be bad.

writing tests now.

GarisonLotus:
ok, functional test written.  I Think this is ready to merge, unless you guys find something else wrong with it.

stacker/providers/aws/interactive.py
```diff
@@ -39,8 +39,8 @@ def requires_replacement(changeset):
         list: A list of changes that require replacement, if any.
 
     """
-    return [r for r in changeset if r["ResourceChange"]["Replacement"] ==
-            'True']
+    return [r for r in changeset if "Replacement" in r["ResourceChange"] and r[
```
phobologic: I think this might be better written as:

```
return [r for r in changeset if r["ResourceChange"].get("Replacement", False) == "True"]
```

GarisonLotus:
@phobologic - changed.  Of course that works as well and the tests passed.

phobologic:
Looks great, thanks @GarisonLotus!

PR #466, opened 2017-08-23
fixing a KeyError

GarisonLotus:
This has bothered me for awhile, although it's not something that happens very often. The use case where I have run into this exception is when I use `--stacks` and specify a single stack that uses an output lookup from another stack.  The keyerror was the stack name.  So now we catch it and throw the existing StackDoesNotExist exception.

I also updated the text in the exception to help thouse that don't understand why it's being thrown.

### DIFF
--- a/stacker/actions/build.py
+++ b/stacker/actions/build.py
@@ -276,7 +276,6 @@ class Action(BaseAction):
                 run_func=self._launch_stack,
                 requires=dependencies.get(stack_name),
             )
-
         return plan
 
     def _get_dependencies(self):
--- a/stacker/actions/build.py
+++ b/stacker/actions/build.py
@@ -267,11 +267,14 @@ class Action(BaseAction):
         stacks = self.context.get_stacks_dict()
         dependencies = self._get_dependencies()
         for stack_name in self.get_stack_execution_order(dependencies):
-            plan.add(
-                stacks[stack_name],
-                run_func=self._launch_stack,
-                requires=dependencies.get(stack_name),
-            )
+            try:
+                plan.add(
+                    stacks[stack_name],
+                    run_func=self._launch_stack,
+                    requires=dependencies.get(stack_name),
+                )
+            except KeyError:
+                raise StackDoesNotExist(stack_name)
         return plan
 
     def _get_dependencies(self):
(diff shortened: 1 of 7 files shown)


GarisonLotus:
I want to write a unit test here, but I probably won't have time for a few days.

GarisonLotus:
Oops, I just noticed I got my changes I was going to push into a seperate PR for modifying the test yaml... let me revert that real quick so this PR is only related to this KeyError.

GarisonLotus:
Test for my KeyError exception being caught works: https://codecov.io/gh/remind101/stacker/pull/466/src/stacker/actions/build.py#L276

Just a note here - I added an extra non-related test since I noticed a scenario we didn't really test real well: the `-stacks` flag while specifying stacks that did not match the yaml passed in. It's covered now.

stacker/actions/build.py
```diff
@@ -267,11 +267,14 @@ def _generate_plan(self, tail=False):
         stacks = self.context.get_stacks_dict()
         dependencies = self._get_dependencies()
         for stack_name in self.get_stack_execution_order(dependencies):
-            plan.add(
-                stacks[stack_name],
-                run_func=self._launch_stack,
-                requires=dependencies.get(stack_name),
-            )
+            try:
```
phobologic: It would probably be safer to only wrap the call that can throw the Exception you're seeing - in this case, I believe it's the call to `stacks[stack_name]` not something inside `plan.add`.  If that's the case, I'd change this to:

```
try:
    stack = stacks[stack_name]
except KeyError:
    raise StackDoesNotExist(stack_name)

plan.add(
    stack,
    ...
)
```

What do you think?

GarisonLotus: @phobologic, yeah.. You are correct in the issue is stacks[stack_name] and a more simply way to check this would be better than wrapping the whole plan.add.

GarisonLotus:
@phobologic - I tried first doing exactly what you suggested, but then a `make test` was showing a F841 due to me not using the variable. So I removed the variable assignment. 

stacker/actions/build.py:271:17: F841 local variable 'stackname' is assigned to but never used

I went with just 
```
            try:
                stacks[stack_name]
            except KeyError:
                raise StackDoesNotExist(stack_name)
```

Works and tests passing.

phobologic:
@GarisonLotus in my example I had you using the `stack` variable later in the `plan.add()` - did you not use that, and instead call `stacks[stack_name]` again possibly?

GarisonLotus:
@phobologic - I totally missed that part of your comment. :-)

Updated to assign stacks[stack_name] to a variable that is used in the plan.

phobologic:
Awesome, thanks @GarisonLotus !

PR #469, opened 2017-08-26
Remove Interactive Provider

phobologic:
This is in preparation for "protected mode" for stacks, by allowing the
caller of `update_stack` on the provider to choose whether or not to force
interactive mode.

This also rolls back a small diff when we were preparing to use the
interactive provider when using the `diff` command.


danielkza:
I use the interactive provider for most stacks when I'm developing or updating a blueprint. Does it make sense to allow the stack to force, but still make it possible to enable manually in the command line?

danielkza:
@phobologic Yep, 100% my fault for not reading the code before asking. Removing the duplicate code seems like a great idea.

phobologic:
Thanks!

phobologic:
Thanks folks!

PR #472, opened 2017-08-27
Add protected mode for stacks

### LINKED ISSUE #392: protected config for stacks
Right now we have a few stacks (mostly RDS databases) that we've had locked for ever (basically since ew created them), and they haven't been updated via stacker that entire time.  Now, almost 2 years later, we're too afraid to actually run cloudformation/stacker against them because of what might happen.

This made me realize that `locked` is probably a bad pattern in the first place, though it was really our only option when stacker was first created, because changesets didn't exist.

What I really want is a hybrid interactive/non-interactive mode.  Where it acts like the regular mode unless it hits a stack that is `protected` and will result in a change.

The way we could do this is just make the default provider use changesets as well, but automatically submit them no matter what they are - unless the stack has `protected: true`, in which case it drops to the interactive console.

This seems like the best of both worlds - you get the hands-off mode until something you don't expect to change tries to change.

phobologic:
Fixes #392

This requires #469 to be merged first, and shouldn't be reviewed till
that is done (lots of duplicate code).

I had a hard time coming up with any additional tests - any ideas on
that would be appreciated.  bats doesn't seem to allow any sort of
interaction with the command, so I can't really test the interactive
mode changes easily.


ejholmes:
> bats doesn't seem to allow any sort of
interaction with the command, so I can't really test the interactive
mode changes easily.

You just need to provide some input on STDIN, right? Something like this should do the trick:

```console
echo "y" | stacker build <(config)
```

ejholmes:
This looks pretty good from a code standpoint, but I'm curious how it works for UX. Previously, we'd disable the loop logger when `--interactive` was provided, since it overwrites output. It looks like if you set a stack as protected, the loop logger is gonna overwrite the changset output.

tests/suite.bats
```diff
@@ -347,3 +347,69 @@ EOF
...
+  
+EOF
+  }
+
+  teardown() {
+    stacker destroy --force <(config)
+  }
+
+  # First create the stack
+  export PROTECTED="false"
```
ejholmes: I think you could just pass this in the line below:

```console
stacker build --interactive <(PROTECTED=false config)
```

Issue #393, opened 2017-06-02
UX flow with initial stacker build failures and rollbacks


GarisonLotus:
# Feedback welcome - this is an open discussion item
I wanted to start a conversation regarding this: In stacker, when I run a stacker build on an initial deployment of a brand new stack that happens to fail, the stack itself goes into a ROLLBACK_IN_PROGRESS status eventually ending up in a ROLLBACK_COMPLETE state with no way to run build again.  This is the expected behavior with stacker as it matches cloudformation.

However for a user who is an AWS novice, getting out of this state with stacker... is a different experience than someone uploading templates in the CFN console directly.

I have had several conversations with our in-house development team about their use of Stacker around their user experience they have had with it since a lot of them are using it now for the first time. If a developer with no AWS or CFN experience were to run into this problem, stacker actually appears to hinder their progress moreso vs CFN natively. Let me explain why by going over both flows:

## CFN console native UX workflow
In the AWS console, if I were to incorrectly create a template that fails, I get stuck in a ROLLBACK_IN_PROGRESS, followed by a ROLLBACK_COMPLETE state that is unrecoverable. I can then look at the template itself to figure out the problem or the AWS console to determine what needs to be fixed. After resolution and when I got to resubmit, the UI doesn't allow me to perform the "update stack" operations, instead only allows you to select "delete stack", thus walking you through their intended flow of deleting the stack first. 

In this flow (although not perfect), a user is pretty much told... you must delete this stack before you can try again. It's the only option, so it's natural for a user to chose it.

## Stacker UX workflow
In stacker, when I do a stacker build of that same mistake CFN deployment (via a similar mistake in the blueprint let's say), I get told that the stack is in a ROLLBACK_IN_PROGRESS state, but nothing else.  I can then look at the template itself to figure out the problem, or the AWS console, just like with CFN natively.  When I go to resubmit via stacker build, I get told an unfriendly ` An error occurred (ValidationError) when calling the CreateChangeSet operation` exception due to the stack being actually in a "ROLLBACK_COMPLETE" state and I have to figure out what is going on and how to fix it.

These are very different experiences.

## UX flow issues with stacker

So, I have two issues I can see with the stacker the above workflow:

- The user is never told the actual state the stack eventually ends up in. They are shown that a rollback is in progress, but not that they have to do something above and beyond fixing their template due to it actually being in a rollback complete state.

- On the next build, the user is sent on a red herring trip to search google for whatever `CreateChangeSet operation` means.

[...]

PR #473, opened 2017-08-27
Add handling of stack rollbacks and recreations

danielkza:
Currently stacker bails out at the first sight of an error when handling a stack. While that works for simple cases, it does not make it easy to handle rollbacks, re-creations and other exceptional situations automatically.

To improve that, implement:
- Waiting for rollbacks of each stack during deployment
- Propagate failures from stacks to ther dependents
- Support deleting and re-creating stacks that failed their initial deployment, when confirmed interactively by an user, or through a command line option
- Fail plans gracefully by printing which stacks did not succeed after completing everything, instead of just a traceback

To make sure nothing is broken in the process:
- Re-implement the build action tests to cover multiple different cases and the corresponding state transitions
- Add integration tests that cover rollbacks during new stacks and updates, including dependencies
- Add integration tests for re-creating stacks that previously failed

This PR can only land after #469.

### DIFF
--- a/stacker/actions/base.py
+++ b/stacker/actions/base.py
@@ -1,8 +1,10 @@
 import copy
 import logging
+import sys
 
 import botocore.exceptions
 from stacker.session_cache import get_session
+from stacker.exceptions import PlanFailed
 
 from stacker.util import (
     ensure_s3_bucket,
@@ -121,9 +123,13 @@ class BaseAction(object):
         return template_url
 
     def execute(self, *args, **kwargs):
-        self.pre_run(*args, **kwargs)
-        self.run(*args, **kwargs)
-        self.post_run(*args, **kwargs)
+        try:
+            self.pre_run(*args, **kwargs)
+            self.run(*args, **kwargs)
+            self.post_run(*args, **kwargs)
+        except PlanFailed as e:
+            logger.error(e.message)
+            sys.exit(1)
 
     def pre_run(self, *args, **kwargs):
         pass
--- a/stacker/actions/build.py
+++ b/stacker/actions/build.py
@@ -290,8 +290,6 @@ class Action(BaseAction):
                 logger.debug("Updating existing stack: %s", stack.fqn)
                 return SubmittedStatus("updating existing stack")
             else:
-                logger.debug("Destroying stack for recreation: %s",
-                             stack.fqn)
                 return SubmittedStatus("destroying stack for re-creation")
         except StackDidNotChange:
             return DidNotChangeStatus()
--- a/stacker/actions/build.py
+++ b/stacker/actions/build.py
@@ -1,6 +1,7 @@
 import logging
 
 from .base import BaseAction
+
 from ..providers.base import Template
 from .. import util
 from ..exceptions import (
@@ -16,6 +17,7 @@ from ..status import (
     DidNotChangeStatus,
     SubmittedStatus,
     CompleteStatus,
+    FailedStatus,
     SUBMITTED
 )
 
@@ -217,17 +219,41 @@ class Action(BaseAction):
             provider_stack = None
 
         old_status = kwargs.get("status")
+        recreate = False
         if provider_stack and old_status == SUBMITTED:
             logger.debug(
                 "Stack %s provider status: %s",
                 stack.fqn,
                 self.provider.get_stack_status(provider_stack),
             )
-            if self.provider.is_stack_completed(provider_stack):
-                submit_reason = getattr(old_status, "reason", None)
-                return CompleteStatus(submit_reason)
+
+            if self.provider.is_stack_rolling_back(provider_stack):
+                if 'rolling back' in old_status.reason:
+                    return old_status
+
+                logger.debug("Stack %s entered a roll back", stack.fqn)
+                if 'updating' in old_status.reason:
+                    reason = 'rolling back update'
+                else:
+                    reason = 'rolling back new stack'
+
+                return SubmittedStatus(reason)
             elif self.provider.is_stack_in_progress(provider_stack):
                 logger.debug("Stack %s in progress.", stack.fqn)
+                return old_status
+            elif self.provider.is_stack_destroyed(provider_stack):
+                logger.debug("Stack %s finished deleting", stack.fqn)
+                recreate = True
+                # Continue with creation afterwards
+            elif self.provider.is_stack_failed(provider_stack):
+                reason = old_status.reason
+                if 'rolling' in reason:
+                    reason = reason.replace('rolling', 'rolled')
+
+                return FailedStatus(reason)
+            elif self.provider.is_stack_completed(provider_stack):
+                return CompleteStatus(old_status.reason)
+            else:
                 return old_status
 
         logger.debug("Resolving stack %s", stack.fqn)
@@ -238,26 +264,34 @@ class Action(BaseAction):
         tags = build_stack_tags(stack)
         parameters = self.build_parameters(stack, provider_stack)
 
-        new_status = None
-        if not provider_stack:
-            new_status = SubmittedStatus("creating new stack")
+        if recreate:
+            logger.debug("Re-creating stack: %s", stack.fqn)
+            self.provider.create_stack(stack.fqn, template, parameters,
+                                       tags)
+            return SubmittedStatus("re-creating stack")
+        elif not provider_stack:
             logger.debug("Creating new stack: %s", stack.fqn)
             self.provider.create_stack(stack.fqn, template, parameters,
                                        tags)
-        else:
-            if not should_update(stack):
-                return NotUpdatedStatus()
-            try:
-                new_status = SubmittedStatus("updating existing stack")
+            return SubmittedStatus("creating new stack")
+        elif not should_update(stack):
+            return NotUpdatedStatus()
+
+        try:
+            if self.provider.prepare_stack_for_update(stack.fqn, tags):
                 existing_params = provider_stack.get('Parameters', [])
                 self.provider.update_stack(stack.fqn, template,
                                            existing_params, parameters, tags,
                                            force_interactive=stack.protected)
-                logger.debug("Updating existing stack: %s", stack.fqn)
-            except StackDidNotChange:
-                return DidNotChangeStatus()
 
-        return new_status
+                logger.debug("Updating existing stack: %s", stack.fqn)
+                return SubmittedStatus("updating existing stack")
+            else:
+                logger.debug("Destroying stack for recreation: %s",
+                             stack.fqn)
+                return SubmittedStatus("destroying stack for re-creation")
+        except StackDidNotChange:
+            return DidNotChangeStatus()
 
     def _template(self, blueprint):
         """Generates a suitable template based on whether or not an S3 bucket
--- a/stacker/actions/build.py
+++ b/stacker/actions/build.py
@@ -245,6 +245,9 @@ class Action(BaseAction):
                 logger.debug("Stack %s finished deleting", stack.fqn)
                 recreate = True
                 # Continue with creation afterwards
+            # Failure must be checked *before* completion, as both will be true
+            # when completing a rollback, and we don't want to consider it as
+            # a successful update.
             elif self.provider.is_stack_failed(provider_stack):
                 reason = old_status.reason
                 if 'rolling' in reason:
--- a/stacker/exceptions.py
+++ b/stacker/exceptions.py
@@ -195,3 +195,14 @@ class UnableToExecuteChangeSet(Exception):
                    "%s" % (change_set_id, stack_name, execution_status))
 
         super(UnableToExecuteChangeSet, self).__init__(message)
+
+
+class StackUpdateBadStatus(Exception):
+
+    def __init__(self, stack_name, stack_status, reason, *args, **kwargs):
+        self.stack_name = stack_name
+        self.stack_status = stack_status
+
+        message = ("Stack: \"%s\" cannot be updated nor re-created from state "
+                   "%s: %s" % (stack_name, stack_status, reason))
+        super(StackUpdateBadStatus, self).__init__(message, *args, **kwargs)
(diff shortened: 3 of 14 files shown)


danielkza:
Rebased to latest master.

GarisonLotus:
This might allow closure of one of our oldest issues: #8

GarisonLotus:
Also, when merged, we will be able to close #393 - UX flow with initial stacker build failures and rollbacks.

danielkza:
Rebased again!

GarisonLotus:
so I played around with this tonight with several different scenarios in mind. I feel that everything is damn near perfect with this PR and I thank danielkza for putting in the effort I wish I had time for.   My only suggestion, is the text displayed when trying to redeploy into a `rollback_complete` stack doesn't truly walk through a novice stacker user through the easy path of resolution. My only suggestion (this is my being picky), is the error message presented when not in interactive mode stating `Re-creation disallowed by user option`should be modified. I agree that as is, this is very direct and to the point, it just doesn't convey they only need to run it again with your new flag of `--recreate-failed`. Something like `Re-creation disallowed by default, rerun with --recreate-failed to force relaunch` would solve that.

This is only a suggestion however, everything here seems to work and could be merged as is.


It would also be nice to add some documentation.

danielkza:
>  Something like Re-creation disallowed by default, rerun with --recreate-failed to force relaunch would solve that.

That's a wonderful suggestion and completely sensible, I'll improve the message.

> It would also be nice to add some documentation.

I agree, but I didn't find a good place to add it. Maybe the [commands](https://stacker.readthedocs.io/en/latest/commands.html) page? Or a new page entirely?

GarisonLotus:
Yeah, I don't see why the commands page wouldn't work.  It's where I would look.

You're awesome! Thanks.

stacker/actions/base.py
```diff
@@ -121,9 +123,13 @@ def s3_stack_push(self, blueprint, force=False):
...
     def execute(self, *args, **kwargs):
-        self.pre_run(*args, **kwargs)
-        self.run(*args, **kwargs)
-        self.post_run(*args, **kwargs)
+        try:
+            self.pre_run(*args, **kwargs)
+            self.run(*args, **kwargs)
+            self.post_run(*args, **kwargs)
+        except PlanFailed as e:
+            logger.error(e.message)
```
phobologic: This would hide the traceback stack, right?  Should we instead do `logger.exception`?

stacker/plan.py
```diff
@@ -197,11 +209,23 @@ def _single_run(self):
         """Executes a single run through the plan, touching each step."""
         for step_name, step in self.list_pending():
             waiting_on = []
+            failed = False
             for required_stack in step.requires:
+                if self[required_stack].failed:
+                    logger.debug(
+                        'Failing stack \"%s\" due to failed dependency '
+                        '\"%s\"', step_name, required_stack)
+                    step.set_status(FailedStatus("dependency has failed"))
```
phobologic: think it's worth identifying the dependency here?

stacker/providers/aws/default.py
```diff
@@ -362,30 +363,63 @@ def create_change_set(cfn_client, fqn, template, parameters, tags,
     return changes, change_set_id
 
 
+def check_tags_present(actual, expected):
```
phobologic: could you write a small docstring for this for the api docs?

stacker/providers/aws/default.py
```diff
@@ -362,30 +363,63 @@ def create_change_set(cfn_client, fqn, template, parameters, tags,
...
+    actual_set = set((item["Key"], item["Value"]) for item in actual)
+    expected_set = set((item["Key"], item["Value"]) for item in expected)
+
+    return actual_set >= expected_set
```
phobologic: It's ok if the actual set has items the expected set doesn't?

stacker/providers/aws/default.py
```diff
@@ -362,30 +363,63 @@ def create_change_set(cfn_client, fqn, template, parameters, tags,
...
+        "UPDATE_ROLLBACK_IN_PROGRESS"
+    )
+
+    FAILED_STATUSES = (
+        "CREATE_FAILED",
+        "ROLLBACK_FAILED",
+        "ROLLBACK_COMPLETE",
+        "DELETE_FAILED",
+        "UPDATE_ROLLBACK_FAILED",
+        "UPDATE_ROLLBACK_COMPLETE",
```
phobologic: This is in both Complete & Failed - is that correct?

stacker/actions/build.py
```diff
@@ -238,26 +264,34 @@ def _launch_stack(self, stack, **kwargs):
...
                                            force_interactive=stack.protected)
-                logger.debug("Updating existing stack: %s", stack.fqn)
-            except StackDidNotChange:
-                return DidNotChangeStatus()
 
-        return new_status
+                logger.debug("Updating existing stack: %s", stack.fqn)
+                return SubmittedStatus("updating existing stack")
+            else:
+                logger.debug("Destroying stack for recreation: %s",
```
phobologic: This doesn't seem to actually do any destroying?

stacker/tests/actions/test_build.py
```diff
@@ -261,6 +200,161 @@ def test_Raises_StackDoesNotExist_from_lookup_non_included_stack(self):
...
+                    'Tags': []}
+
+        patch_object(self.provider, 'get_stack', side_effect=get_stack)
+        patch_object(self.provider, 'update_stack')
+        patch_object(self.provider, 'create_stack')
+        patch_object(self.provider, 'destroy_stack')
+
+        patch_object(self.build_action, "s3_stack_push")
+
+    def _advance(self, new_provider_status, expected_status, expected_reason):
```
phobologic: holy crap I love this new test suite - thanks so much @danielkza, this is great.

One thing I would suggest - should we have tests for when `recreate_failed` is false?

tests/README.md
```diff
@@ -18,9 +18,10 @@ This directory contains the functional testing suite for stacker. It exercises a
   ```console
   $ ./stacker.yaml.sh | stacker build -
   ```
-4. In the AWS console, generate a new IAM access key pair for the user and set it in your shell:
+4. Grab the generated key pair for the user and set it in your shell:
 
   ```console
+  $ ./stacker.yaml.sh | stacker info -
```
phobologic: Hah, this is literally a question I had in my "todo" list at work:

https://www.dropbox.com/s/5tgbm4jl7hd343b/Screenshot%202017-09-08%2013.49.55.png?dl=0

@ejholmes what do you think?

tests/suite.bats
```diff
@@ -409,3 +409,230 @@ EOF
...
+  stacker build <(bad_config)
+  assert "$status" -eq 1
+  assert_has_line "Using default AWS provider mode"
+  assert_has_line "${STACKER_NAMESPACE}-recreate-failed-interactive: pending"
+  assert_has_line "${STACKER_NAMESPACE}-recreate-failed-interactive: submitted (creating new stack)"
+  assert_has_line "${STACKER_NAMESPACE}-recreate-failed-interactive: submitted (rolling back new stack)"
+  assert_has_line "${STACKER_NAMESPACE}-recreate-failed-interactive: failed (rolled back new stack)"
+
+  # Updating the stack should prompt to re-create it.
+  stacker build -i <(good_config) <<< $'y\n'
```
ejholmes: I'm ashamed to admit I've never seen the`$'...'` syntax before. What does this do exactly?

stacker/tests/fixtures/mock_blueprints.py
```diff
@@ -129,6 +140,28 @@ def create_template(self):
...
+    VARIABLES = {
+        "StringVariable": {
+            "type": str,
+            "default": ""}
+    }
+
+    def create_template(self):
+        t = self.template
+        t.add_resource(WaitConditionHandle("BrokenDummy"))
+        t.add_resource(WaitCondition(
```
ejholmes: Can you add a comment on the specific part of this that fails, and why (I'm assuming Timeout?)

tests/README.md
ejholmes: I have an adverse reaction to putting secrets in CloudFormation outputs (plaintext, bad ACL's), but given the user already has a pretty scoped set of permissions, it's prob ok.

tests/suite.bats
```diff
@@ -409,3 +409,230 @@ EOF
...
+  stacker build <(config)
+  assert "$status" -eq 1
+  assert_has_line "Using default AWS provider mode"
+  assert_has_line "${STACKER_NAMESPACE}-dependent-rollback-parent: pending"
+  assert_has_line "${STACKER_NAMESPACE}-dependent-rollback-child:  pending"
+  assert_has_line "${STACKER_NAMESPACE}-dependent-rollback-parent: submitted (creating new stack)"
+  assert_has_line "${STACKER_NAMESPACE}-dependent-rollback-parent: submitted (rolling back new stack)"
+  assert_has_line "${STACKER_NAMESPACE}-dependent-rollback-parent: failed (rolled back new stack)"
+  assert_has_line "${STACKER_NAMESPACE}-dependent-rollback-child:  failed (dependency has failed)"
+  assert_has_line "The following stacks failed their plans: dependent-rollback-parent, dependent-rollback-child"
```
ejholmes: Minor nit. The noun "Plan" has always referred to the entire graph of stacks, so it's confusing for me to see it applied as a noun for an individual stack.

Maybe just:

> The following stacks failed: dependent-rollback-parent, dependent-rollback-child

stacker/actions/base.py
danielkza: `PlainFailed` should only be raised when CF reaches a failure state, which should mean the traceback doesn't help much. Other kinds of exceptions should bubble through and end execution.

stacker/plan.py
danielkza: The listing of failed stacks at the end should be enough to figure out what failed. But being explicit can't hurt.

stacker/providers/aws/default.py
danielkza: Sure, will do.

stacker/providers/aws/default.py
danielkza: Won't it always have at least the automatically generated CloudFormation tags?

stacker/actions/build.py
danielkza: `prepare_stack_for_update` only returns `False` if the stack will be deleted. The mesage could be placed in a better spot to make that more clear.

danielkza: Yes, at the point the destroying has already been scheduled by `prepare_stack_for_update`. I suppose the message could be clearer and/or better placed.

stacker/providers/aws/default.py
danielkza: Yes. An update that contains errors should cause a rollback, but even when the rollback succeeds the overall update still failed, an I expect stacker to fail too.

stacker/tests/fixtures/mock_blueprints.py
danielkza: Timeout has a maximum value, which `2 ** 32` should always exceed. I'll add a comment to explain that.

tests/suite.bats
danielkza: Makes sense and is actually shorter, thanks!

tests/suite.bats
danielkza: It expands character escape sequences and nothing else. Double quotes should have the exact same result in this particular case.

tests/README.md
danielkza: The goal was to avoid having to go to the console or do an API call to grab the key. As @ejholmes  said, since the user is meant to have limited perms. and be temporary it shouldn't be terribly risky.

stacker/tests/actions/test_build.py
danielkza: I think I might just set it to `False` instead, and just enabled it on the tests that actually care about it.

tests/suite.bats
ejholmes: TIL. Thanks!

stacker/tests/fixtures/mock_blueprints.py
danielkza: Documented.

tests/suite.bats
danielkza: Done.

stacker/tests/actions/test_build.py
danielkza: Changed the default to `False`, and only enable it in the re-creation test.

stacker/actions/build.py
danielkza: The message was actually redundant, as one was printed at `warn` level in `prepare_stack_for_update`. I removed it.

stacker/providers/aws/default.py
danielkza: Added comments to point this out in the declaration site and the stack state handling code.

stacker/providers/aws/default.py
danielkza: Done.

stacker/plan.py
danielkza: Updated the message and moved it to `warn` level so it's always seen.

danielkza:
Updated for all the review comments and added docs.

stacker/exceptions.py
```diff
@@ -214,7 +214,6 @@ def __init__(self, failed_stacks, *args, **kwargs):
         self.failed_stacks = failed_stacks
 
         stack_names = ', '.join(stack.name for stack in failed_stacks)
-        message = ("The following stacks failed their plans: %s\n"
-                   % (stack_names,))
+        message = ("The following stacks failed: %s\n" % (stack_names,))
```
phobologic: minor nit: Shouldn't need the outer parenthesis on this, since it doesn't cross 80 characters.

danielkza: Fixed :)

phobologic:
Awesome, thanks for tackling this @danielkza :)

PR #474, opened 2017-08-31
lambda hook new optional follow_symlinks bool

GarisonLotus:
So I asked some developers that use stacker for things that bothered them, this was one of the answers I got.  I figured I'd create a PR with a potential backwards compatible solution. (Default is False, the way it used to work)

Updated docstrings with examples as well.  Let me know if there is a different way you would suggest making this change.

### DIFF
--- a/stacker/hooks/aws_lambda.py
+++ b/stacker/hooks/aws_lambda.py
@@ -384,6 +384,8 @@ def upload_lambda_functions(context, provider, **kwargs):
             use by the provider.
         prefix (str, optional): S3 key prefix to prepend to the uploaded
             zip name.
+        followsymlinks (bool, optional): Will determine if symlinks should
+            be followed and included with the zip artfiact. Default: False
         functions (dict):
             Configurations of desired payloads to build. Keys correspond to
             function names, used to derive key names for the payload. Each
@@ -434,6 +436,7 @@ def upload_lambda_functions(context, provider, **kwargs):
                 data_key: lambda
                 args:
                   bucket: custom-bucket
+                  followsymlinks: True
                   prefix: cloudformation-custom-resources/
                   functions:
                     MyFunction:
--- a/stacker/hooks/aws_lambda.py
+++ b/stacker/hooks/aws_lambda.py
@@ -377,7 +377,7 @@ def upload_lambda_functions(context, provider, **kwargs):
         prefix (str, optional): S3 key prefix to prepend to the uploaded
             zip name.
         follow_symlinks (bool, optional): Will determine if symlinks should
-            be followed and included with the zip artfiact. Default: False
+            be followed and included with the zip artifact. Default: False
         functions (dict):
             Configurations of desired payloads to build. Keys correspond to
             function names, used to derive key names for the payload. Each
@@ -428,7 +428,7 @@ def upload_lambda_functions(context, provider, **kwargs):
                 data_key: lambda
                 args:
                   bucket: custom-bucket
-                  follow_symlinks: True
+                  follow_symlinks: true
                   prefix: cloudformation-custom-resources/
                   functions:
                     MyFunction:
--- a/stacker/hooks/aws_lambda.py
+++ b/stacker/hooks/aws_lambda.py
@@ -93,7 +93,7 @@ def _calculate_hash(files, root):
     return file_hash.hexdigest()
 
 
-def _find_files(root, includes, excludes, walk):
+def _find_files(root, includes, excludes, follow_symlinks):
     """List files inside a directory based on include and exclude rules.
 
     This is a more advanced version of `glob.glob`, that accepts multiple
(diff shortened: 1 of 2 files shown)


GarisonLotus:
Scenarios tested with existing function/yaml: 

1.  no followsymlinks Argument set (the backwards compatible test) - works
2. adding "followsymlinks" Argument with int value - works, ignored
3. adding "followsymlinks" Argument with str value - works, ignored
4. setting up new symlink and adding "followsymlinks" Argument with False bool value - works
5. setting up new symlink and adding "followsymlinks" Argument with True bool value - works

GarisonLotus:
I don't think I realized we ran this hook through our unit tests.  I just noticed that on codecov. Guess I will need to write some tests.

stacker/hooks/aws_lambda.py
```diff
@@ -103,9 +103,11 @@ def _find_files(root, includes, excludes):
         root (str): base directory to list files from.
         includes (list[str]): inclusion patterns. Only files matching those
             patterns will be included in the result.
-        includes (list[str]): exclusion patterns. Files matching those
+        excludes (list[str]): exclusion patterns. Files matching those
             patterns will be excluded from the result. Exclusions take
             precedence over inclusions.
+        walk (bool): If true, symlinks will be included in the resulting
```
phobologic: any reason not to name this `follow_symlinks`?

stacker/hooks/aws_lambda.py
```diff
@@ -116,22 +118,36 @@ def _find_files(root, includes, excludes):
     """
 
     root = os.path.abspath(root)
-    file_set = formic.FileSet(directory=root, include=includes,
-                              exclude=excludes)
+    if walk is True:
```
phobologic: should just be `if walk:`

phobologic: You can actually also avoid this if/else and the `_walk` function entirely with `funtools.partial` like this:

```
file_set = formic.FileSet(directory=root, include=includes,
                          exclude=excludes,
                          walk=partial(os.walk, followlinks=walk)

```

stacker/hooks/aws_lambda.py
```diff
@@ -116,22 +118,36 @@ def _find_files(root, includes, excludes):
...
+
+def _zip_from_file_patterns(root, includes, excludes, walk):
     """Generates a ZIP file in-memory from file search patterns.
 
     Args:
         root (str): base directory to list files from.
         includes (list[str]): inclusion patterns. Only files  matching those
             patterns will be included in the result.
-        includes (list[str]): exclusion patterns. Files matching those
+        excludes (list[str]): exclusion patterns. Files matching those
```
phobologic: Thank you for catching/fixing this :)

stacker/hooks/aws_lambda.py
```diff
@@ -116,22 +118,36 @@ def _find_files(root, includes, excludes):
...
             patterns will be excluded from the result. Exclusions take
             precedence over inclusions.
+        walk (bool): If true, symlinks will be included in the resulting
```
phobologic: I'd go ahead and use `follow_symlinks` here as well.

stacker/hooks/aws_lambda.py
```diff
@@ -468,6 +489,18 @@ def create_template(self):
         provider.region
     )
 
+    # Check if we should walk / follow symlinks
+    followsymlinks = kwargs.get('followsymlinks')
```
danielkza: I'd like it more if `False` was used as a default value, but other non-boolean values were rejected instead of silently ignored. Something like:

```python
follow_symlinks = kwargs.get('follow_symlinks', False)
if not isinstance(follow_symlinks, bool):
    raise ValueError('follow_symlinks option must be a boolean')

# ...
``

File: stacker/hooks/aws_lambda.py [hunk 6]
```diff
@@ -366,6 +384,8 @@ def upload_lambda_functions(context, provider, **kwargs):
             use by the provider.
         prefix (str, optional): S3 key prefix to prepend to the uploaded
             zip name.
+        followsymlinks (bool, optional): Will determine if symlinks should
```
danielkza: I think `follow_symlinks` would be nicer to read and use.

stacker/hooks/aws_lambda.py
GarisonLotus: hmm yeah, I like your idea better. Thanks for the suggestion... updating now.

GarisonLotus:
@phobologic and @danielkza - I updated per your suggestions.  I'm going to write some tests now.

stacker/hooks/aws_lambda.py
phobologic: @GarisonLotus ^^ any reason not to name it follow_symlinks rather than walk?  Seems  more natural.

phobologic:
Hmm - some sort of strange merge happened here @GarisonLotus - not sure what happened, but the PR has a bunch of other stuff that was merged into master.  Can you clean it up?

GarisonLotus:
@phobologic - I removed the erroneous commits that got pulled in from commits merged into master. This PR should now just be my changes. Also included now are four tests written and passing. Tests included are:

1. no follow_symlinks Argument set (the backwards compatible test) - works
2. adding "follow_symlinks" Argument with str value - raises valueerror, that's good
3. setting up new symlink and adding "follow_symlinks" Argument with False bool value - works
4. setting up new symlink and adding "follow_symlinks" Argument with True bool value - works

I realize that tests 1 and 3 are basicly the same thing... but I figured that both are testing different variants of configurations, it was worthy of seperate tests.

phobologic:
Thanks @GarisonLotus - made a small typo fix, but this looks great.

PR #478, opened 2017-09-17
Move S3 templates into sub-dirs in S3

### LINKED ISSUE #347: Add option to specify namespace for S3 bucket
Hey guys,

This issue is more of a nice-to-have, but we have a relatively large number of namespaces and have chosen to use a single `stacker_bucket` for storing stack templates. The problem we are facing is that stacker writes the templates to the root of the chosen bucket, rather than using any kind of namespace. This makes it difficult to identify the exact template we are looking for without actually downloading/inspecting it.

It looks like this is where the object path is set:
https://github.com/remind101/stacker/blob/master/stacker/actions/base.py#L36

It would be great for organization if we could use `$namespace` or some sort of prefix option to group the templates into their respective folders within the bucket.

phobologic:
Fixes #347

This adds another layer to the s3 bucket space to try and keep things a
bit cleaner.  If you had a namespace of `mynamespace`, a bucket of
`mybucket`, and a stack called `vpc`, the old method would have resulted
in an s3 key of:

`s3://mybucket/vpc-$version.json`

It would now result in:

`s3://mybucket/mynamespace-vpc/vpc-$version.json`

Not 100% sure that's the best namespace use, but it seems specific
enough and should be helpful.

I'm pretty sure this will result in slower runs the first time after
updating to this version of stacker, only because it'll have to upload
all it's templates again, but it shouldn't result in any CloudFormation
updates.  I'll verify this.

### DIFF
--- a/stacker/actions/base.py
+++ b/stacker/actions/base.py
@@ -24,7 +24,9 @@ def stack_template_key_name(blueprint):
     Returns:
         string: Key name resulting from blueprint.
     """
-    return "%s-%s.json" % (blueprint.name, blueprint.version)
+    name = blueprint.name
+    return "%s/%s-%s.json" % (blueprint.context.get_fqn(name), name,
+                              blueprint.version)
 
 
 def stack_template_url(bucket_name, blueprint, endpoint):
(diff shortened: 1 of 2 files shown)


phobologic:
Good call, @ejholmes - updated to have the `stack_templates` prefix.

PR #487, opened 2017-10-03
add s3 remote package sources

troyready:
Extends `package_sources` to support tarballs & zips on S3.

### DIFF
--- a/stacker/util.py
+++ b/stacker/util.py
@@ -8,15 +8,20 @@ import re
 import shutil
 import subprocess
 import sys
+import tarfile
 import tempfile
 import time
 import yaml
+import zipfile
 from yaml.constructor import ConstructorError
 from yaml.nodes import MappingNode
 from collections import OrderedDict
 from git import Repo
 
+import botocore.client
 import botocore.exceptions
+
+from stacker.session_cache import get_session
 
 logger = logging.getLogger(__name__)
 
@@ -562,9 +567,68 @@ class SourceProcessor():
 
     def get_package_sources(self):
         """Makes remote python packages available for local use."""
+        # Checkout S3 repositories specified in config
+        for config in self.sources.get('s3', []):
+            self.fetch_s3_package(config=config)
         # Checkout git repositories specified in config
         for config in self.sources.get('git', []):
             self.fetch_git_package(config=config)
+
+    def fetch_s3_package(self, config):
+        """Makes a remote S3 archive available for local use
+
+        Args:
+            config (:class:`stacker.config.GitPackageSource`): git config
+                dictionary
+
+        """
+        if config['key'].endswith('.tar.gz'):
+            archive_type = '.tar.gz'
+            dir_name = self.sanitize_uri_path(
+                "s3-%s-%s" % (config['bucket'], config['key'][:-7])
+            )
+        elif config['key'].endswith('.zip'):
+            archive_type = '.zip'
+            dir_name = self.sanitize_uri_path(
+                "s3-%s-%s" % (config['bucket'], config['key'][:-4])
+            )
+        else:
+            raise ValueError(
+                "Archive type could not be determined for S3 object \"%s\" "
+                "in bucket %s." % (config['key'], config['bucket'])
+            )
+
+        # We can skip downloading the archive if it's already been cached
+        cached_dir_path = os.path.join(self.package_cache_dir, dir_name)
+        if not os.path.isdir(cached_dir_path):
+            tmp_dir = tempfile.mkdtemp(prefix='stacker')
+            tmp_package_path = os.path.join(tmp_dir, dir_name)
+            try:
+                tmp_archive_path = os.path.join(tmp_dir,
+                                                dir_name + archive_type)
+                s3_resource = get_session(region=None).resource('s3')
+                if config.get('requester_pays', False):
+                    extra_args = {'RequestPayer': 'requester'}
+                else:
+                    extra_args = None
+                s3_resource.Bucket(config['bucket']).download_file(
+                    config['key'],
+                    tmp_archive_path,
+                    ExtraArgs=extra_args
+                )
+                if archive_type == '.tar.gz':
+                    with tarfile.open(tmp_archive_path, 'r:gz') as tar:
+                        tar.extractall(path=tmp_package_path)
+                else:
+                    with zipfile.ZipFile(tmp_archive_path, 'r') as zip_ref:
+                        zip_ref.extractall(tmp_package_path)
+                shutil.move(tmp_package_path, self.package_cache_dir)
+            finally:
+                shutil.rmtree(tmp_dir)
+
+        # Update sys.path & merge in remote configs (if necessary)
+        self.update_paths_and_config(config=config,
+                                     pkg_dir_name=dir_name)
 
     def fetch_git_package(self, config):
         """Makes a remote git repository available for local use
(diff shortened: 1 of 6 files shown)


stacker/util.py
```diff
@@ -531,10 +536,69 @@ def create_cache_directories(self):
 
     def get_package_sources(self):
         """Makes remote python packages available for local use."""
+        # Checkout S3 repositories specified in config
```
phobologic: Does the fact that s3 comes before git have implications?  What if someone defines an s3 package source after a git source, could they be potentially confused by the results?

stacker/util.py
```diff
@@ -531,10 +536,69 @@ def create_cache_directories(self):
...
+            self.fetch_s3_package(config=config)
         # Checkout git repositories specified in config
         for config in self.sources.get('git', []):
             self.fetch_git_package(config=config)
 
+    def fetch_s3_package(self, config):
+        """Makes a remote S3 archive available for local use
+
+        Args:
+            config (:class:`stacker.config.GitPackageSource`): git config
```
phobologic: bad class specification?

stacker/util.py
```diff
@@ -531,10 +536,69 @@ def create_cache_directories(self):
...
+                "s3-%s-%s" % (config['bucket'], config['key'][:-4])
+            )
+        else:
+            raise ValueError(
+                "Archive type could not be determined for S3 object \"%s\" "
+                "in bucket %s." % (config['key'], config['bucket'])
+            )
+
+        # We can skip downloading the archive if it's already been cached
+        cached_dir_path = os.path.join(self.package_cache_dir, dir_name)
```
phobologic: The caching logic here seems like it shares similarities with the git package sources - should this be factored out into a common method(s)?

stacker/util.py
```diff
@@ -531,10 +536,69 @@ def create_cache_directories(self):
...
+        if not os.path.isdir(cached_dir_path):
+            tmp_dir = tempfile.mkdtemp(prefix='stacker')
+            tmp_package_path = os.path.join(tmp_dir, dir_name)
+            try:
+                tmp_archive_path = os.path.join(tmp_dir,
+                                                dir_name + archive_type)
+                s3_resource = get_session(region=None).resource('s3')
```
phobologic: How do we handle different s3 regions?  Does it make sense for this to use the default stacker bucket if no bucket is specified?  I spent a good chunk of time trying to get this just right in the other uses of s3, so there's probably some useful code there (see the `aws_lambda` hook as well as the methods we use in the provider).

stacker/util.py
```diff
@@ -531,10 +536,69 @@ def create_cache_directories(self):
...
+    def fetch_s3_package(self, config):
+        """Makes a remote S3 archive available for local use
+
+        Args:
+            config (:class:`stacker.config.GitPackageSource`): git config
+                dictionary
+
+        """
+        if config['key'].endswith('.tar.gz'):
+            archive_type = '.tar.gz'
```
phobologic: What about an object called something like an `Extractor` or something similar, with a simple method/interface.  Then you'd specify the extractor to be used here, and later call it's `.extract` method or something similar?

phobologic: It'd also mean that you could do away with the chain of if/elif statements and instead have a dict that maps suffixes to extractors:

```
extractor_map = {".tar.gz": TarGzipExtractor, ".tar", TarExtractor, ".zip": ZipExtractor}
extractor = None
for suffix, klass in extractor_map.iteritems():
    if config["key"].endswith(suffix):
        extractor = klass
        break
```

troyready:
@phobologic should be better now. The most egregious issue was obviously the double looping over file extensions which has been eliminated with the new classes.

Regarding the `package_sources` ordering:
I don't think the pre-determined order will introduce any additional confusion. The most non-intuitive part in my experience is what happens when sys.path names conflict, which ends up being a pain no matter what (regardless of order or whether it's introduced via the `sys_path` keyword).

Regarding the duplicated caching logic in `fetch_s3_package` & `fetch_git_package`:
I agree that there does seem to be a couple of lines of overlap here (and it'll be even more if I or anyone ever got around to adding more package_sources). I'm not sure the best way to DRY it without making the flow really weird, since the parts before and after it are significantly different.

Regarding s3 regions:
In my testing the simple file download here bypasses most (and hopefully all) of the s3 region issues that normally come up. I've tested it with buckets in different regions in the same config and they all downloaded properly.

troyready:
One thing I forgot to answer: it hadn't occurred to me before to fallback to the stacker bucket for sources, but I kinda like the idea. I'm not sure there's a clean way to get the name at this point (without the Context object), so it seems better to me now to leave the bucket name as an explicit requirement

phobologic:
Thanks for the updates!  One thing I'm curious about: How does one update the s3 package if it changes?  Is it expected that you would change the name of the s3 archive when they upload it, and then modify the config?

If so, I wonder if we should do something a little more complicated here - maybe head the s3 object first to get the timestamp when it was last updated, and if it's newer than the current date, update?

troyready:
@phobologic yeah, works for me. I envisioned most usage as targeting generated 'versioned' artifacts (i.e. stacker_blueprints at tag v1.0.3, so a file of stacker_blueprints-v1.0.3.zip), but I can see the use of using the latest version. I'll add the support

troyready:
@phobologic I've updated the PR to make files updated in place work as expected.

I think the simplified logic I put in should work predictably -- there's no date comparison, just "do I have the version currently on s3?". Since last-modified can only increase, it should give the experience of always using the latest file while avoiding any sort of errors from datetime comparisons.

stacker/util.py
```diff
@@ -505,9 +505,65 @@ def ensure_s3_bucket(s3_client, bucket_name, bucket_region):
...
+    @staticmethod
+    def extension():
+        """Return archive extension."""
+        return '.zip'
+
+
 class SourceProcessor():
     """Makes remote python package sources available in the running python
        environment."""
+    CACHE_DATE_FORMAT = '%Y-%m-%d-%H-%M-%S'
```
phobologic: Hmm, any issues with time zones here?

phobologic: Also, can we go with a standard like iso8601 if we're not going to use a unix timestamp?

stacker/util.py
```diff
@@ -547,53 +603,80 @@ def fetch_s3_package(self, config):
...
-                s3_resource.Bucket(config['bucket']).download_file(
+                extractor.archive = os.path.join(
+                    tmp_dir,
+                    dir_name + extractor.extension()
+                )
+                session.resource('s3').Bucket(config['bucket']).download_file(
                     config['key'],
-                    tmp_archive_path,
-                    ExtraArgs=extra_args
+                    extractor.archive,
```
phobologic: Just to future-proof this some, can we have a method called `set_archive` on the base `Extractor` class that does this?  That way we can override it later if we need to in an easier way.

troyready:
@phobologic good thoughts. I believe I've addressed both and rebased off of master (I can squash it down too if you'd like).

stacker/util.py
```diff
@@ -554,24 +631,108 @@ def __init__(self, sources, stacker_cache_dir=None):
...
+
+        """
+        extractor_map = {'.tar.gz': TarGzipExtractor,
+                         '.tar': TarExtractor,
+                         '.zip': ZipExtractor}
+        extractor = None
+        for suffix, klass in extractor_map.iteritems():
+            if config['key'].endswith(suffix):
+                extractor = klass()
+                dir_name = self.sanitize_uri_path(
```
phobologic: Could you throw a debug statement here detailing the extractor chosen for the given path?

stacker/util.py
```diff
@@ -554,24 +631,108 @@ def __init__(self, sources, stacker_cache_dir=None):
...
+            tmp_dir = tempfile.mkdtemp(prefix='stacker')
+            tmp_package_path = os.path.join(tmp_dir, dir_name)
+            try:
+                extractor.set_archive(os.path.join(tmp_dir, dir_name))
+                session.resource('s3').Bucket(config['bucket']).download_file(
+                    config['key'],
+                    extractor.archive,
+                    ExtraArgs=extra_s3_args
+                )
+                extractor.extract(tmp_package_path)
```
phobologic: Can you throw some debug statements around these two steps?  I feel like things could break here, and knowing where it breaks because of the debug statements will be helpful.

troyready:
@phobologic yeah, sounds good -- they've been added

PR #490, opened 2017-10-09
Update awacs to 0.7.x

bwhaley:
0.7.x has a few properties that I need. Can we bump that dependency in stacker?


phobologic:
Yes, definitely.  I totally should have refreshed our dependencies - troposphere just released 2.0 for example.  Thanks for this - I have another PR coming that'll update more dependencies.

Issue #480, opened 2017-09-18
Lookups cause infinite loop when Python typecasts the value to False


LyndonHook:
The below line causes lookups to go into an infinite loop when the value is an empty string, zero or False. Basically any value that Python casts to False.

https://github.com/remind101/stacker/blob/master/stacker/variables.py#L114

I'm wondering if this is by design? If not can we change it to something like the below?
```
if self._resolved_value is not None:
    return self._resolved_value
else:
    return self._value
```
Happy to create the PR for it too.

phobologic:
Hey @LyndonHook - thanks for the report.  An infinite loop isn't intentional :)  That said, I have yet to see this issue.  Can you give a sample config/blueprint/environment that would cause this?  Also what version of stacker are you seeing this in?  Thanks!

LyndonHook:
The scenario I have is based on custom lookups that return integer values or boolean values. Although you can replicate it with the default lookup using a blank default value.

`CertificateArn: ${default certificate_arn::}`

The above sets the default value to an empty string which evaluates to False. If `self._resolved_value` is "False" then the original value is returned causing `self.lookups` always be a list of 1 or more.

We're currently using 1.0.2 but I've also tested it on 1.0.4.

phobologic:
Ok, yeah, I think your suggested fix is probably the right thing to do.  I can't think of a case where the resolved value should be None.  If you want to submit a PR that'd be awesome :)  Thanks @LyndonHook

LyndonHook:
#494

russellballestrini:
Resolved by #494

PR #494, opened 2017-10-17
Prevent lookup infinite loop

LyndonHook:
Prevent lookup infinite loop when self._resolved_value is type cast to False. Added test that checks by using the default lookup with an empty string.


LyndonHook:
#480

LyndonHook:
Just giving this a bit of a nudge...

phobologic:
Thanks @LyndonHook - looks good, sorry for the delay!

LyndonHook:
No problem @phobologic! Thanks for the merge.

PR #500, opened 2017-10-26
raise error on dupe sibling YAML keys

### LINKED ISSUE #491: Build should fail when duplicate stack names exist in a template
If/when a template has the same stack defined twice (as happens with a bad copy/paste), it should fail validation and abort the run.

russellballestrini:
Fixes #491 

Most of this code is derived from upstream pyYAML library.

https://bitbucket.org/xi/pyyaml/src/10c55a624429f28191ef2bc1d9db44684946eb5d/lib/yaml/constructor.py?at=default&fileviewer=file-view-default#constructor.py-120:135

	modified:   stacker/config/__init__.py
	modified:   stacker/tests/test_config.py
	modified:   stacker/util.py

### DIFF
--- a/stacker/config/__init__.py
+++ b/stacker/config/__init__.py
@@ -355,10 +355,11 @@ class Config(Model):
 
     def validate_stacks(self, data, value):
         if value:
-            names = set()
-            for i, stack in enumerate(value):
-                if stack.name in names:
-                    raise ValidationError(
-                        "Duplicate stack %s found at index %d."
-                        % (stack.name, i))
-                names.add(stack.name)
+            stack_names = [stack.name for stack in value]
+            if len(set(stack_names)) != len(stack_names):
+                # only loop / enumerate if there is an issue.
+                for i, stack_name in enumerate(stack_names):
+                    if stack_names.count(stack_name) != 1:
+                        raise ValidationError(
+                            "Duplicate stack %s found at index %d."
+                            % (stack_name, i))
--- a/stacker/tests/test_config.py
+++ b/stacker/tests/test_config.py
@@ -14,6 +14,8 @@ from stacker.environment import parse_environment
 from stacker import exceptions
 from stacker.lookups.registry import LOOKUP_HANDLERS
 
+from yaml.constructor import ConstructorError
+
 config = """a: $a
 b: $b
 c: $c"""
@@ -97,7 +99,7 @@ class TestConfig(unittest.TestCase):
         error = ex.exception.errors['stacks'][0]
         self.assertEquals(
             error.__str__(),
-            "Duplicate stack bastion found at index 1.")
+            "Duplicate stack bastion found at index 0.")
 
     def test_dump_unicode(self):
         config = Config()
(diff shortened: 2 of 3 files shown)


stacker/config/__init__.py
```diff
@@ -355,10 +356,12 @@ def validate(self):
 
     def validate_stacks(self, data, value):
         if value:
-            names = set()
-            for i, stack in enumerate(value):
-                if stack.name in names:
+            stack_names = map(attrgetter('name'), value)
+            if len(set(stack_names)) == len(stack_names):
```
phobologic: Is returning None a valid return value for schematics validations?  I think to make this a little more clear, you might want to just pull the loop below into the opposite of this if statement.  ie:

```
if len(set(stack_names)) != len(stack_names):
    for i, stack_name in ... etc
```

stacker/tests/test_config.py
```diff
@@ -398,6 +400,36 @@ def test_render_parse_load_namespace_fallback(self):
...
+            parse(yaml_config)
+
+    def test_raise_error_on_duplicate_stack_name(self):
+        yaml_config = """
+        namespace: prod
+        stacks:
+          - name: vpc
+            class_path: blueprints.VPC
+          - name: vpc
+            class_path: blueprints.VPC
```
phobologic: Can you change this class_path to something different?  Just want to make sure it's only testing the stack name, and nothing else.

stacker/util.py
```diff
@@ -296,17 +298,40 @@ def yaml_to_ordered_dict(stream, loader=yaml.SafeLoader):
         OrderedDict: Parsed YAML.
     """
     class OrderedLoader(loader):
-        """Subclass of Python yaml class."""
-        pass
+        """Subclass of the given Python loader yaml class."""
```
phobologic: I think the name of this should probably be changed - maybe `OrderedUniqueLoader` or something?  Also, probably worth describing what it does in the docstring rather than this generic one.

stacker/tests/test_config.py
phobologic: Also, can you write a test that tests that the dictionary version of stacks fails the unique validation?

stacker/config/__init__.py
russellballestrini: Python returns `None` implicitly when a function or method exits without a return.

stacker/config/__init__.py
```diff
@@ -355,10 +356,12 @@ def validate(self):
 
     def validate_stacks(self, data, value):
         if value:
-            names = set()
-            for i, stack in enumerate(value):
-                if stack.name in names:
+            stack_names = map(attrgetter('name'), value)
```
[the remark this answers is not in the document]
russellballestrini: fixed

stacker/config/__init__.py
russellballestrini: fixed

stacker/util.py
russellballestrini: fixed

stacker/tests/test_config.py
russellballestrini: fixed

PR #505, opened 2017-11-13
Add global config for troposphere template indent

smverleye:
AWS CloudFormation limits templates uploaded to S3 to 460,800 bytes (https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cloudformation-limits.html) which we've managed to hit. 

An easy solution to cut out half the size of the templates we're generating is reducing the troposphere template indent.

This adds a setting in the context to change the indent value passed through to troposphere. Default is left at 4 spaces.


phobologic:
Actually, it looks like the tests are having issues - if you have a moment, can you take a look?  Thanks.

smverleye:
I actually did take a look at the failing test cases but couldn't figure out the usage of python mock; do you guys have a guide you used? @phobologic

phobologic:
@smverleye sorry for the delay, I've pushed some minor updates that should fix the issue - can you take a look and let me know if it looks good to you?

From the commit message:
```
    Fixes the tests that were failing

    This was because we were using MagicMock as the Context before, and now
    that context is actually used in blueprints (for the indent level), that
    no longer worked.  I did some minor refactoring of `mock_context` to
    make it a bit more useful as well.
```

smverleye:
Looks great @phobologic thanks, and hopefully I can fix up the test cases myself next time.

phobologic:
Awesome - glad to help, and thanks for the PR @smverleye !

PR #507, opened 2017-11-15
Allow duplicate keys for most mappings except stacks.

russellballestrini:
modified:   stacker/tests/test_config.py
	modified:   stacker/util.py

### DIFF
--- a/stacker/tests/test_config.py
+++ b/stacker/tests/test_config.py
@@ -400,7 +400,24 @@ stacks:
         config.validate()
         self.assertEquals(config.namespace, "prod")
 
-    def test_raise_constructor_error_on_duplicate_key(self):
+    def test_allow_most_keys_to_be_duplicates_for_overrides(self):
+        yaml_config = """
+        namespace: prod
+        stacks:
+          - name: vpc
+            class_path: blueprints.VPC
+            variables:
+              CIDR: 192.168.1.0/24
+              CIDR: 192.168.2.0/24
+        """
+        doc = parse(yaml_config)
+        self.assertEqual(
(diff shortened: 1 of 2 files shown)


acmcelwee:
Is this to allow defaults with overrides or something along those lines? If so, what about warn on dupes if they're not in the hard fail set of keys?

russellballestrini:
Yeah, I prefer my original attempt to fix this issue.

It was found that some end users may be setting a default for a some keys and then redefine the keys later on to override.

https://github.com/remind101/stacker/pull/500

The YAML spec is sort of confusing. It seems to indicate that keys "MUST" be unique in some places.

```
JSON's RFC4627 requires that mappings keys merely “SHOULD” be unique, while YAML insists they “MUST” be. Technically, YAML therefore complies with the JSON spec, choosing to treat duplicates as an error. In practice, since JSON is silent on the semantics of such duplicates, the only portable JSON files are those with unique keys, which are therefore valid YAML files.
```

Reference:  http://www.yaml.org/spec/1.2/spec.html section 1.3. Relation to JSON

stacker/tests/test_config.py
```diff
@@ -408,10 +408,21 @@ def test_raise_constructor_error_on_duplicate_key(self):
             class_path: blueprints.VPC
             class_path: blueprints.Fake
         """
-        with self.assertRaises(ConstructorError):
-            parse(yaml_config)
+        # with self.assertRaises(ConstructorError):
```
phobologic: Can you remove these commented out lines?

phobologic: Also, it seems like this test is named incorrectly.

phobologic:
Thanks @russellballestrini

PR #512, opened 2017-11-22
Fixed typos with strings running together in the help output

mwalling:
I noticed that these strings were running together in the output while looking at something else.


russellballestrini:
@mwalling Thanks!

PR #516, opened 2017-12-06
hooks: keypair: add support for storing keys in SSM parameter store

danielkza:


stacker/hooks/keypair.py
```diff
@@ -24,18 +37,28 @@ def ensure_keypair_exists(provider, context, **kwargs):
...
+            parameter to receive the generated private key, instead of
+            importing it or storing it locally.
+        ssm_key_id (:class:`str`): optional ID of a KMS key to encrypt the SSM
+            parameter with. If ommitted, the default key will be used.
 
     Returns: boolean for whether or not the hook succeeded.
-
     """
+
+    keypair_name = kwargs["keypair"]
```
phobologic: You know, it looks like `keypair` is a required argument - might as well make it a required argument in the function signature as well, what do you think?

stacker/hooks/keypair.py
```diff
@@ -24,18 +37,28 @@ def ensure_keypair_exists(provider, context, **kwargs):
...
+    ssm_parameter_name = kwargs.get("ssm_parameter_name")
```
phobologic: I'd make both of these optional arguments (ie: default to `None`) in the function signature.

stacker/hooks/keypair.py
```diff
@@ -46,62 +69,97 @@ def ensure_keypair_exists(provider, context, **kwargs):
...
-        "import or create keypair \"%s\"? (import/create/Cancel) " % (
-            keypair_name,
-        ),
-    )
-    if create_or_upload == "import":
-        path = raw_input("path to keypair file: ")
-        full_path = utils.full_path(path)
-        if not os.path.exists(full_path):
-            logger.error("Failed to find keypair at path: %s", full_path)
+    if ssm_parameter_name:
```
phobologic: Can you pull this out into a separate function?  It seems like a reasonable amount of code that could live in its own function.

danielkza:
Sorry for taking so long, but I'll implement your suggestions ASAP @phobologic

phobologic:
No worries @danielkza - it likely won't go out in 1.2.0rc1, but we can get it in the next release when you're ready!  Thanks!

danielkza:
Replaced by #715.

PR #517, opened 2017-12-13
add serverless transform / CREATE change set type support

troyready:
Trying to perform a `create_stack` with a template including a
'Transform' generates the following error:
```
botocore.exceptions.ClientError: An error occurred (ValidationError) when calling the CreateStack operation: CreateStack cannot be used with templates containing Transforms.
```

This detects these templates and ensures that they are only
deployed via change sets.

### DIFF
--- a/stacker/actions/build.py
+++ b/stacker/actions/build.py
@@ -266,6 +266,7 @@ class Action(BaseAction):
         template = self._template(stack.blueprint)
         tags = build_stack_tags(stack)
         parameters = self.build_parameters(stack, provider_stack)
+        force_change_set = stack.blueprint.template.transform is not None
 
         if recreate:
             logger.debug("Re-creating stack: %s", stack.fqn)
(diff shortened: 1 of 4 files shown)


stacker/actions/build.py
```diff
@@ -274,18 +275,32 @@ def _launch_stack(self, stack, **kwargs):
             return SubmittedStatus("re-creating stack")
         elif not provider_stack:
             logger.debug("Creating new stack: %s", stack.fqn)
-            self.provider.create_stack(stack.fqn, template, parameters,
-                                       tags)
+            if force_change_set:
```
phobologic: This is a little confusing - I think it likely makes more sense to have this happen inside `create_stack` - maybe provide a flag there that will allow `create_stack` to use changesets with the `CREATE` `ChangeSetType`?

troyready:
@phobologic yeah, I'll take a stab at that. I meant to comment before that I'm not particularly tied to any implementation, just want to fix the "issue".

troyready:
@phobologic it does seem better to have the change in create_stack. I've updated it.

phobologic:
@troyready ok, yeah, this was way easier to follow.  Thanks for taking another stab at it!  I made a small change to a log message, and merged upstream master.  This looks great, thanks!

PR #521, opened 2018-01-05
add stacker.cmd for windows installs

### LINKED ISSUE #441: stacker script doesn't work out of the box on Windows
Windows doesn't do the right thing with [scripts/stacker](https://github.com/remind101/stacker/blob/1.0.4/scripts/stacker) -- just having the shebang at the top of the file doesn't get the python interpreter invoked.

Looks like awscli handles this by shipping a [separate invocation script for windows](https://github.com/aws/aws-cli/blob/1.11.121/bin/aws.cmd).

troyready:
Adapted from awscli's implementation, this allows `stacker` to be executed natively  on Windows

Fixes #441

### DIFF
--- a/scripts/stacker.cmd
+++ b/scripts/stacker.cmd
@@ -1,0 +1,45 @@
+@echo OFF
+REM="""
+setlocal
+set PythonExe=""
+set PythonExeFlags=
+
+for %%i in (cmd bat exe) do (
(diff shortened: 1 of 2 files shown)


phobologic:
I honestly have no ability to really review this, other than from the most simple of reviews, since we don't use windows.  Looks fine to me though!

PR #523, opened 2018-01-26
DAG

ejholmes:
Third times the charm :)

Fixes https://github.com/remind101/stacker/issues/300
Fixes https://github.com/remind101/stacker/issues/186
Fixes https://github.com/remind101/stacker/issues/384

This is an internal refactor to change the Plan implementation to be backed by a Directed Acyclic Graph. Most of this work was already done a long time ago in https://github.com/remind101/stacker/pull/406, but this just brings it up to speed with the master branch, with a few important caveats:

### Not parallelized

This implementation is _not_ parallelized. This PR paves the way to make it easy to implement a 100% parallelized graph walk, but I didn't implement it here, primarily to make review easier, but also because it doesn't matter much based on the current implementation (more on that below). I think we should get the graph in `master`, and then add a parallelized walk after the fact since that will need special attention to deal with concurrency issues (logging, user input, etc).

In theory, this sounds like it should reduce performance, but ironically it has little impact. I benchmarked before and after on our internal stacker graph, which is pretty big:

**before (stacker 1.1.3)**

```console
$ time stacker build

real    8m31.716s
user    0m0.139s
sys     0m0.100s
```

**after (dag)**

```console
$ time stacker build

real    8m33.819s
user    0m0.142s
sys     0m0.094s
```

However, because the `--stacks` flag now works like you'd expect it to, this branch can actually make most workflows a lot faster, even without concurrency.

### Notes for reviewer

For the time being, I'd like to focus on functionality, and making sure everything is feature complete, and all of the integration tests pass. There's probably some missing doc's, and I'll get to those things later.

Also, to make review easier, instead of making changes to `stacker.plan` I made a new `stacker.plan2`, and swapped out call sites to use the new graph based plan implementation. This will likely replace `stacker.plan` before merging, unless we want to keep the old `stacker.plan` around for API compatibility.

The `stacker.dag` module probably doesn't need much attention, since it's already been reviewed previously, and is stilled based on `py-dag`. Feel free to review it if you want.

### TODO

* [x] Bring back the check pointed loop logger (I just changed this to a simplified logger to make refactoring easier)
* [x] Bring back some removed tests.
* [x] Re-implement the `--tail` flag.
* [x] Re-implement dump

### DIFF
--- a/stacker/dag/__init__.py
+++ b/stacker/dag/__init__.py
@@ -1,0 +1,366 @@
+import collections
+import logging
+from copy import copy, deepcopy
+from collections import deque
+
+logger = logging.getLogger(__name__)
+
+try:
+    from collections import OrderedDict
+except ImportError:
+    from ordereddict import OrderedDict
+
+
+class DAGValidationError(Exception):
+    pass
+
+
+class DAG(object):
+    """ Directed acyclic graph implementation. """
+
+    def __init__(self):
+        """ Construct a new DAG with no nodes or edges. """
+        self.reset_graph()
+
+    def add_node(self, node_name):
+        """ Add a node if it does not exist yet, or error out.
+
+        Args:
+            node_name (str): The unique name of the node to add.
+
+        Raises:
+            KeyError: Raised if a node with the same name already exist in the
+                      graph
+        """
+        graph = self.graph
+        if node_name in graph:
+            raise KeyError('node %s already exists' % node_name)
+        graph[node_name] = set()
+
+    def add_node_if_not_exists(self, node_name):
+        """ Add a node if it does not exist yet, ignoring duplicates.
+
+        Args:
+            node_name (str): The name of the node to add.
+        """
+        try:
+            self.add_node(node_name)
+        except KeyError:
+            pass
+
+    def delete_node(self, node_name):
+        """ Deletes this node and all edges referencing it.
+
+        Args:
+            node_name (str): The name of the node to delete.
+
+        Raises:
+            KeyError: Raised if the node does not exist in the graph.
+        """
+        graph = self.graph
+        if node_name not in graph:
+            raise KeyError('node %s does not exist' % node_name)
+        graph.pop(node_name)
+
+        for node, edges in graph.iteritems():
+            if node_name in edges:
+                edges.remove(node_name)
+
+    def delete_node_if_exists(self, node_name):
+        """ Deletes this node and all edges referencing it.
+
+        Ignores any node that is not in the graph, rather than throwing an
+        exception.
+
+        Args:
+            node_name (str): The name of the node to delete.
+        """
+
+        try:
+            self.delete_node(node_name)
+        except KeyError:
+            pass
+
+    def add_edge(self, ind_node, dep_node):
+        """ Add an edge (dependency) between the specified nodes.
+
+        Args:
+            ind_node (str): The independent node to add an edge to.
+            dep_node (str): The dependent node that has a dependency on the
+                            ind_node.
+
+        Raises:
+            KeyError: Either the ind_node, or dep_node do not exist.
+            DAGValidationError: Raised if the resulting graph is invalid.
+        """
+        graph = self.graph
+        if ind_node not in graph:
+            raise KeyError('independent node %s does not exist' % ind_node)
+        if dep_node not in graph:
+            raise KeyError('dependent node %s does not exist' % dep_node)
+        test_graph = deepcopy(graph)
+        test_graph[ind_node].add(dep_node)
+        test_dag = DAG()
+        test_dag.graph = test_graph
+        is_valid, message = test_dag.validate()
+        if is_valid:
+            graph[ind_node].add(dep_node)
+        else:
+            raise DAGValidationError(message)
+
+    def delete_edge(self, ind_node, dep_node):
+        """ Delete an edge from the graph.
+
+        Args:
+            ind_node (str): The independent node to delete an edge from.
+            dep_node (str): The dependent node that has a dependency on the
+                            ind_node.
+
+        Raises:
+            KeyError: Raised when the edge doesn't already exist.
+        """
+        graph = self.graph
+        if dep_node not in graph.get(ind_node, []):
+            raise KeyError("No edge exists between %s and %s." % (
+                ind_node, dep_node
+                )
+            )
+        graph[ind_node].remove(dep_node)
+
+    def transpose(self):
+        """ Builds a new graph with the edges reversed.
+
+        Returns:
+            :class:`stacker.dag.DAG`: The transposed graph.
+        """
+        graph = self.graph
+        transposed = DAG()
+        for node, edges in graph.items():
+            transposed.add_node(node)
+        for node, edges in graph.items():
+            # for each edge A -> B, transpose it so that B -> A
+            for edge in edges:
+                transposed.add_edge(edge, node)
+        return transposed
+
+    def walk(self, walk_func):
+        """ Walks each node of the graph in reverse topological order.
+        This can be used to perform a set of operations, where the next
+        operation depends on the previous operation. It's important to note
+        that walking happens serially, and is not paralellized.
+
+        Args:
+            walk_func (:class:`types.FunctionType`): The function to be called
+                on each node of the graph.
+
+        Returns:
+            bool: True if the function succeeded on every node, otherwise
+                  False.
+        """
+        nodes = self.topological_sort()
+        # Reverse so we start with nodes that have no dependencies.
+        nodes.reverse()
+
+        failed = False
+        for n in nodes:
+            if not walk_func(n):
+                failed = True
+
+        return not failed
+
+    def rename_edges(self, old_node_name, new_node_name):
+        """ Change references to a node in existing edges.
+
+        Args:
+            old_node_name (str): The old name for the node.
+            new_node_name (str): The new name for the node.
+        """
+        graph = self.graph
+        for node, edges in graph.items():
+            if node == old_node_name:
+                graph[new_node_name] = copy(edges)
+                del graph[old_node_name]
+
+            else:
+                if old_node_name in edges:
+                    edges.remove(old_node_name)
+                    edges.add(new_node_name)
+
+    def predecessors(self, node):
+        """ Returns a list of all immediate predecessors of the given node
+
+        Args:
+            node (str): The node whose predecessors you want to find.
+
+        Returns:
+            list: A list of nodes that are immediate predecessors to node.
+        """
+        graph = self.graph
+        return [key for key in graph if node in graph[key]]
+
+    def downstream(self, node):
+        """ Returns a list of all nodes this node has edges towards.
+
+        Args:
+            node (str): The node whose downstream nodes you want to find.
+
+        Returns:
+            list: A list of nodes that are immediately downstream from the
+                  node.
+        """
+        graph = self.graph
+        if node not in graph:
+            raise KeyError('node %s is not in graph' % node)
+        return list(graph[node])
+
+    def all_downstreams(self, node):
+        """Returns a list of all nodes ultimately downstream
+        of the given node in the dependency graph, in
+        topological order.
+
+        Args:
+             node (str): The node whose downstream nodes you want to find.
+
+        Returns:
+            list: A list of nodes that are downstream from the node.
+        """
+        nodes = [node]
+        nodes_seen = set()
+        i = 0
+        while i < len(nodes):
+            downstreams = self.downstream(nodes[i])
+            for downstream_node in downstreams:
+                if downstream_node not in nodes_seen:
+                    nodes_seen.add(downstream_node)
+                    nodes.append(downstream_node)
+            i += 1
+        return filter(
+                lambda node: node
+                in nodes_seen, self.topological_sort())
+
+    def filter(self, nodes):
+        """ Returns a new DAG with only the given nodes and their
+        dependencies.
+
+        Args:
+            nodes (list): The nodes you are interested in.
+
+        Returns:
+            :class:`stacker.dag.DAG`: The filtered graph.
+        """
+
+        filtered_dag = DAG()
+
+        # Add only the nodes we need.
+        for node in nodes:
+            filtered_dag.add_node_if_not_exists(node)
+            for edge in self.all_downstreams(node):
+                filtered_dag.add_node_if_not_exists(edge)
+
+        # Now, rebuild the graph for each node that's present.
+        for node, edges in self.graph.items():
+            if node in filtered_dag.graph:
+                filtered_dag.graph[node] = edges
+
+        return filtered_dag
+
+    def all_leaves(self):
+        """ Return a list of all leaves (nodes with no downstreams)
+
+        Returns:
+            list: A list of all the nodes with no downstreams.
+        """
+        graph = self.graph
+        return [key for key in graph if not graph[key]]
+
+    def from_dict(self, graph_dict):
+        """ Reset the graph and build it from the passed dictionary.
+
+        The dictionary takes the form of {node_name: [directed edges]}
+
+        Args:
+            graph_dict (dict): The dictionary used to create the graph.
+
+        Raises:
+            TypeError: Raised if the value of items in the dict are not lists.
+        """
+
+        self.reset_graph()
+        for new_node in graph_dict.iterkeys():
+            self.add_node(new_node)
+        for ind_node, dep_nodes in graph_dict.iteritems():
+            if not isinstance(dep_nodes, collections.Iterable):
+                raise TypeError('%s: dict values must be lists' % ind_node)
+            for dep_node in dep_nodes:
+                self.add_edge(ind_node, dep_node)
+
+    def reset_graph(self):
+        """ Restore the graph to an empty state. """
+        self.graph = OrderedDict()
+
+    def ind_nodes(self):
+        """ Returns a list of all nodes in the graph with no dependencies.
+
+        Returns:
+            list: A list of all independent nodes.
+        """
+        graph = self.graph
+
+        dependent_nodes = set(
+            node for dependents
+            in graph.itervalues() for node in dependents)
+        return [node for node in graph.keys() if node not in dependent_nodes]
+
+    def validate(self):
+        """ Returns (Boolean, message) of whether DAG is valid. """
+        if len(self.ind_nodes()) == 0:
+            return (False, 'no independent nodes detected')
+        try:
+            self.topological_sort()
+        except ValueError as e:
+            return (False, e.message)
+        return (True, 'valid')
+
+    def topological_sort(self):
+        """ Returns a topological ordering of the DAG.
+
+        Returns:
+            list: A list of topologically sorted nodes in the graph.
+
+        Raises:
+            ValueError: Raised if the graph is not acyclic.
+        """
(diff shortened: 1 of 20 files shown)


stacker/dag/__init__.py
```diff
@@ -0,0 +1,264 @@
...
+class DAG(object):
+    """ Directed acyclic graph implementation. """
+
+    def __init__(self):
+        """ Construct a new DAG with no nodes or edges. """
+        self.reset_graph()
+
+    def add_node(self, node_name, graph=None):
+        """ Add a node if it does not exist yet, or error out. """
+        if not graph:
```
phobologic: Do you have any idea why this library does this so often?  I haven't seen this pattern used really ever before.  I thought maybe for testing, but it seems like a messy way to handle that.

stacker/dag/__init__.py
```diff
@@ -0,0 +1,264 @@
...
+        """ Add an edge (dependency) between the specified nodes. """
+        if not graph:
+            graph = self.graph
+        if ind_node not in graph:
+            raise KeyError('node %s does not exist' % ind_node)
+        if dep_node not in graph:
+            raise KeyError('node %s does not exist' % dep_node)
+        test_graph = deepcopy(graph)
+        test_graph[ind_node].add(dep_node)
+        is_valid, message = self.validate(test_graph)
```
phobologic: It seems like the `validate` method should throw the exception itself, and have anything that calls it deal with the results if they want to catch them.

stacker/dag/__init__.py
```diff
@@ -0,0 +1,264 @@
...
+            graph[ind_node].add(dep_node)
+        else:
+            raise DAGValidationError(message)
+
+    def delete_edge(self, ind_node, dep_node, graph=None):
+        """ Delete an edge from the graph. """
+        if not graph:
+            graph = self.graph
+        if dep_node not in graph.get(ind_node, []):
+            raise KeyError('this edge does not exist in graph')
```
phobologic: Can you add more info to this exception?  I actually think his overuse of KeyError, instead of custom Exceptions, is a little sloppy.  In any case, I think it's going to be necessary to at least have the two nodes in the exception, so we have a useful error message.

stacker/dag/__init__.py
```diff
@@ -0,0 +1,264 @@
...
+                    edges.remove(old_task_name)
+                    edges.add(new_task_name)
+
+    def predecessors(self, node, graph=None):
+        """ Returns a list of all predecessors of the given node """
+        if graph is None:
+            graph = self.graph
+        return [key for key in graph if node in graph[key]]
+
+    def downstream(self, node, graph=None):
```
phobologic: `descendants` maybe to match `predecessors` more?

When I was cleaning this up I called them: `ancestors` (recursive parents) and `descendants` (recursive children) - I also had `children` and `parents`, as well as properties for `has_children`, `has_descendants`, etc.  `has_children` and `has_descendants` were both the same call, because you couldn't have descendants if you didn't have children.

phobologic: Oh, sorry, this is what I woul dhave called `children` - `all_downstream` is what I would have called `descendants`

stacker/dag/__init__.py
```diff
@@ -0,0 +1,264 @@
...
+
+        # Add only the nodes we need.
+        for node in nodes:
+            dag.add_node_if_not_exists(node)
+            for edge in self.all_downstreams(node, graph=graph):
+                dag.add_node_if_not_exists(edge)
+
+        # Now, rebuild the graph for each node that's present.
+        for node, edges in graph.items():
+            if node in dag.graph:
```
phobologic: This if seems dangerous, since we silently throw away any nodes that aren't in dag.graph, which really shouldn't ever happen, right?

stacker/dag/__init__.py
```diff
@@ -0,0 +1,264 @@
...
+        """ Reset the graph and build it from the passed dictionary.
+
+        The dictionary takes the form of {node_name: [directed edges]}
+        """
+
+        self.reset_graph()
+        for new_node in graph_dict.iterkeys():
+            self.add_node(new_node)
+        for ind_node, dep_nodes in graph_dict.iteritems():
+            if not isinstance(dep_nodes, list):
```
phobologic: I think you want to compare this against `collections.Iterable`, rather than just list.  A tuple would be perfectly ok  here as well.

stacker/dag/__init__.py
```diff
@@ -0,0 +1,264 @@
...
+                raise TypeError('dict values must be lists')
```
phobologic: Give the node name here, just so you can find the offending node in the case of an error.

stacker/dag/__init__.py
```diff
@@ -0,0 +1,264 @@
...
+            for dep_node in dep_nodes:
+                self.add_edge(ind_node, dep_node)
+
+    def reset_graph(self):
+        """ Restore the graph to an empty state. """
+        self.graph = OrderedDict()
+
+    def ind_nodes(self, graph=None):
```
phobologic: I think this is short for `independent_nodes` - can we name it that instead?

stacker/dag/__init__.py
```diff
@@ -0,0 +1,264 @@
...
+            self.add_node(new_node)
+        for ind_node, dep_nodes in graph_dict.iteritems():
+            if not isinstance(dep_nodes, list):
+                raise TypeError('dict values must be lists')
+            for dep_node in dep_nodes:
+                self.add_edge(ind_node, dep_node)
+
+    def reset_graph(self):
+        """ Restore the graph to an empty state. """
+        self.graph = OrderedDict()
```
phobologic: Is the fact that the graph is a dict going to cause issues with duplicate keys?  Like if someone puts two stack definitions in, could it result in the latter ovewriting the former?

We could still use OrderedDict, just sub-class it and override `__setitem__` so that it checks for duplicates if this is a concern.

stacker/dag/__init__.py
```diff
@@ -0,0 +1,264 @@
...
+
+    def ind_nodes(self, graph=None):
+        """ Returns a list of all nodes in the graph with no dependencies. """
+        if graph is None:
+            graph = self.graph
+
+        dependent_nodes = set(
```
phobologic: I'm having a really hard time reading this - any way to dumb it down that isn't going to destroy performance?

stacker/dag/__init__.py
```diff
@@ -0,0 +1,264 @@
...
+                in_degree[v] -= 1
+                if in_degree[v] == 0:
+                    queue.appendleft(v)
+
+        if len(l) == len(graph):
+            return l
+        else:
+            raise ValueError('graph is not acyclic')
+
+    def size(self):
```
phobologic: You could also give this class a `__len__` method so that calling `len()` on it would return the size.

stacker/exceptions.py
```diff
@@ -217,3 +217,17 @@ def __init__(self, failed_stacks, *args, **kwargs):
...
+
+class GraphError(Exception):
+    """Raised when the graph is invalid (e.g. acyclic dependencies)
+    """
+
+    def __init__(self, exception, stack=None, dependency=None):
+        self.stack = stack
+        self.dependency = dependency
+        self.exception = exception
+        message = ("Error detected when adding '%s' "
```
phobologic: This message will be broken if stack  & dependency aren't set.

stacker/plan2.py
```diff
@@ -0,0 +1,297 @@
...
+        graph.add_step(step)
+
+    for step in steps:
+        for dep in step.requires:
+            graph.connect(step, dep)
+
+    return graph
+
+
+class Graph(object):
```
phobologic: Curious if this should just be combined with the `Plan` class?

stacker/plan2.py
```diff
@@ -0,0 +1,297 @@
...
+
+    @property
+    def step_names(self):
+        return [step.name for step in self.steps]
+
+    def keys(self):
+        return self.step_names
+
+
+class UnlimitedSemaphore(object):
```
phobologic: Is this just planning ahead for multi-process/threading?

ejholmes: Yeah, this is leftover from multi-threading. I'll remove this since it's unnecessary right now.

stacker/plan2.py
ejholmes: I originally started with that, but I think it's cleaner to keep these things separate:

1. `Graph` encapsulates a graph of `Step` objects, and has operations to make working on the graph itself easy (transposition, walking, etc).
2. `Plan` is more like a convenience class for Actions to consume (dump, outline, etc).

`Plan` could probably eventually get removed, as most of the methods on it may be better somewhere else, but I primarily kept it around to keep API compatibility with the existing `Plan` implementation.

ejholmes:
Functional tests are all passing, so the only next steps to get this mergeable should be:

1. Implement `--tail` flag.
2. Implement `--dump` flag.
3. Re-implement the loop logger.

ejholmes:
@phobologic when you wanna review this, this should be 100% feature complete and pretty close to being mergeable.

A lot of your original comments were around the implementation of `py-dag`. I'm not really inclined to change it very much, since it's algorithmically correct and has good test coverage. Would you rather pull in `py-dag` as a third party dependency, or leave it vendored as is?

stacker/exceptions.py
ejholmes: Where this is used, those things are never not present.

phobologic: Cool - you should make them required then, so we don't hit an edge case.

phobologic:
A few of my comments are around things that look dangerous or error prone (duplicate key comment, the comment about ignoring mising edges)- if you think they aren't and can explain it, I'd be fine ignoring those.

Others are just from general usefulness.  Better errors, easier to read code, etc.

I'm fine with skipping the stuff about children/descendants, since I know that's mostly just graph theory terms being used instead.

All in all, I'll take a small performance hit if it means better maintainability in the future.  That said, I don't expect my suggestions/questions to actually result in much of a hit at the scale we're operating and due to the fact that most of the time is spent in talking with AWS.

I think we should go ahead and vendor - though we should give the Author credit at the top of the code most likely.

Ok, now going to review the latest.

stacker/dag/__init__.py
ejholmes: No clue. It seems unnecessary. I'll try to add a commit that removes this.

stacker/tests/test_plan2.py
```diff
@@ -0,0 +1,241 @@
...
+        app = Stack(
+            definition=generate_definition(
+                'app', 1, requires=['namespace-db.1']),
+            context=self.context)
+
+        with self.assertRaises(GraphError) as expected:
+            build_plan(
+                description="Test",
+                steps=[Step(vpc, None), Step(db, None), Step(app, None)])
+        message = ("Error detected when adding 'namespace-db.1' "
```
phobologic: You know, you probably wouldn't need this message test if we had different exceptions ;)  Just saying.

stacker/__init__.py
```diff
@@ -1 +1,25 @@
+from stacker.plan2 import Step, build_plan
+
 __version__ = "1.1.4"
+
+
+def plan(description=None, action=None,
```
phobologic: What's the plan going forward with this? We probably can't name this function "plan" if we move `plan2` to plan.

stacker/dag/__init__.py
phobologic: Thanks, yeah, I couldn't figure it out - I thought it'd be a test thing, but just now looking at tests I don't think it is.  Strange.

stacker/dag/__init__.py
ejholmes: That's what should happen. It's iterating through the full graph of nodes, and only re-adding edges for nodes that exist in the filtered graph.

phobologic: Ahh, my bad - mixing dag & graph in this method made it confusing.  Maybe having it be `original_graph` and `filtered_graph` would make it less confusing.  Not a huge deal for now.

stacker/dag/__init__.py
ejholmes: That shouldn't be possible, since there's two places that check this now:

1. [Config level](https://github.com/remind101/stacker/blob/f07e6f2c6280d45d67e76a7d53434232b4312390/stacker/config/__init__.py#L370-L379)
2. [Graph level](https://github.com/remind101/stacker/blob/f9ba21857a2b7a17d4d58e2a6ab820da9e982fe9/stacker/dag/__init__.py#L28-L29)

I can add a functional test for this though, if we don't already have one.

stacker/__init__.py
ejholmes: The plan (no pun intended) would be to replace `plan.py` with `plan2.py` before merging, unless we want to keep the original `plan.py` around for backwards compatibility. I only went with `plan2.py` because overwriting `plan.py` in the PR would make reviewing it pretty hard, since it's a re-write.

stacker/tests/test_plan2.py
```diff
@@ -0,0 +1,241 @@
...
+        vpc = Stack(
+            definition=generate_definition('vpc', 1),
+            context=self.context)
+        bastion = Stack(
+            definition=generate_definition('bastion', 1, requires=[vpc.fqn]),
+            context=self.context)
+
+        plan = build_plan(description="Test", steps=[Step(vpc), Step(bastion)])
+
+        self.assertEqual(plan.graph.dag.graph, {
```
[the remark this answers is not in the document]
ejholmes: This is really just a bad test on my part. `dag` should be private within the Graph class, but no way to enforce that with python :(.

I'll add a `to_dict` method to Graph, to make testing this cleaner.

stacker/tests/actions/test_destroy.py
```diff
@@ -72,95 +86,3 @@ def test_destroy_stack_complete_if_state_submitted(self):
         # if we have processed the step and then can't find the stack, it means
         # we successfully deleted it
         self.assertEqual(status, COMPLETE)
-
-    def test_destroy_stack_in_parallel(self):
```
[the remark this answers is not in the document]
ejholmes: I don't think so. I think the purpose of the original test here was to test that the logic for ordering was correct, but that's now done at a lower level within the graph.

I should probably bring back the test below this one though (`test_destroy_stack_step_statuses`).

stacker/__init__.py
phobologic: Right, but what about the `plan` method?  If you have a `plan` method in `stacker/__init__.py` and a `stacker/plan` directory, you're going to have conflicts when someone does something like: `from stacker import plan` or `from stacker.plan import *` (maybe not on the last one, I honestly don't know!)

stacker/tests/test_plan2.py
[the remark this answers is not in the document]
phobologic: Cool - I also just realized the name of this test is probably wrong.

stacker/plan2.py
```diff
@@ -0,0 +1,345 @@
...
+        self.set_status(SKIPPED)
+
+    def submit(self):
+        """A shortcut for set_status(SUBMITTED)"""
+        self.set_status(SUBMITTED)
+
+
+def build_plan(description=None, steps=None,
+               step_names=None, reverse=False):
+    graph = build_graph(steps)
```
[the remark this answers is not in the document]
ejholmes: I can't think of a scenario where that could actually happen from the `stacker` CLI, since there's config validation to ensure that stacks isn't empty, but I'll add a test for that here.

stacker/__init__.py
ejholmes: Ah, gotcha. I think this `plan` method should actually be moved to `stacker.actions.base`, since that's where it's intended to be used.

tests/suite.bats
```diff
@@ -629,10 +629,7 @@ EOF
   assert "$status" -eq 1
```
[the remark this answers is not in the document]
ejholmes: In the implementation in this PR, if a node in the graph fails (returns False), the entire graph walk will halt. It's not actually necessary to halt the entire graph walk, only nodes that depend on the failed node. I can update the implementation of the graph walk to support the old behavior.

ejholmes:
Alright @phobologic. I think I've addressed a majority of the comments here.

I also reverted back to the original behavior for failed child stacks, so that the entire graph of stacks is executed, but a given stack will not execute if any of it's dependencies are in a "not ok" state (see cd0c0f6b344444bee45992c1a166769b41c6b660)

ejholmes:
@Lowercases found a bug where `CancelExecution` wasn't handled (pressing `n` in interactive mode). Added a commit that handles this flow, and also an integration test, since we didn't have one.

stacker/plan2.py
```diff
@@ -0,0 +1,345 @@
...
+
+        return self.graph.walk(walk_func)
+
+    def execute(self, **kwargs):
+        return self.walk(**kwargs)
+
+    def walk(self):
+        """Walks each step in the underlying graph, in topological order."""
+
+        def walk_func(step):
```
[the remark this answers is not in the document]
ejholmes: I think we definitely should make that change, since it's annoying when a stack throws an exception, and you have no context about what stack has the issue. But we should probably make that change separate from this PR, since that's not existing behavior.

We'll also need to think about how we handle exceptions in a parallelized world, since the current method of just throwing an exception and aborting won't work with multi-threads/processes.

stacker/plan2.py
```diff
@@ -0,0 +1,400 @@
...
+
+    for step in steps:
+        for dep in step.requires:
+            graph.connect(step, dep)
+
+    return graph
+
+
+class Graph(object):
+    """Graph represents a graph of steps.
```
phobologic: This is super helpful - thanks!

ejholmes:
Cool. I'll update us internally to run this branch before merging.

ejholmes:
There's some talk about whether or not the `--stacks` flag should stop at the given stack, or continue to execute stacks that transitively depend on the provided stack.

Let's say you have a graph like the following:



In the current behavior, if you specify `--stacks db`, stacker will effectively filter the graph to the following:



If there's a change to the `db` stack (like a new instance url or w/e), then the `app` stack won't see it.

We could change the behavior here so that the graph includes the dependencies of `db`, as well as anything that transitively depends on it. So the graph would be filtered to:

acmcelwee:
Personally, the expected behavior of the `--stacks` flag is as you described it in the first example, and how it's currently implemented in the PR. That said, I think the second behavior could be useful in other occasions. It might be worth thinking about how both could coexist with different flags or modes.

russellballestrini:
The version is this PR is `faster` but I'm pretty sure the version in the 3rd chart is `safer` (and puts the system as a whole in a more correct state).

I would prefer correctness over speed.

I like @acmcelwee point, if we can support both, (`--stacks-fast`) I think that could be really cool.

ejholmes:
Yeah, that makes sense to me. I think I'll update this PR to the more complete version.

If there's a need for the old behavior, we can add new flags for that.

ejholmes:
Aaaaand...I think I'm gonna flipflop my preference back to the fast version. I tried the complete version internally, and it's pretty easy for the graph to explode making `--stacks` mostly useless.

I'll leave out specific stack names, but on our internal graph, when I targeted a stack that builds out an ECS cluster, the fast versions results in 4 stacks (3 dependencies). With the full version, the graph explodes to 100 stacks as a result of the parents transitive dependencies.

ejholmes:
Thanks everyone for the comments and feedback. Super excited about what this is gonna enable for stacker moving forward.

PR #528, opened 2018-02-03
Always dump the current output/result

phobologic:
This fixes an issue that was introduced in #467. The original
functionality was to always dump the current output to `.json-result`,
that way it was easy to build & compare the output.

With the current implementation (from #467) we raise an exception if the
old results aren't available, but don't create the new results.  This is
an issue whenever creating a new test.

@danielkza can you verify that this doesn't break anything that you were trying to fix in #467?  I can't see how it would.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Upcoming/Master
 
+- assertRenderedBlueprint always dumps current results [GH-528]
 
 ## 1.1.4 (2018-01-26)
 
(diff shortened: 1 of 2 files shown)


danielkza:
> can you verify that this doesn't break anything that you were trying to fix in #467? I can't see how it would.

I can't see how it would either, and getting more information for debugging is always nice.

phobologic:
Thanks @danielkza !

PR #529, opened 2018-02-07
Remove DescribeStacks call from prepare_stack_for_update

ejholmes:
This is a small performance optimization to remove an unnecessary DescribeStacks call from the `prepare_stack_for_update` method.

Before this change, the outbound network calls for a single noop stack update would look like this:

1. [DescribeStacks](https://github.com/remind101/stacker/blob/044ca52e2453dbddcc54d4a72714962bb04842f6/stacker/actions/build.py#L217): Check if stack is created or not, and to build a list of parameters
2. [DescribeStacks](https://github.com/remind101/stacker/blob/044ca52e2453dbddcc54d4a72714962bb04842f6/stacker/actions/build.py#L263) * N: Resolve any output lookups from other stacks
3. [HeadObject](https://github.com/remind101/stacker/blob/044ca52e2453dbddcc54d4a72714962bb04842f6/stacker/actions/base.py#L141): Check if the template has already been uploaded.
4. [DescribeStacks](https://github.com/remind101/stacker/blob/044ca52e2453dbddcc54d4a72714962bb04842f6/stacker/actions/build.py#L285): Check if the stack is in a rollback_complete state, and needs to be deleted)
5. [UpdateStack](https://github.com/remind101/stacker/blob/044ca52e2453dbddcc54d4a72714962bb04842f6/stacker/actions/build.py#L287): Perform the update.

The 3rd DescribeStacks (step 4) is unnecessary, and can be easily removed, since we already have a DescribeStacks response from the first call.

### DIFF
--- a/stacker/providers/aws/default.py
+++ b/stacker/providers/aws/default.py
@@ -694,7 +694,7 @@ class Provider(BaseProvider):
         else:
             return self.default_update_stack
 
-    def prepare_stack_for_update(self, fqn, tags):
+    def prepare_stack_for_update(self, stack, tags):
         """Prepare a stack for updating
 
         It may involve deleting the stack if is has failed it's initial
@@ -705,7 +705,7 @@ class Provider(BaseProvider):
             enabled by the user, or because interactive mode is on.
 
         Args:
-            fqn (str): fully-qualified name of the stack to work on
+            stack (dict): a stack object returned from get_stack
             tags (list): list of expected tags that must be present in the
                 stack if it must be re-created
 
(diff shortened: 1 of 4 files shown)


ejholmes:
@danielkza since you originally added `prepare_stack_for_update` in https://github.com/remind101/stacker/pull/473, do you mind sanity checking this as well?

stacker/providers/aws/default.py
```diff
@@ -694,7 +694,7 @@ def select_update_method(self, force_interactive, force_change_set):
         else:
             return self.default_update_stack
 
-    def prepare_stack_for_update(self, fqn, tags):
```
phobologic: Need to update the docstring.

ejholmes:
> have you ran the integration tests?

Yep, integration tests pass.

PR #530, opened 2018-02-07
add raw json/yaml template support

### LINKED ISSUE #444: Support for Native CloudFormation Templates
We tried Stacker and the idea is great and well executed. The problem is that we have a bunch of stacks already written in Troposphere that output JSON/YAML CF templates. What would be the possibility of adding native support for these? It wouldn't be difficult to import the Stacker package(s) into the Troposphere templates and call something like 

```
stacker.create_blueprint(template.to_json())
```

Or something similar. You could even wrap the Troposphere template in a class and call that class from Stacker.

troyready:
Fixes #444


troyready:
@phobologic @ejholmes I'm mildly surprised at how easy this ended up being; I've tested it with a couple of yaml & json templates without issue.

Could use some help figuring out how to use schematics to model the option to specify exactly one of `class_path` or `template_path`. I looked at the Model docs a bit but it didn't jump out at me -- curious if one of you knows or could figure it out faster than I can?

docs/config.rst
```diff
@@ -315,7 +315,11 @@ A stack has the following keys:
   The base name for the stack (note: the namespace from the environment
   will be prepended to this)
 **class_path:**
-  The python class path to the Blueprint to be used.
+  The python class path to the Blueprint to be used. Specify this or
+  ``template_path`` for the stack.
+**template_path:**
```
ejholmes: I kinda feel like this could all be done without even adding a new config option. The `RawTemplateBlueprint` could just use a variable to specify the template path.

For example:

```yaml
stacks:
- name: vpc
  class_path: stacker.blueprints.RawTemplateBlueprint
  variables:
    template_path: templates/vpc.yaml
```

Whether or not that's better is probably debatable, but it keeps stacker config's more generic, which is good.

troyready: I think it's worth of the pain of class_path vs template_path to avoid any other config values on where the template is.

troyready:
@ejholmes thanks for the feedback; you're right, it's much nicer fixing the blueprint to do the right thing.

I've gone through and tested build/destroy/diff now and think this is in a good place.

stacker/util.py
```diff
@@ -493,6 +493,48 @@ def get_client_region(client):
     return client._client_config.region_name
 
 
+def get_template_file_format(template_path):
```
ejholmes: Can we move this methods to `stacker.blueprints.raw`, since they're only used there? At the very least, it'll make reading `RawTemplateBlueprint` easier.

stacker/blueprints/raw.py
```diff
@@ -0,0 +1,147 @@
...
+        super(RawTemplateBlueprint, self).__init__(name,
+                                                   context,
+                                                   mappings,
+                                                   description)
+        self._raw_template_path = raw_template_path
+        self._raw_template_format = get_template_file_format(raw_template_path)
+
+    def render_template(self):
+        """Load template and generate its md5 hash."""
+        with open(self.raw_template_path, 'r') as template:
```
ejholmes: We're still reading the template twice. Once here, and again in `util.get_template_params`. We should only need to read the template from disk once into memory here, and then use the in memory contents to get params.

stacker/blueprints/raw.py
```diff
@@ -0,0 +1,147 @@
...
+            rendered = template.read()
+        version = hashlib.md5(rendered).hexdigest()[:8]
+        parsed_template = yaml.load(rendered)
+        if 'Transform' in parsed_template:
+            self.template.add_transform(parsed_template['Transform'])
```
ejholmes: Why do we do this? What significance does the troposphere Template play in this blueprint?

stacker/util.py
```diff
@@ -493,6 +493,48 @@ def get_client_region(client):
...
+    Args:
+        template_path (str): Path to the template on disk.
+
+    Returns:
+        dict: Template parameters.
+    """
+    params = {}
+
+    if not os.path.isfile(template_path):
+        raise EnvironmentError("Could not find CFN template at path "
```
ejholmes: Do we have a test for this? It seems like this isn't a defined exception:

```console
$ git grep EnvironmentError'
```

troyready:
@ejholmes good thoughts, thanks. I think I've addressed each comment.

stacker/blueprints/raw.py
```diff
@@ -0,0 +1,147 @@
...
+        # in the definition
+        try:
+            value = var_def["Default"]
+        except KeyError:
+            raise MissingVariable(blueprint_name, var_name)
+
+    return value
+
+
+class RawTemplateBlueprint(Blueprint):  # pylint: disable=abstract-method
```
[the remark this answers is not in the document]
troyready: Yeah, agreed that it doesn't seem to re-use much. I like it having a override/subset relationship, but I could be persuaded otherwise.

ejholmes:
@troyready I opened a PR against yours with a few small changes, and some more tests: https://github.com/troyready/stacker/pull/1.

troyready:
Thanks! Merged and rebased.

ejholmes:
I'll let @phobologic give the final 👍. Otherwise, this looks awesome to me.

troyready:
I've confirmed this can be rebased into a standalone change on v1.1.4 for a non-DAG release without issue.

So, if it makes sense to delay a DAG release until #531 , then a v1_2 release branch (branched off 1.1.4) could cherry-pick 1d222ec64d to get this branch's changes only.

PR #531, opened 2018-02-07
Parallel graph walk

### LINKED ISSUE #279: General slowness

### LINKED ISSUE #357: Parallelize resolution of lookups
These happen serially right now, and if you have a lot of variables that use looks that make external requests (kms, output, etc) this can effect performance significantly.

ejholmes:
Fixes #279
Closes #357

Now that the [DAG](https://github.com/remind101/stacker/pull/523) has been merged, the single biggest performance improvement we can make to `stacker build` is to switch the graph walk to a parallel walk.

This PR isn't quite ready to merge (but works well, I've used it internally to make changes), so I'm primarily opening this up to start talking about the implementation, and possible changes we need to make. I don't think that we should include this in the `1.2` release of stacker, so it gives us time to polish this and test it internally first.

This is a multi-threaded implementation of the graph walk, which will walk the graph as fast as the graph allows. I think multi-threading is ultimately easier to do than multi-processing, since there's actually not very much we need to make thread safe within stacker itself, because of the nature of the graph. There was some talk in the past about using async io, which would be more resource efficient, but I think would complicate the implementation, but if someone wants to give that a try, be my guest.

Note that, I'm basing this branch on another branch that removes the loop logger, and moves to a simple sequential logger. It makes parallelism a lot easier to deal with.

### Perf

I tested this against our internal stacker config (153 stacks) and it drops execution time from ~8 minutes to ~2 minutes. I think there's still room for a lot of optimization here.

**before**

```console
$ time stacker build
real    8m32.462s
```

**after**

```console
$ time stacker build
real    2m23.987s
```

FWIW, I have not yet run into throttling on DescribeStacks after https://github.com/remind101/stacker/pull/529 (which this PR includes) and the change to `set_outputs`.

### Prerequsites

* https://github.com/remind101/stacker/pull/532

### TODO

* [x] We should add something like a `--max-parallel` flag to specify the maximum allowed parallelism. This would just control a semaphore that wraps a `build`/`destroy`.
* [x] In interactive mode, we'd need a lock around asking for approval.
* [ ] Think about thread safety more.
* [x] Handle SIGTERM/SIGINT gracefully.


stacker/actions/build.py
```diff
@@ -255,6 +255,7 @@ def _launch_stack(self, stack, **kwargs):
 
                 return FailedStatus(reason)
             elif self.provider.is_stack_completed(provider_stack):
+                self.provider.set_outputs(stack.fqn, provider_stack)
```
ejholmes: This is a simple change to ensure that output lookups are parallelized. I'm not sure this is really the best implementation for this, which is why there's no comment explaining this yet.

In a nutshell, when a stack is "UPDATE_COMPLETE", this will store the outputs for the stack on the provider. When another stack references an output from a dependency, this ensures that there's no sequential blocking DescribeStacks calls to do the output lookups from dependencies, and also makes outputs thread safe.

stacker/dag/__init__.py
```diff
@@ -364,3 +365,92 @@ def size(self):
...
+                    "%s waiting for %s to complete",
+                    n,
+                    ", ".join(deps))
+
+            # Wait for all dependencies to complete sucessfully.
+            if not wait_for(deps):
+                logger.debug("cancelling %s. "
+                             "Some dependencies "
+                             "were not satisfied", n)
+                return False
```
ejholmes: Note to self. This should just continue to walk to the graph, instead of stopping. Conditional execution of the node is handled [further up the call stack](https://github.com/remind101/stacker/blob/044ca52e2453dbddcc54d4a72714962bb04842f6/stacker/plan.py#L349-L352), which is a better place to abort.

ejholmes:
Alright, this is getting awesome. I've added a few commits:

1. Added a `-j/--max-parallel` flag to control concurrency.
2. Interactive mode is handled. So if two stacks request changes, only 1 thread can ask the user for input at a time.
3. SIGTERM/SIGINT are handled gracefully. Stacker will finish whatever it's currently working on, then exit.

Here's an example of two stacks request changes in interactive mode:

[asciicast](https://asciinema.org/a/tFzkoM871ezXwYITNXyYtPe00)

And an example of ^C (SIGINT/SIGTERM):

[asciicast](https://asciinema.org/a/XvLkmoVOJV50e8mpSyivonam9)

I'll get us dogfooding this internally, but I think implementation wise, this is looking pretty solid.

stacker/commands/stacker/base.py
```diff
@@ -1,10 +1,32 @@
...
+
+    def cancel_execution(signum, frame):
+        cancel.set()
+
+    signal.signal(signal.SIGINT, cancel_execution)
+    signal.signal(signal.SIGTERM, cancel_execution)
+    return cancel
+
+
+def build_semaphore(concurrency):
```
ejholmes: Might be cleaner to change this to `build_walker` and return an object for walking the graph (`stacker.dag.walk`/`stacker.dag.walk_threaded`), with the semaphore built-in.

This way, when `--max-parallel=1` we can disable threaded execution entirely, which may be useful for debugging in some cases.

stacker/commands/stacker/build.py
```diff
@@ -31,6 +31,12 @@ def add_arguments(self, parser):
                                  "dependencies. Can be specified more than "
                                  "once. If not specified then stacker will "
                                  "work on all stacks in the config file.")
+        parser.add_argument("-j", "--max-parallel", action="store", type=int,
+                            default=0,
+                            help="The maximum number of stacks to execute in "
```
ejholmes: Should expand the docs on this.

ejholmes:
A couple people here have hit throttling while using this. I think there's 2 things that can be done:

1. [Better exponential backoff on API calls](https://github.com/remind101/stacker/pull/533)
2. Less aggressive polling during updates.

ejholmes:
This is working well for us internally. I'm going to merge this into master and plan on doing a release candidate tomorrow to get more external people testing it.

PR #536, opened 2018-02-18
avoid loading GitPython unnecessarily

troyready:
this should eliminate the need for setting the [GIT_PYTHON_REFRESH environment variable](https://github.com/gitpython-developers/GitPython/blob/2.1.8/git/cmd.py#L236-L288) on systems without git installed


phobologic:
Thanks @troyready

PR #538, opened 2018-02-19
fix loading of yaml templates with CFN tags

troyready:
Ref!/Sub!/etc will no longer fail to load

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,7 +6,6 @@ VERSION = "1.2.0rc1"
 src_dir = os.path.dirname(__file__)
 
 install_requires = [
-    "awscli>=1.11.37",
     "troposphere>=1.9.0",
     "botocore>=1.6.0",
     "boto3>=1.3.1,<1.5.0",
--- a/setup.py
+++ b/setup.py
@@ -6,6 +6,7 @@ VERSION = "1.2.0rc1"
 src_dir = os.path.dirname(__file__)
 
 install_requires = [
+    "awscli>=1.11.37",
     "troposphere>=1.9.0",
     "botocore>=1.6.0",
     "boto3>=1.3.1,<1.5.0",
--- a/stacker/actions/diff.py
+++ b/stacker/actions/diff.py
@@ -4,11 +4,10 @@ import logging
 import sys
 from operator import attrgetter
 
-from awscli.customizations.cloudformation.yamlhelper import yaml_parse
-
 from .base import plan, build_walker
 from . import build
 from .. import exceptions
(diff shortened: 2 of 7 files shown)


troyready:
The test is failing on `error: botocore 1.8.44 is installed but botocore==1.8.45 is required by set(['awscli'])`, but that new explicit version dependency doesn't conflict with stacker's `botocore>=1.6.0`

troyready:
Another option here would be to use cfn_flip, which is [already pulled in by troposphere](https://github.com/cloudtools/troposphere/commit/0ed3fe09c7ad736b70750d1644fae7300f2b8f94).

I didn't like its helper as much though because it's just a yaml loader which ends up making an object with a bunch of OrderedDicts . Not sure if adding some translation on top of that (instead of just using the awscli helper) is better.

setup.py
```diff
@@ -6,6 +6,7 @@
 src_dir = os.path.dirname(__file__)
 
 install_requires = [
+    "awscli>=1.11.37",
```
ejholmes: This is a pretty heavy dependency for what amounts to about [80 LOC](https://github.com/aws/aws-cli/blob/10cbd8d6aec62a0e02e3e75d9624559a703fca4f/awscli/customizations/cloudformation/yamlhelper.py).

I think I'd rather just vendor that in stacker, rather than adding a dependency on the AWS CLI.

phobologic:
I was thinking that as well @troyready - I'm honestly more hesitant with what you have here just because of the reliance on awscli - which I think folks might conflict with quite a bit.

troyready:
I definitely like the idea of using cfn_flip more, but I'm not sure how to implement it in a way that won't be kinda gross.

Accordingly, I just did the vendoring.

stacker/actions/diff.py
```diff
@@ -4,11 +4,10 @@
 import sys
 from operator import attrgetter
 
-import yaml
-
 from .base import plan, build_walker
 from . import build
 from .. import exceptions
+from ..awscli_yamlhelper import yaml_parse
```
ejholmes: Could we put this in `cloudformation` module and name the method `parse_cloudformation_template` or something along the lines of that? I realize this mostly comes from how the AWS CLI named it, but I think we can name it a little better.

```python
from ..cloudformation import parse_cloudformation_template
```

troyready: I think a cloudformation module or a function in util.py does make sense. Might be nice to have it wrap the vendored awscli_yamlhelper module to keep the vendored file as distinct/pristine as possible?

troyready:
Added the wrapper function

PR #541, opened 2018-02-23
fix diff with local yaml templates

troyready:
#540 fixed the generally broken diffs in #530 , but introduced a problem with yaml templates.

### DIFF
--- a/stacker/actions/diff.py
+++ b/stacker/actions/diff.py
@@ -143,9 +143,27 @@ def diff_parameters(old_params, new_params):
     return diff
 
 
+def normalize_json(template):
+    """Normalize our template for diffing.
+
+    Args:
+        template(str): string representing the template
+
+    Returns:
+        list: json representation of the parameters
+    """
+    obj = parse_cloudformation_template(template)
+    json_str = json.dumps(obj, sort_keys=True, indent=4)
+    result = []
+    lines = json_str.split("\n")
+    for line in lines:
+        result.append(line + "\n")
+    return result
+
+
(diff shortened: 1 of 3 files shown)


stacker/tests/actions/test_diff.py
```diff
@@ -81,3 +84,61 @@ def test_diff_parameters_no_changes(self):
...
+            '            }, \n',
+            '            "Type": "AWS::S3::Bucket"\n',
+            '        }, \n',
+            '        "Dummy": {\n',
+            '            "Type": "AWS::CloudFormation::WaitConditionHandle"\n',
+            '        }\n',
+            '    }\n',
+            '}\n'
+        ]
+        self.assertEquals(normalized_template, diff._normalize_json(template))
```
ejholmes: We should just pull this off of the `Diff` class and into a function. Easier to test since you don't need to initialize anything.

troyready:
@ejholmes good thought; helped me catch another error in the display logic. I also moved the normalize_json function out as requested.

stacker/actions/diff.py
```diff
@@ -156,9 +174,10 @@ def print_stack_changes(stack_name, new_stack, old_stack, new_params,
     template_changes = list(lines)
     if not template_changes:
         print "*** No changes to template ***"
-    else:
-        param_diffs = diff_parameters(old_params, new_params)
+    param_diffs = diff_parameters(old_params, new_params)
+    if param_diffs:
```
ejholmes: Ah, I noticed this bug recently as well. Thanks for fixing!

PR #543, opened 2018-02-28
Improved signal handling

ejholmes:
This has a few small improvements to handle SIGTERM/SIGINT signals better:

* Fixes a bug where the SIGTERM/SIGINT signal handler wasn't being immediately handled. See https://stackoverflow.com/questions/25676835/signal-handling-in-multi-threaded-python
* Logs a line when SIGTERM/SIGINT are handled.
* Logs a line at the end of the plan execution to show a summary of stacks that failed (same behavior as what we have in 1.1).

/cc @GarisonLotus @troyready

### DIFF
--- a/tests/suite.bats
+++ b/tests/suite.bats
@@ -673,6 +673,7 @@ EOF
   assert_has_line "${STACKER_NAMESPACE}-dependent-rollback-parent: submitted (rolling back new stack)"
   assert_has_line "${STACKER_NAMESPACE}-dependent-rollback-parent: failed (rolled back new stack)"
   assert_has_line "${STACKER_NAMESPACE}-dependent-rollback-child: failed (dependency has failed)"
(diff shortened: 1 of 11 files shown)


tests/suite.bats
```diff
@@ -673,6 +673,7 @@ EOF
   assert_has_line "${STACKER_NAMESPACE}-dependent-rollback-parent: submitted (rolling back new stack)"
   assert_has_line "${STACKER_NAMESPACE}-dependent-rollback-parent: failed (rolled back new stack)"
   assert_has_line "${STACKER_NAMESPACE}-dependent-rollback-child: failed (dependency has failed)"
+  assert_has_line "The following stacks failed: dependent-rollback-parent, dependent-rollback-child"
```
ejholmes: This is actually just a revert to how 1.1 behaves: https://github.com/remind101/stacker/blob/1.1.4/tests/suite.bats#L637

Issue #546, opened 2018-03-01
AWS parameter type name instead of CFNType


ejholmes:
Right now, if you want to specify a variable as a CloudFormation parameter, you would import the appropriate CFNType, then use that as the type field. For example:

```python
from stacker.blueprints.variables.types import EC2VPCId

VARIABLES = {
    "VpcId": {
        "type": EC2VPCId
    }
}
```

This works fine, but I think we could make a little simpler and easier to use. If instead, you could do the following:

```python
VARIABLES = {
    "VpcId": {
        "type": "AWS::EC2::VPC::Id"
    }
}
```

And stacker would internally convert it to a `EC2VPCId`. This behavior would be pretty similar to what we had in 0.8.

PR #547, opened 2018-03-01
Add support for specifying variable types using named parameter types

ejholmes:
Closes https://github.com/remind101/stacker/issues/546

This just makes it a little easier to use CFNType's, by just specifying the name of the type you want for the parameter. The following now works:

```python
VARIABLES = {
    "VpcId": {
        "type": "AWS::EC2::VPC::Id"
    }   
}
```

And it'll be automatically normalized to:

```python
VARIABLES = {
    "VpcId": {
        "type": EC2VPCId
    }
}
```

### DIFF
--- a/stacker/blueprints/base.py
+++ b/stacker/blueprints/base.py
@@ -40,6 +40,20 @@ PARAMETER_PROPERTIES = {
 }
 
 
+def normalize_defined_variables(variables):
+    """Looks for any variable definitions that are a "str", and assumes that
+    they're the name of a CloudFormation parameter type, in which case the type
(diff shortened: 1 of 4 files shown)


stacker/blueprints/base.py
```diff
@@ -40,6 +40,20 @@
 }
 
 
+def normalize_defined_variables(variables):
+    """Looks for any variable definitions that are a "str", and assumes that
+    they're the name of a CloudFormation parameter type, in which case the type
+    is converted to a CFNType.
+
```
phobologic: Needs `Args`

PR #550, opened 2018-03-07
Refactor output lookup

ejholmes:
This refactors the mechanics of the `${output}` lookup so that it uses the internal `stacker.Stack` as the lookup point.

With the addition of the DAG, in order to ensure that output lookups didn't involve unnecessary network calls, we changed `stacker.actions.build` to call `provider.set_outputs` to store the stacks outputs when the step was completed.

https://github.com/remind101/stacker/blob/c1aad4f8713c64c90abc02e6c10ad2eda1a18c8e/stacker/actions/build.py#L262-L264

 The output lookup would then use this cached value, through `provider.get_output`.

https://github.com/remind101/stacker/blob/c1aad4f8713c64c90abc02e6c10ad2eda1a18c8e/stacker/lookups/handlers/output.py#L44-L49

This was the smallest change that we could make in the DAG branch, while keeping everything fast and functional, but won't work well moving forward as we think about cross account/region support, as well as the possibility of adding new node types to the graph (e.g. hooks).

---

In this PR, when a stack is "done", we store the outputs on the internal `stacker.Stack` object. The `output` plugin looks up the stack object, then returns the output from there. This has a couple of advantages:

1. It ensures that we're not keeping global state for cached lookups, so we don't have a potential for a race.
2. It'll work better when we think about cross account/region stacks using profiles. You can think of a scenario where, stacker is managing two stacks, both named "RoleModel" in separate accounts. The `output` lookup needs to be aware that these nodes in the graph are separate things.

### DIFF
--- a/stacker/context.py
+++ b/stacker/context.py
@@ -129,20 +129,27 @@ class Context(object):
             list: a list of :class:`stacker.stack.Stack` objects
 
         """
-        stacks = []
-        definitions = self._get_stack_definitions()
-        for stack_def in definitions:
-            stack = Stack(
-                definition=stack_def,
-                context=self,
-                mappings=self.mappings,
-                force=stack_def.name in self.force_stacks,
-                locked=stack_def.locked,
-                enabled=stack_def.enabled,
-                protected=stack_def.protected,
-            )
-            stacks.append(stack)
-        return stacks
+        if not hasattr(self, "_stacks"):
+            stacks = []
+            definitions = self._get_stack_definitions()
+            for stack_def in definitions:
+                stack = Stack(
+                    definition=stack_def,
+                    context=self,
+                    mappings=self.mappings,
+                    force=stack_def.name in self.force_stacks,
+                    locked=stack_def.locked,
+                    enabled=stack_def.enabled,
+                    protected=stack_def.protected,
+                )
+                stacks.append(stack)
+            self._stacks = stacks
+        return self._stacks
+
+    def get_stack(self, name):
+        for stack in self.get_stacks():
+            if stack.name == name:
+                return stack
 
     def get_stacks_dict(self):
         return dict((stack.fqn, stack) for stack in self.get_stacks())
--- a/stacker/lookups/handlers/output.py
+++ b/stacker/lookups/handlers/output.py
@@ -5,48 +5,25 @@ TYPE_NAME = "output"
 Output = namedtuple("Output", ("stack_name", "output_name"))
 
 
-def handler(value, provider=None, context=None, fqn=False, rfqn=False,
-            **kwargs):
+def handler(value, context=None, **kwargs):
     """Fetch an output from the designated stack.
 
     Args:
         value (str): string with the following format:
             <stack_name>::<output_name>, ie. some-stack::SomeOutput
-        provider (:class:`stacker.provider.base.BaseProvider`): subclass of the
-            base provider
         context (:class:`stacker.context.Context`): stacker context
-        fqn (bool): boolean for whether or not the
-            :class:`stacker.context.Context` should resolve the `fqn` of the
-            stack.
-        rfqn (bool): boolean for whether or not the
-            :class:`stacker.context.Context` should resolve the `fqn` of the
-            stack prefixed by the namespace variable
 
     Returns:
         str: output from the specified stack
 
     """
 
-    if rfqn:
-            value = "%s%s%s" % (
-                    context.namespace,
-                    context.namespace_delimiter,
-                    value
-            )
-
-    if provider is None:
-        raise ValueError('Provider is required')
     if context is None:
         raise ValueError('Context is required')
 
     d = deconstruct(value)
-
-    stack_fqn = d.stack_name
-    if not fqn:
-        stack_fqn = context.get_fqn(d.stack_name)
-
-    output = provider.get_output(stack_fqn, d.output_name)
-    return output
+    stack = context.get_stack(d.stack_name)
+    return stack.outputs[d.output_name]
 
 
 def deconstruct(value):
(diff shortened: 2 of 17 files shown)


stacker/context.py
```diff
@@ -129,22 +129,28 @@ def get_stacks(self):
...
+                    mappings=self.mappings,
+                    force=stack_def.name in self.force_stacks,
+                    locked=stack_def.locked,
+                    enabled=stack_def.enabled,
+                    protected=stack_def.protected,
+                )
+                self._stacks.append(stack)
+        return self._stacks
+
+    def get_stack(self, name):
```
ejholmes: Needs docs.

Also, if you're wondering why this doesn't use `get_stacks_dict`, that function indexes by `stack.fqn`, which we don't want. We want to index by the name of the node in the graph.

stacker/lookups/handlers/rxref.py
```diff
@@ -11,13 +11,31 @@
...
+from .output import deconstruct
 
 TYPE_NAME = "rxref"
 
-# rxref is the same as the `output` handler, except the value already contains
-# the relative fully qualified name for the stack we're fetching the output
-# from.
-handler = partial(output_handler, rfqn=True)
+
+def handler(value, provider=None, context=None, **kwargs):
```
ejholmes: Since the `xref` and `rxref` lookups can't re-use the `output` lookup, I just went ahead and re-implemented them to be more explicit, since they were pretty hard to follow before with all of the function currying.

stacker/lookups/handlers/rxref.py
```diff
@@ -11,13 +11,31 @@
...
+        str: output from the specified stack
+    """
+
+    if provider is None:
+        raise ValueError('Provider is required')
+    if context is None:
+        raise ValueError('Context is required')
+
+    d = deconstruct(value)
+    stack_fqn = context.get_fqn(d.stack_name)
```
ejholmes: The original `rxref` implementation does the following, instead of `context.get_fqn`:

https://github.com/remind101/stacker/blob/c1aad4f8713c64c90abc02e6c10ad2eda1a18c8e/stacker/lookups/handlers/output.py#L30-L35

I'm not sure I understand the original reasoning for that, since it looks identical to what `context.get_fqn` does.

stacker/stack.py
```diff
@@ -71,14 +71,14 @@ def __init__(self, definition, context, variables=None, mappings=None,
...
+        self.outputs = None
 
     def __repr__(self):
         return self.fqn
 
     @property
     def requires(self):
-        requires = set([self.context.get_fqn(r) for r in
-                        self.definition.requires or []])
+        requires = set(self.definition.requires or [])
```
ejholmes: Technically this is a breaking change, but one that I doubt would effect very many people. `requires` would no longer support specifying an `fqn`, instead it would only support using the node name in the graph (same as how the `output` lookup works).

stacker/lookups/handlers/output.py
```diff
@@ -5,48 +5,25 @@
...
 
     d = deconstruct(value)
-
-    stack_fqn = d.stack_name
-    if not fqn:
-        stack_fqn = context.get_fqn(d.stack_name)
-
-    output = provider.get_output(stack_fqn, d.output_name)
-    return output
+    stack = context.get_stack(d.stack_name)
```
ejholmes: Because I figure this question will get asked, "What happens if `get_stack` returns None, or outputs haven't been set?":

The graph ensures that that scenario can't happen (remember, the graph is being built out based on the `output` lookups). If for some reason it _did_ happen, then the step would just blow up, and any downstream steps would be cancelled.

stacker/context.py
```diff
@@ -129,22 +129,29 @@ def get_stacks(self):
...
+                stacks.append(stack)
+            self._stacks = stacks
+        return self._stacks
+
+    def get_stack(self, name):
+        for stack in self.get_stacks():
+            if stack.name == name:
+                return stack
+
+    def get_stacks_dict(self, fqn=True):
```
phobologic: It doesn't look like the `fqn` argument is used?

stacker/providers/aws/default.py
```diff
@@ -895,12 +895,11 @@ def get_stack_tags(self, stack, **kwargs):
...
         if stack_name not in self._outputs:
             stack = self.get_stack(stack_name)
-            self.set_outputs(stack_name, stack)
+            self._outputs[stack_name] = get_output_dict(stack)
         return self._outputs[stack_name]
 
-    def set_outputs(self, stack_name, stack):
-        self._outputs[stack_name] = get_output_dict(stack)
-        return
+    def get_output_dict(self, stack):
```
phobologic: Is this used anywhere?

stacker/stack.py
phobologic: Can you run this by the folks in the stacker channel, just to make sure we aren't breaking anything for anyone we know of?

stacker/context.py
ejholmes: Good catch.

stacker/providers/aws/default.py
ejholmes: Yeah, this gets used in the actions to strip outputs from `provider_stack`. I'd rather `providers` just return something like a `stacker.providers.base.Stack` object instead, but that's a bigger refactor.

ejholmes:
> Can you run this by the folks in the stacker channel, just to make sure we aren't breaking anything for anyone we know of?

I ran it poll and nobody said they were using this. Makes sense; just using `requires` is pretty unlikely in itself, and someone going through the effort to do `requires: [${namespace}-stack-name]` is probably even more unlikely.

acmcelwee:
@ejholmes re:
> just using requires is pretty unlikely in itself

FWIW, we mostly uses requires + native export/import over the output/param pattern because of the strong guarantees we get from pushing the dependency information into the CFN domain.

ejholmes:
> FWIW, we mostly uses requires + native export/import over the output/param pattern because of the strong guarantees we get from pushing the dependency information into the CFN domain.

Cool, good to know. We don't have any plans to get rid of `requires`, so you're still good as long as you're using the "short name" (the name of the stack without the `namespace` added to it).

PR #553, opened 2018-03-10
Cross account support

ejholmes:
Depends on https://github.com/remind101/stacker/pull/551

This is based on the RFC in https://github.com/remind101/stacker/wiki/RFC:-Profiles

Closes https://github.com/remind101/stacker/issues/263
Fixes https://github.com/remind101/stacker/issues/277

With this change, you can specify a boto3 profile to use for a given stack. This can be used for cross account provisioning and linking of stacks. For example, say you wanted to provision a hot and cold version of an application in multiple regions and multiple accounts, you could do something like this:

```yaml
stacks:
- name: hot/vpc
  stack_name: vpc
  profile: hot
- name: cold/vpc
  stack_name: vpc
  profile: cold
- name: app
  profile: hot
  variables:
    VpcId: ${output hot/vpc::VpcId}
- name: cold/app
  stack_name: app
  profile: cold
  variables:
    VpcId: ${output cold/vpc::VpcId}
```

```console
$ echo <<EOF > .aws/config
[profile hot]
region = us-east-1
role_arn = arn:aws:iam::1234567:role/Stacker

[profile prod/cold]
region = us-west-1
role_arn = arn:aws:iam::7654321:role/Stacker
EOF
```

```console
$ AWS_CONFIG_FILE=.aws/config.prod stacker build stacker.yaml
```

## TODO

* [x] Docs around using AWS profiles with stacker.
* [x] Functional tests

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -360,7 +360,12 @@ A stack has the following keys:
   (optional): If provided, specifies the name of the region that the
   CloudFormation stack should reside in. If not provided, the default region
   will be used (``AWS_DEFAULT_REGION``, ``~/.aws/config`` or the ``--region``
-  flag).
+  flag). If both ``region`` and ``profile`` are specified, the value here takes
+  precedence over the value in the profile.
+**profile**:
+  (optional): If provided, specifies the name of a AWS profile to use when
+  performing AWS API calls for this stack. This can be used to provision stacks
+  in multiple accounts or regions.
 
 Here's an example from stacker_blueprints_, used to create a VPC::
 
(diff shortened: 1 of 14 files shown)


docs/config.rst
```diff
@@ -359,6 +359,10 @@ A stack has the following keys:
 **region**:
   (optional): If provided, specifies the name of the region that the
   CloudFormation stack should reside in.
+**profile**:
```
phobologic: What happens if this is provided, and region is provided?  Pretty sure you can provide a region as part of a profile, so curious which wins out (and it should probably be documented)

stacker/session_cache.py
```diff
@@ -1,11 +1,21 @@
...
+def getpass(*args):
+    """Wraps getpass to lock the UI."""
+    try:
+        ui.lock()
+        return getpass_orig(*args)
+    finally:
+        ui.unlock()
+
+
+def get_session(region, profile=None):
```
phobologic: needs updated docstring

stacker/session_cache.py
```diff
@@ -15,13 +25,12 @@ def get_session(region):
         :class:`boto3.session.Session`: A boto3 session with
             credential caching
     """
-    session = botocore.session.get_session()
-    if region is not None:
-        session.set_config_variable('region',  region)
-    c = session.get_component('credential_provider')
+    session = boto3.Session(region_name=region, profile_name=profile)
```
phobologic: I think the reason we had to use botocore originally was because boto3 didn't have this ability at the time.  Do we need to update our dependency for boto3 to a specific minimum version for this?

stacker/session_cache.py
```diff
@@ -1,11 +1,21 @@
...
 import os
-import botocore
 import boto3
 import logging
+from getpass import getpass as getpass_orig
+from .ui import ui
 
 
-def get_session(region):
+def getpass(*args):
```
phobologic: Worth moving into the ui library, since we might need this elsewhere?

stacker/session_cache.py
ejholmes: We lock on "boto3>=1.3.1", [which has support for both of these params](https://github.com/boto/boto3/blob/1.3.1/boto3/session.py#L48-L50), so we should be ok.

docs/config.rst
ejholmes: I'll add this to the docs. If both `profile` and `region` are provided, the region that's used in the config overrides the region in the `profile`.

ejholmes:
> how are the s3 buckets handled in this world?

We already talked about this offline, but just adding here: Special consideration will need to happen when using multi-account and uploading templates to s3. Wherever the stacker bucket is provisioned, it'll need a bucket policy that allows any external accounts access to read templates from the bucket. My plan is to update the docs before a 1.3 release with a guide on how to setup a multi-account config.

The easiest way to use this for now is to set `stacker_bucket: ''` so templates are sent direct to CloudFormation.

> how are the lookups handled in this world?

`output` plugin just works, and like multi-region, the `xref` and `rxref` plugins will use the profile in the given stack when performing the lookup.

> > how are the hookups handled in this world?

Same answer  as https://github.com/remind101/stacker/pull/551#issuecomment-372494919. Hooks will use the default AWS config for now.

PR #556, opened 2018-03-14
Remove python-dateutil from setup.py

ejholmes:
Recent versions of botocore lock this to [python-dateutil>=2.1,<2.7.0](https://github.com/boto/botocore/commit/90d7692702be1a423af15e0f49b58365f2a400f2). It may be best for us to just remove the version requirement here and rely on botocore's.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -14,7 +14,6 @@ install_requires = [
     "formic~=0.9b",
     "gitpython~=2.0",
     "schematics~=2.0.1",
-    "python-dateutil~=2.0"
 ]
 
 tests_require = [


setup.py
```diff
@@ -14,7 +14,6 @@
     "formic~=0.9b",
     "gitpython~=2.0",
     "schematics~=2.0.1",
-    "python-dateutil~=2.0"
```
ejholmes: We could also consider mirroring `botocore`'s version spec, since we do have an explicit dependency on `python-dateutil` added in https://github.com/remind101/stacker/pull/487. Not sure what's best.

phobologic: I say keep it in here, set it to `>=2.0` or something, and make sure botocore is first in the list of dependencies.  That should handle it, right?

ejholmes: Isn't botocore already first in the list of dependencies here?

ejholmes:
Gonna merge this for now. If there's a better way to specify this dependency with setup tools, I'll address it before the 1.3 release.

PR #557, opened 2018-03-14
Don't require class_path or template_path for locked stacks

ejholmes:
Technically, "locked" stacks don't actually need a `class_path` or a `template_path`, since they're readonly. This change also fixes a bug introduced in https://github.com/remind101/stacker/pull/550, where the `output` plugin would fail with the following when looking up outputs on locked stacks:

```
FailedVariableLookup: Couldn't resolve lookups in variable `StringVariable`. 'NoneType' object has no attribute '__getitem__'
```

Tests are gonna fail right now until https://github.com/remind101/stacker/pull/556 is merged.


ejholmes:
Good call. I updated the functional test to test for that scenario as well. I'll merge this once CI is passing.

PR #559, opened 2018-03-14
Log step name, instead of stack fqn

ejholmes:
Brought up in https://github.com/remind101/stacker/pull/551#issuecomment-372726710, this changes how steps are logged, so that we log the short "step name" instead of the full stack name. The reasoning for this is that, with multi-account/region support now in, it's entirely possible to have multiple stacks with the same name in a single config (e.g. multiple `vpc`'s in different regions). This also sorta sets the stage for supporting other types of steps in the graph, that aren't CloudFormation stacks (e.g. hooks, targets, etc).

In verbose mode, you're still able to correlate a node in the graph, with the resulting CloudFormation stack name:

```
[2018-03-13T20:48:15] DEBUG app1 stacker.actions.build:288(_launch_stack): Creating new stack: ejholmes-stacker-tests-app1
```

### DIFF
--- a/stacker/plan.py
+++ b/stacker/plan.py
@@ -59,7 +59,7 @@ class Step(object):
         return "<stacker.plan.Step:%s>" % (self.stack.fqn,)
 
     def __str__(self):
-        return self.stack.fqn
+        return self.stack.name
 
     def run(self):
         """Runs this step until it has completed successfully, or been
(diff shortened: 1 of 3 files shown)


ejholmes:
Gonna merge this since we already agreed this is the right thing to do, and it's a 1 line change (aside from updating the tests).

PR #563, opened 2018-03-14
Add a --profile flag to specify default profile

ejholmes:
This adds a `--profile` flag, similar to the `--region` flag that specifies a global default profile for stacker to use for all AWS API calls, when no explicit `profile` is specified.

Another way to do this is is to specify the `AWS_PROFILE` or `AWS_DEFAULT_PROFILE` environment variable, but [in certain configurations that won't actually work](https://github.com/boto/botocore/issues/1401).

### DIFF
--- a/stacker/session_cache.py
+++ b/stacker/session_cache.py
@@ -11,6 +11,8 @@ logger = logging.getLogger(__name__)
 # https://docs.python.org/3/glossary.html#term-global-interpreter-lock
 credential_cache = {}
 
+default_profile = None
+
 
 def get_session(region, profile=None):
     """Creates a boto3 session with a cache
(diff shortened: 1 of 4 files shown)


stacker/session_cache.py
```diff
@@ -23,6 +25,9 @@ def get_session(region, profile=None):
         :class:`boto3.session.Session`: A boto3 session with
             credential caching
     """
+    if profile is None:
+        profile = default_profile
+
```
phobologic: Do you think it's worth adding a debug log in here so we can see what profiles/regions sessions are being created in?

phobologic:
Should we have some documentation about why this flag exists?  Maybe in the flag help syntax itself?  Also does it overwrite `AWS_PROFILE` if both are provided?

ejholmes:
Good call. Extended the flag docs and added some debug logging to `get_session`:

```
[2018-03-15T17:05:52] DEBUG dev/vpc stacker.session_cache:29(get_session): No AWS profile explicitly provided. Falling back to default.
[2018-03-15T17:05:52] DEBUG dev/vpc stacker.session_cache:32(get_session): Building session using profile "dev" in region "us-west-1"
```

PR #564, opened 2018-03-14
hooks: lambda: allow uploading pre-built payloads

danielkza:


stacker/hooks/aws_lambda.py
```diff
@@ -296,30 +338,27 @@ def _upload_function(s3_conn, bucket, prefix, name, options, follow_symlinks):
...
 
-    # os.path.join will ignore other parameters if the right-most one is an
-    # absolute path, which is exactly what we want.
-    if not os.path.isabs(root):
-        root = os.path.abspath(os.path.join(get_config_directory(), root))
-    zip_contents, content_hash = _zip_from_file_patterns(root,
-                                                         includes,
-                                                         excludes,
-                                                         follow_symlinks)
+        return _upload_prebuilt_zip(s3_conn, bucket, prefix, name, options,
```
ejholmes: Kinda bugs me that we have to methods that call `_upload_code`. I think `_upload_prebuilt_zip` and `_build_and_upload_zip` should be changed so that we return a tuple `(zip_file, version)` and then we call `_upload_code` once.

stacker/hooks/aws_lambda.py
```diff
@@ -93,6 +93,18 @@ def _calculate_hash(files, root):
     return file_hash.hexdigest()
 
 
+def _calculate_prebuilt_hash(f):
```
ejholmes: I think we can just re-use [_calculate_hash](https://github.com/remind101/stacker/blob/e123052c6f549559bea1426983786803011f72eb/stacker/hooks/aws_lambda.py#L73-L93).

```python
_calculate_hash("path/to/zip-file.zip")
```

danielkza: That won't be the exact MD5 of the file, as it will append a null character to the stream when calculating.

stacker/hooks/aws_lambda.py
danielkza: I've kept them separated due to the prebuilt case passing the open file through instead of loading all the contents. When that happens `_upload_code` needs to be called in the scope of the context manager, which will require some weird conditionals if the call is unified.

stacker/hooks/aws_lambda.py
ejholmes: Is that a problem? We just need a consistent hash.

stacker/hooks/aws_lambda.py
ejholmes: Gotcha. `_build_and_upload_zip` currently doesn't use the context manager when passing the zip file down to `_upload_code`. Think we should change it so we're consistent?

stacker/hooks/aws_lambda.py
danielkza: Probably not, but it seemed more natural to me (as I just used `md5sum` to figure out the reference values).

phobologic:
@ejholmes @danielkza - thinking about making a 1.4 release soon.  Should this be in it?  What's left to figure out?

danielkza:
@phobologic Rebased and refactored a bit to reduce code duplication.

danielkza:
Rebased, including a refactor of the Lambda hook tests.

danielkza:
Fixed the broken tests.

Issue #112, opened 2015-11-02
Support for stack policies


ejholmes:
CloudFormation has support for setting [stack policies](http://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/protect-stack-resources.html) to prevent updates to protected resources. We should support this within stacker.

PR #570, opened 2018-03-16
Add support for stack policies

ejholmes:
Closes https://github.com/remind101/stacker/issues/112

This is something I've wanted for a long time. Stack policies are a great way to add an extra layer of protection around sensitive resources, like databases, VPC subnets, etc.

It's entirely possible to make use of stack policies outside of stacker (we have for some time now), but it ends up being pretty manual. It'd be great to be able to automate it with stacker.

### DIFF
--- a/stacker/providers/aws/default.py
+++ b/stacker/providers/aws/default.py
@@ -366,6 +366,7 @@ def generate_cloudformation_args(stack_name, parameters, tags, template,
                                  capabilities=DEFAULT_CAPABILITIES,
                                  change_set_type=None,
                                  service_role=None,
+                                 stack_policy=None,
                                  change_set_name=None):
     """Used to generate the args for common cloudformation API interactions.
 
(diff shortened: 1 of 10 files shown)


stacker/providers/aws/default.py
```diff
@@ -414,6 +415,12 @@ def generate_cloudformation_args(stack_name, parameters, tags, template,
     else:
         args["TemplateBody"] = template.body
 
+    if stack_policy:
+        if stack_policy.url:
```
phobologic: Is this for future code, or am I misunderstanding?  I don't see any mention of `stack_policy.url` in this PR.

phobologic: Also, can you throw some a debug statement or two in this block?  That way we can see if someone is using a stack policy with `-v`

ejholmes: Yeah, just for the future, in case we ever support uploading to S3 first. I'll add some debug statements to this.

phobologic: Can you throw a comment in there, just so that if someone sees this they aren't confused in the future?

PR #571, opened 2018-03-17
Document creating stacker bucket with CloudFormation.

ejholmes:
As talked about in https://github.com/remind101/stacker/pull/497#issuecomment-373598384, this documents how more advanced stacker uses that might want more control over creation of the bucket (e.g. tags, default encryption, bucket policies, created in a specific AWS account or region, etc).

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -56,10 +56,29 @@ config.
 If you want to change this, provide the **stacker_bucket** top level key word
 in the config.
 
-The bucket will be created in the same region that the stacks will be launched
-in.  If you want to change this, or if you already have an existing bucket
-in a different region, you can set the **stacker_bucket_region** to
-the region where you want to create the bucket.
+The bucket will be created in the default region (either from ``AWS_REGION`` or
+the ``--region`` CLI flag).  If you want to change this, or if you already have
+an existing bucket in a different region, you can set the
+**stacker_bucket_region** to the region where you want to create the bucket.
+
+Alternatively, if you'd like to have more control over creation of the stacker
+bucket, you can add a stack to your config to create the bucket, "bootstrap"
+the bucket stack, then set the ``stacker_bucket`` value in your config after
+the bucket has been created. Stacker includes a base blueprint to create a
+stacker bucket with AES encryption by default::
+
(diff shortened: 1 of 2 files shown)


ejholmes:
Hey @troyready / @GarisonLotus. This is a slightly more generalized version of what we do internally.

Eventually, I'd like to support automating this more in the config, like I talk about in https://github.com/remind101/stacker/pull/497#issuecomment-366143771, but it's probably worth documenting that his method can be used already.

stacker/blueprints/__init__.py
```diff
@@ -0,0 +1,33 @@
...
+                    SSEAlgorithm="AES256"))
+
+        return s3.Bucket(
+            "StackerBucket",
+            BucketName=bucket_name,
+            BucketEncryption=s3.BucketEncryption(
+                ServerSideEncryptionConfiguration=[aes]))
+
+    def create_template(self):
+        self.template.add_resource(self.bucket)
```
phobologic: Should probably have the bucket name as an output.

stacker/blueprints/__init__.py
```diff
@@ -0,0 +1,33 @@
+from stacker.blueprints.base import Blueprint
+
+from troposphere import Ref
```
phobologic: I'd replace `Ref` with `NoValue` and just use that later, since that's the only use of Ref.

docs/config.rst
```diff
@@ -56,10 +56,29 @@ config.
...
-The bucket will be created in the same region that the stacks will be launched
-in.  If you want to change this, or if you already have an existing bucket
-in a different region, you can set the **stacker_bucket_region** to
-the region where you want to create the bucket.
+The bucket will be created in the default region (either from ``AWS_REGION`` or
+the ``--region`` CLI flag).  If you want to change this, or if you already have
+an existing bucket in a different region, you can set the
+**stacker_bucket_region** to the region where you want to create the bucket.
+
+Alternatively, if you'd like to have more control over creation of the stacker
```
phobologic: Maybe mention that if they want more control over creation of the bucket, they can use the `Bucket` blueprint from `stacker_blueprints` as well.

PR #572, opened 2018-03-20
Add targets

ejholmes:
This adds "targets" to stacker, which are just named nodes in the graph that specify dependencies.

Why would you want this? Primarily just as an easy way to specify a logical grouping of stacks that you can use with the `--targets` flag (e.g. maybe you want a target for "databases" or "apps", etc). Essentially, these are no different than stacks, except they don't incur any AWS API calls, and don't take up a precious CloudFormation stack spot.

If you familiar with [targets in systemd](https://www.freedesktop.org/software/systemd/man/systemd.target.html), the concept is exactly the same; a node in the graph that does nothing.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -348,6 +348,9 @@ A stack has the following keys:
   (optional) a list of other stacks this stack requires. This is for explicit
   dependencies - you do not need to set this if you refer to another stack in
   a Parameter, so this is rarely necessary.
+**required_by:**
+  (optional) a list of other stacks or targets that require this stack. It's an
+  inverse to ``requires``.
 **tags:**
   (optional) a dictionary of CloudFormation tags to apply to this stack. This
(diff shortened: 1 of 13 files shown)


ejholmes:
I don't have much that I want to add to this, but don't necessarily want to merge it until we've got a chance to use it at Remind.

danielkza:
This would be very useful to help integrate hooks with the DAG, which is something I'm currently looking into. I can do some manual testing to help out, anything in particular you are looking for @ejholmes ?

ejholmes:
> I can do some manual testing to help out, anything in particular you are looking for?

Nothing in particular, just some normal testing/feedback. I just rebased this on top of master if you decide you want to try it.

ejholmes:
Well, it's broken against master. I'll take a look at updating this soon.

PR #575, opened 2018-03-24
Example cross-account configuration/structure

ejholmes:
Before we release a 1.3 RC, I want to provide some example documentation for setting up stacker to support cross-account provisioning of stacks.

This is roughly based on how we use stacker internally to provision stacks across all of our AWS accounts. I've tried to keep it as simple as possible, while maintaining good security practices (temporary creds, mfa and encrypted stacker bucket).

Also, it's important to note, this is just _one_ way that you can configure everything. This way works well for us (Remind), but ultimately, everything that applies to http://boto3.readthedocs.io/en/latest/guide/configuration.html, applies to stacker. A solid understanding of AWS CLI configuration and IAM helps.

### DIFF
--- a/examples/cross-account/README.md
+++ b/examples/cross-account/README.md
@@ -1,0 +1,96 @@
+This is a secure example setup to support cross-account provisioning of stacks with stacker. It:
+
+1. Sets up an appropriate [AWS Config File](https://docs.aws.amazon.com/cli/latest/topic/config-vars.html) in [.aws/config] for stacker to use, with profiles for a "master", "prod" and "stage" AWS account.
+2. Configures a stacker bucket in the "master" account, with permissions that allows CloudFormation in "sub" accounts to fetch templates.
+
+## Setup
+
+### Create IAM roles
+
+First things first, we need to create some IAM roles that stacker can assume to make changes in each AWS account. This is generally a manual step after you've created a new AWS account.
+
+In each account, create a new stack using the [stacker-role.yaml](./templates/stacker-role.yaml) CloudFormation template. This will create an IAM role called `Stacker` in the target account, with a trust policy that will allow the `Stacker` role in the master account to `sts:AssumeRole` it.
+
+Once the roles have been created, update the `role_arn`'s in [.aws/config] to match the ones that were just created.
+
+```console
+$ aws cloudformation describe-stacks \
+  --profile <profile> \
+  --stack-name <stack name> \
+  --query 'Stacks[0].Outputs' --output text
+StackerRole     arn:aws:iam::<account id>:role/Stacker
+```
+
+### GetSessionToken
+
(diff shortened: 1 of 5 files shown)


examples/cross-account/README.md
```diff
@@ -0,0 +1,76 @@
...
+
+## Setup
+
+### Create IAM roles
+
+First things first, we need to create some IAM roles that stacker can assume to make changes in each AWS account. This is generally a manual step after you've created a new AWS account.
+
+In each account, create a new stack using the [stacker-role.yaml](./templates/stacker-role.yaml) CloudFormation template. This will create an IAM role called `Stacker` in the target account, with a trust policy that will allow the `Stacker` role in the master account to `sts:AssumeRole` it.
+
+Once the roles have been created, update the `role_arn`'s in [.aws/config] to match. If you created them with the name `Stacker`, then you'll just need to update the AWS account id references.
```
phobologic: I think this means to match the role that was created with the previous template - maybe show how to get that with `stacker info`?  I've even automated this in a private branch I have to mess with the bats tests:

```
# cat tests/Makefile
permissions:
	./stacker.yaml.sh | stacker build -

test: permissions
	$(eval AWS_ACCESS_KEY_ID := $(shell ./stacker.yaml.sh | stacker info - 2>&1 | awk '/AccessKeyId/ {print $$3}'))
	$(eval AWS_SECRET_ACCESS_KEY := $(shell ./stacker.yaml.sh | stacker info - 2>&1 | awk '/SecretAccessKey/ {print $$3}'))
	AWS_ACCESS_KEY_ID=$(AWS_ACCESS_KEY_ID) AWS_SECRET_ACCESS_KEY=$(AWS_SECRET_ACCESS_KEY) bats .
```

examples/cross-account/README.md
```diff
@@ -0,0 +1,76 @@
...
+After we have some IAM roles that stacker can assume, and some temporary credentials, we'll want to create a stacker bucket in the master account, and allow the Stacker roles in sub-accounts access to fetch templates from it.
+
+To do that, first, change the "Roles" variable in [stacker.yaml], then:
+
+```console
+$ aws-vault exec default # GetSessionToken + MFA
+$ AWS_CONFIG_FILE=.aws/config stacker build --stacks stacker-bucket stacker.yaml
+```
+
+Once the bucket has been created, replace `stacker_bucket` with the name of the bucket in [stacker.yaml].
```
phobologic: Again, might be useful to show what to do with stacker info.

examples/cross-account/README.md
ejholmes: Can't really use `stacker info` at this point, since the roles haven't actually been applied to stacker yet (chicken and egg), so I just updated this with an `aws` CLI example.

PR #576, opened 2018-03-24
Add "graph" subcommand

ejholmes:
This adds a new `stacker graph` subcommand, which can be used to print the dependency graph of a stacker config file in one of two formats:

* `dot` will print out the graph in the [graphviz dot format](https://en.wikipedia.org/wiki/DOT_(graph_description_language))
* `json` will print out the graph in a machine readable JSON format.

### DIFF
--- a/stacker/actions/graph.py
+++ b/stacker/actions/graph.py
@@ -1,0 +1,75 @@
+import logging
+import sys
+import json
+
+from .base import BaseAction, plan
+
+
+logger = logging.getLogger(__name__)
+
+
+def each_step(graph):
+    """Returns an iterator that yields each step and it's direct
+    dependencies.
+    """
+
+    steps = graph.topological_sort()
+    steps.reverse()
+
+    for step in steps:
+        deps = graph.downstream(step.name)
+        yield (step, deps)
+
+
+def dot_format(out, graph, name="digraph"):
+    """Outputs the graph using the graphviz "dot" format."""
+
+    out.write("digraph %s {\n" % name)
+    for step, deps in each_step(graph):
+        for dep in deps:
+            out.write("  \"%s\" -> \"%s\";\n" % (step, dep))
+
+    out.write("}\n")
+
+
+def json_format(out, graph):
+    """Outputs the graph in a machine readable JSON format."""
+    steps = {}
+    for step, deps in each_step(graph):
+        steps[step.name] = {}
+        steps[step.name]["deps"] = [dep.name for dep in deps]
+
+    json.dump({"steps": steps}, out, indent=4)
+    out.write("\n")
+
+
+FORMATTERS = {
+    "dot": dot_format,
+    "json": json_format,
+}
+
+
+class Action(BaseAction):
+
+    def _generate_plan(self):
+        return plan(
+            description="Print graph",
+            action=None,
+            stacks=self.context.get_stacks(),
+            targets=self.context.stack_names)
+
+    def run(self, format=None, reduce=False, *args, **kwargs):
+        """Generates the underlying graph and prints it.
+
+        """
+        plan = self._generate_plan()
+        if reduce:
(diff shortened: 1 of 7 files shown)


stacker/dag/__init__.py
```diff
@@ -161,6 +161,30 @@ def walk(self, walk_func):
         for n in nodes:
             walk_func(n)
 
+    def transitive_reduction(self):
```
ejholmes: Note: This algorithm is currently way too un-performant to actually merge. I need to update this with a more efficient transitive reduction.

stacker/actions/graph.py
```diff
@@ -0,0 +1,68 @@
...
+    steps = {}
+    for step, deps in each_step(graph):
+        steps[step.name] = {}
+        steps[step.name]["deps"] = [dep.name for dep in deps]
+
+    json.dump({"steps": steps}, out, indent=4)
+    out.write("\n")
+
+
+FORMATTERS = {
```
phobologic: I tend to throw all constants at the top of the file so they're immediately apparent - that said, just a personal preference.  Feel free to ignore :)

stacker/actions/graph.py
```diff
@@ -0,0 +1,68 @@
...
+            action=None,
+            stacks=self.context.get_stacks(),
+            targets=self.context.stack_names)
+
+    def run(self, format=None, *args, **kwargs):
+        """Generates the underlying graph and brings it.
+
+        """
+        plan = self._generate_plan()
+        fn = FORMATTERS.get(format)
```
phobologic: I'd probably replace `get` with direct access of the dictionary.  What you have now, if the wrong format is given, will result in an AttributeError vs a KeyError.  Even better, might be worth having a specific exception for this.

Also - should there be a default for format?  None doesn't seem right.

stacker/plan.py
```diff
@@ -206,6 +206,8 @@ def build_graph(steps):
         for dep in step.requires:
             graph.connect(step, dep)
 
+    graph.transitive_reduction()
```
phobologic: Curious why this was added to every build_graph call?  Was it meant to provide better performance?

ejholmes: Technically, performing the transitive reduction produces an identical graph, but, it's really not necessary to do it here, so I'm probably gonna pull it out of this.

We only really want to do it for the `dot` format, so that going to a png doesn't produce a mess. Maybe it should be a flag, since I can see times where you might want to see all edges, and not just the simplified version.

phobologic: Yeah, I figured it was only necessary for the dot format.  I think just using it there is probably fine, though you're far more knowledgeable about how the graph is used than me.

stacker/actions/graph.py
ejholmes: Yeah, normally I would too, but that won't work in this case because were reference functions defined above:

```python
NameError: name 'dot_format' is not defined
```

stacker/plan.py
ejholmes: Made this change. The default is to not perform a reduction, and `--reduce` can be passed to generate a "simpler" graph.

stacker/actions/graph.py
phobologic: Ahh, right - ignore me :)

stacker/commands/stacker/graph.py
```diff
@@ -0,0 +1,31 @@
...
+
+    name = "graph"
+    description = __doc__
+
+    def add_arguments(self, parser):
+        super(Graph, self).add_arguments(parser)
+        parser.add_argument("-f", "--format", default="dot",
+                            choices=list(graph.FORMATTERS.iterkeys()),
+                            help="The format to print the graph in.")
+        parser.add_argument("--reduce", action="store_true",
```
phobologic: Maybe throw a note in here that while this produces a less noisy/better graph diagram, it's slower.

PR #585, opened 2018-04-13
FEATURE: dump blueprints to yaml

zaro0508:
Add the ability to dump stacker blueprints (in troposphere) to yaml format.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -44,6 +44,13 @@ See the `CloudFormation API Reference`_ for allowed stack name characters
 
 .. _`CloudFormation API Reference`: http://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_CreateStack.html
 
+Dump Format
+-----------
+
+By default, stacker will dump templates to 'json' format.  You can pass the **dump_format**
+top level key word in the config to change the format to 'yaml'.
+
+
 S3 Bucket
 ---------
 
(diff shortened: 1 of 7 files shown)


docs/config.rst
```diff
@@ -44,6 +44,13 @@ See the `CloudFormation API Reference`_ for allowed stack name characters
 
 .. _`CloudFormation API Reference`: http://docs.aws.amazon.com/AWSCloudFormation/latest/APIReference/API_CreateStack.html
 
+Dump Format
```
phobologic: Hmm - I think this makes more sense as a command line flag/argument to the `dump` command.  That way it's easy to dump in either format without having to manipulate configuration.  Any chance you could take a stab at that method?

zaro0508: My thought on this is that most people like to look at either json or yml but not really ever want to switch between the two that often.  Therefore i'm not sure how useful a command line switch would be.  This is the kind of thing that I like to set and never change again.

zaro0508: I just realized that it's already possible to override the dump format from a command line switch.

Add `dump_format: ${format}` to stacker.yml

On command line: `stacker build --region us-east-1 -e format=yaml -d DUMP conf/dev.env stacker.yaml`

phobologic: Sorry @zaro0508 - thought I'd responded to this in Slack, but looking back I realized I hadn't.  At this point I really think there are two best paths:

- Make it an optional argument to `build --dump`, like `--dump-format`.  The reason being that it's dump specific, and we don't really have any action specific configuration variables at the top level that I'm aware of.  It also feels very situational - each user would probably want to control this themselves.
- Make it a more generic configuration variable, like `template_format` - but then it'd need to be used in more places, like in the build action when uploading, etc.

I'm open to either of those - what do you think?

PR #586, opened 2018-04-19
Run dump and outline actions offline

zaro0508:
Build does not need access to AWS when run with `dump` or
`outline` options so make it run offline for those actions.


zaro0508:
I somehow messed up this PR, replacing it with PR https://github.com/remind101/stacker/pull/594

PR #589, opened 2018-04-23
Fixes 566

### LINKED ISSUE #566: stacker.status.Status.__cmp__ implementation is wrong
Came across [this](https://github.com/remind101/stacker/blob/db2119270a6d903f5bba6e830b2818769957bc28/stacker/status.py#L7) while adding python3 support and it feels like there could be some subtle fallout. [Per the docs](https://docs.python.org/2/reference/datamodel.html#object.__cmp__) this is supposed to return zero iff the objects are equal, and since ```False == 0``` any Status is equal to any object without a code attribute. The only reason I stumbled across it is ```__cmp__``` isn't supported in py3 and one of the tests fails (forget which offhand), but the test compares a Status to True which currently always succeeds:

```
Python 2.7.8 (default, Oct 18 2017, 15:41:14)
[GCC 4.2.1 Compatible Apple LLVM 8.1.0 (clang-802.0.42)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> import stacker.status
>>> stacker.status.SUBMITTED == stacker.status.PENDING
False
>>> stacker.status.SUBMITTED == True
True
>>> stacker.status.SUBMITTED == False
True
>>> stacker.status.SUBMITTED == 'banana'
True
>>>
```

Everything I needed for py3 support was obvious and mechanical except this.

russellballestrini:
Fixes #566 

One of our tests had a logic error where we were using the returned
result of Step.run (which was always self.ok [which is always True])
and comparing this to a Status object.

The actual test should be comparing Step.status with SKIPPED.

	modified:   stacker/plan.py
	modified:   stacker/status.py
	modified:   stacker/tests/actions/test_destroy.py

### DIFF
--- a/stacker/plan.py
+++ b/stacker/plan.py
@@ -82,6 +82,7 @@ class Step(object):
             if watcher:
                 stop_watcher.set()
                 watcher.join()
+        return self.ok
 
     def _run_once(self):
         try:
--- a/stacker/plan.py
+++ b/stacker/plan.py
@@ -82,7 +82,6 @@ class Step(object):
             if watcher:
                 stop_watcher.set()
                 watcher.join()
-        return self.ok
 
     def _run_once(self):
         try:
--- a/stacker/status.py
+++ b/stacker/status.py
@@ -4,10 +4,17 @@ class Status(object):
         self.code = code
         self.reason = reason or getattr(self, "reason", None)
 
+    def cmp(self, a, b):
+        try:
+            return cmp(a, b)
+        except NameError:
+            # Python3 doesn't have cmp function.
+            return ((a > b) - (a < b))
+
     def __cmp__(self, other):
         if hasattr(other, "code"):
-            return cmp(self.code, other.code)
-        return False
+            return self.cmp(self.code, other.code)
+        raise Exception("Both Status objects must have a `code` attribute.")
 
 
 class PendingStatus(Status):
--- a/stacker/status.py
+++ b/stacker/status.py
@@ -1,20 +1,34 @@
+import operator
+
+
 class Status(object):
     def __init__(self, name, code, reason=None):
         self.name = name
         self.code = code
         self.reason = reason or getattr(self, "reason", None)
 
-    def cmp(self, a, b):
-        try:
-            return cmp(a, b)
-        except NameError:
-            # Python3 doesn't have cmp function.
-            return ((a > b) - (a < b))
-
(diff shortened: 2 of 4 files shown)


stacker/status.py
```diff
@@ -4,10 +4,17 @@ def __init__(self, name, code, reason=None):
...
+        except NameError:
+            # Python3 doesn't have cmp function.
+            return ((a > b) - (a < b))
+
     def __cmp__(self, other):
         if hasattr(other, "code"):
-            return cmp(self.code, other.code)
-        return False
+            return self.cmp(self.code, other.code)
+        raise Exception("Both Status objects must have a `code` attribute.")
```
wrmsr: The correct thing to do here is to ```return NotImplemented``` (and usually if the types are different too). In python2 this will propagate to the default logic of an identity comparison, which is at least internally consistent, and is definitely better than accidentally returning unconditional equality via ```False``` meaning "no difference". In python3 however this code won't even get called as ```__cmp__``` is no longer a magic method but that can be left for the py3 branch.

As before my reason for raising the issue and not just steamrolling over it was concern over the possibility of the ordering of these things mattering somewhere and this bug's behavior subtlely being actually relied upon, but as that isn't the case any correction is an improvement.

stacker/plan.py
```diff
@@ -82,7 +82,6 @@ def run(self):
             if watcher:
                 stop_watcher.set()
                 watcher.join()
-        return self.ok
```
phobologic: Why was this removed?  Are you sure this isn't used anywhere?  I'm pretty sure the return code is used by the walk method.

phobologic: Right, does this cause any issues though?  Like - if we're going for cleanliness of code, you should probably look through  the call sites to figure out if they need it.  Maybe they don't - and we should probably remove more return statements.  @ejholmes will likely know more, but this goes through the graph and into the threads.

russellballestrini: @ejholmes do you remember why you return self.ok was this supposed to be self.status ?

ejholmes: The original DAG implementation used the Step's return value to determine if any dependencies failed, but that's no longer the case since that's done here:

https://github.com/remind101/stacker/blob/91a571dd85a5474d87b8250851e4ae13071512cf/stacker/plan.py#L368-L376

Removing the return value should be safe (and all tests pass, so).

PR #591, opened 2018-05-04
Fix datetime is not JSON serializable error

aarongorka:
`TypeError: datetime.date(2012, 10, 17) is not JSON serializable`

This fixes the above exception raised by json.dumps when the yaml
library loads an unquoted date value e.g. `2012-10-17` as a
`datetime.time` object.


phobologic:
This seems like a fine idea - I can't think of any cases where it would cause any unexpected issues.  @ejholmes any thoughts?

phobologic:
BTW, thanks @aarongorka !

PR #594, opened 2018-05-08
Run dump and outline actions offline

zaro0508:
Build does not need access to AWS when run with `dump` or
`outline` options so make it run offline for those actions.

### DIFF
--- a/stacker/actions/build.py
+++ b/stacker/actions/build.py
@@ -72,6 +72,20 @@ def should_submit(stack):
     return False
 
 
+def should_ensure_cfn_bucket(outline, dump):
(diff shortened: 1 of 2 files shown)


stacker/actions/build.py
```diff
@@ -72,6 +72,20 @@ def should_submit(stack):
...
+
+    Args:
+        outline (bool): The outline action.
+        dump (bool): The dump action.
+
+    Returns:
+        bool: If access to CF bucket is needed, return True.
+
+    """
+    return outline is False and dump is False
```
zaro0508: may need to consider package sources as well.  if user sets up package sources then I think stacker needs to run in online mode.

PR #595, opened 2018-05-11
Change default region in ssmstore lookup.

xiaket:
In most cases, user have specified the AWS region from the command line, and we can just get this region from the provider and retrieve parameters from that region. If this lookup failed, still fall back to `us-east-1`.

### DIFF
--- a/stacker/lookups/handlers/ssmstore.py
+++ b/stacker/lookups/handlers/ssmstore.py
@@ -13,7 +13,8 @@ def handler(value, **kwargs):
 
         [<region>@]ssmkey
 
-    Note: The region is optional, and defaults to us-east-1 if not given.
+    Note: The region is optional, we'll try to get it from the current
+    session, if that failed, it would fall back to us-east-1
 
     For example:
 
@@ -36,7 +37,7 @@ def handler(value, **kwargs):
     """
     value = read_value_from_path(value)
 
-    region = "us-east-1"
+    region = getattr(kwargs.get('provider'), 'region', 'us-east-1')
     if "@" in value:
         region, value = value.split("@", 1)
 


xiaket:
Sorry reviewers, I would love to add a test for this change, but I cannot think of a good way to do it. If you've got some ideas, I'd love to implement them.

CircleCI failed for some obscure reason that I cannot understand, I guess its not related to the change I've made?

xiaket:
Thanks @phobologic ! I had tried to make a little contribution back then and I totally forget about it until you pointed out, that's a stupid mistake.

Anyway, I've also added the tests, hope that will make sense.

phobologic:
Hey @xiaket - sorry for not getting back to you sooner.  This PR actually made me realize that we potentially have a bit bigger of an issue - mostly that our lookups don't use any common way of determining what region to use.  I think that might be a good thing to do going forward - and would be more easily testable in general (really, the logic isn't around if we use the right end point provided with a region - the logic is how our lookups CHOOSE their region).

I'm wondering if we should instead have a common helper method for determining the region used by lookups?  Going to ping some other folks to get their opinion here - @troyready @ejholmes @danielkza  @GarisonLotus @russellballestrini @Lowercases

xiaket:
Thanks, @phobologic for your reply! I agree, adding a new general way to choose the aws region would be really useful.

troyready:
I'd prefer this to go into a 2.0 branch, since it technically changes existing behavior (there may be users deploying to oregon/etc that have parameters in virginia and are relying on the default lookup).

Agree on the idea of adding a common method as well. This implementation seems fine for now - not sure how it interacts with the new multi-profile/region support?

PR #597, opened 2018-05-13
Add helper Makefile for more easier integration tests

phobologic:


russellballestrini:
Looks like you have failing functional tests.

tests/Makefile
```diff
@@ -0,0 +1,8 @@
+permissions:
+	./stacker.yaml.sh | stacker build -
+
+test: permissions
```
ejholmes: We already have a `test-functional` target in the top level Makefile. Maybe move this there? https://github.com/remind101/stacker/blob/master/Makefile#L13

ejholmes: Yeah

phobologic:
I'm going to go ahead and merge this - we can figure out if we want to somehow merge this with the main Makefile tests later.  One thing to keep in mind - this changes the way the tests are run now quite a bit, making it easier to run a single test.  Please see the readme @ejholmes @russellballestrini

PR #601, opened 2018-05-24
Replace remind101 references with cloudtools where appropriate

phobologic:


ejholmes:
Do we need to make any changes to readthedocs? The badge in the readme is broken right now:

phobologic:
I think that was a cached version @ejholmes - this, plus another change, was meant to fix those badges (and it looks good now).  Let me know if you're still seeing the issue.

PR #611, opened 2018-06-25
Add `add_output` helper method on Blueprints

phobologic:
Got sick of constantly typing `Output("Blah", Value="foo")` and figured
we could make this easier.


russellballestrini:
@phobologic do you think we should do some verification of the `name` value so that we check if it is safe?

phobologic:
My thought was to let troposphere do the validation, which it should already be doing in theory.  That way if the rules change, and troposphere changes to match them, we get the benefits immediately.

Issue #612, opened 2018-06-28
A new PyYAML was released stacker competes with dependicies


russellballestrini:
https://github.com/yaml/pyyaml/releases

```
error: PyYAML 4.1 is installed but PyYAML~=3.12 is required by set(['stacker'])
Exited with code 1
```

phobologic:
Fixed in #613 I believe

PR #613, opened 2018-06-28
Use a less strict pinning for PyYAML

russellballestrini:
Related to #612


setup.py
```diff
@@ -9,7 +9,7 @@
     "troposphere>=1.9.0",
     "botocore>=1.6.0",
     "boto3>=1.3.1",
-    "PyYAML~=3.12",
+    "PyYAML>=3.12,<=4.1",
```
phobologic: I'd remove the max here - unless we know it won't work with anything newer.

Issue #598, opened 2018-05-15
TypeError: Got unexpected keyword argument 'retries'


jamietsao:
We recently tried upgrading stacker and got an error:
```
[2018-05-15T15:28:22] ERROR production-bot-refresh-service stacker.plan:92(_run_once): Got unexpected keyword argument 'retries'
Traceback (most recent call last):
  File "/Users/jamie/.pyenv/versions/2.7.11/lib/python2.7/site-packages/stacker/plan.py", line 88, in _run_once
    status = self.fn(self.stack, status=self.status)
  File "/Users/jamie/.pyenv/versions/2.7.11/lib/python2.7/site-packages/stacker/actions/diff.py", line 216, in _diff_stack
    provider_stack = self.provider.get_stack(stack.fqn)
  File "/Users/jamie/.pyenv/versions/2.7.11/lib/python2.7/site-packages/stacker/providers/aws/default.py", line 489, in get_stack
    return self.cloudformation.describe_stacks(
  File "/Users/jamie/.pyenv/versions/2.7.11/lib/python2.7/site-packages/stacker/providers/aws/default.py", line 478, in cloudformation
    max_attempts=MAX_ATTEMPTS
  File "/Users/jamie/.pyenv/versions/2.7.11/lib/python2.7/site-packages/botocore/config.py", line 94, in __init__
    args, kwargs)
  File "/Users/jamie/.pyenv/versions/2.7.11/lib/python2.7/site-packages/botocore/config.py", line 119, in _record_user_provided_options
    'Got unexpected keyword argument \'%s\'' % key)
TypeError: Got unexpected keyword argument 'retries'
[2018-05-15T15:28:22] DEBUG production-bot-refresh-service stacker.plan:148(set_status): Setting bot-refresh-service state to failed.
[2018-05-15T15:28:22] INFO production-bot-refresh-service stacker.ui:36(info): production-bot-refresh-service: failed (Got unexpected keyword argument 'retries')
[2018-05-15T15:28:22] ERROR MainThread stacker.actions.base:198(execute): The following stacks failed: production-bot-refresh-service
```

We were on version 1.0.4 previous and tried upgrading to 1.2.* (I initially tried going to 1.3.* but got the same error and narrowed it down to the 1.2.* upgrade).

I haven't changed anything else (e.g. blueprints, stack config, etc.) so it's very possible that it's a backwards compatibility issue.  It's just hard to debug since the error is not very helpful.

[...]

PR #614, opened 2018-07-01
Include lookup contents when lookups fail

phobologic:
This would have helped with #598 if we had known, from the exception, the
actual lookup that had failed.  The only way I could think of doing this
was to move the failed lookup handling deeper into the stack, though I
think I like the way this works better than before.

Now, when a lookup fails, it'll look like this:

```
FailedVariableLookup: Couldn't resolve lookup in variable `MyVar`, lookup: ${output foo::bar}: (AttributeError) 'NoneType' object has no attribute 'outputs'
```


stacker/tests/lookups/test_registry.py
```diff
@@ -1,12 +1,35 @@
...
 
 
 class TestRegistry(unittest.TestCase):
+    def setUp(self):
+        self.ctx = mock_context()
+        self.provider = mock_provider()
 
     def test_autoloaded_lookup_handlers(self):
-        handlers = ["output", "xref"]
+        handlers = [
```
phobologic: I realized we weren't testing all the auto-loaded lookups, so I went ahead and added them.

PR #618, opened 2018-07-03
Formalize the existence of external (read-only) stacks

danielkza:
Currently, the `ref` and `xref` lookups always use the same AWS profile and region that Stacker is initialized with. That means they cannot access outputs in other accounts or regions. It is possible to create a stack that will just "pull" those outputs by adding `locked: true` to it. But that still does not make it possible to reference a stack outside of the current namespace, if one is configured. Additionally, options that should never take effect for a "read-only" stack were allowed, such as `class_path` or `stack_policy_path`.

To handle that use case, introduce and document the `external: yes` option for stacks. External stacks do not have any of the options related to creating or updating resources, such as `class_path` and `template_path`, and are used only as sources of values to other stacks (through outputs and lookups).
Additionally, add an `fqn` option to external stacks to fix the namespace issue.

# Internal changes

To ensure the separation has the desired effect, enable `strict` mode while parsing configuration into models, but *only* for non-top-level keys. Achieve it by manually filtering unrecognized top-level keys before forwarding the data for parsing.

To better separate external and managed stacks in config, rework the model hierarchy a little: add `BaseStack` that only contains the shared config options between both stack types, make it a superclass of `Stack`, and introduce `ExternalStack`. Unfortunately Schematics 2.0 has a bug where it did not validate models parsed using polymorphism, so upgrade to version 2.1.

Move decisions on whether a stack needs to be submitted or updated to the `stacker.stack.Stack` class, to simply the logic and allow it to differ between external and managed stacks. Introduce `stacker.stack.ExternalStack` with "empty" implementations of all the methods/properties related to templates, blueprints and updates.

# Documentation updates

Rework the Stack documentation to list options allowed for all stacks, external and managed stacks separately. Document everything in YAML format, so that people can copy snippets to their configs from the different parts. Fix some typos and clarify some details in the process.

# Open questions

The `external/managed` terminology is the best I could come up with, but I still feel it can be improved. Any suggestions of alternatives are welcome!

# Notes

I ran functional tests manually on my all account and they passed - I even had to fix one that had `stack_policy_path` misspelled and was no longer accepted due to strict parsing.

### DIFF
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -362,7 +362,10 @@ Managed Stack Configuration
 External Stack Configuration
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-External stacks have no additional configuration options.
+.. literalinclude:: stack-config-external.yml
+   :language: yaml
+   :name: stack-config-external
+
 
 Stacks Example
 ~~~~~~~~~~~~~~
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -367,13 +367,20 @@ External Stack Configuration
    :name: stack-config-external
 
 
-Stacks Example
-~~~~~~~~~~~~~~
+Examples
+~~~~~~~~
 
-Here's an example from stacker_blueprints_, used to create a VPC::
+VPC + Instances
+:::::::::::::::
 
+Here's an example from stacker_blueprints_, used to create a VPC and and two EC2
+Instances::
+
+
+  namespace: example
   stacks:
-    - name: vpc-example
+    - name: vpc
+      stack_name: test-vpc
       class_path: stacker_blueprints.vpc.VPC
       locked: false
       enabled: true
@@ -393,9 +400,49 @@ Here's an example from stacker_blueprints_, used to create a VPC::
           - 10.128.16.0/22
           - 10.128.20.0/22
         CidrBlock: 10.128.0.0/16
-    - name: external-vpc-example
-      stack_name: my-dev-account-vpc
-      profile: dev
+
+    - name: instances
+      stack_name:
+      class_path: stacker_blueprints.ec2.Instances
+      enabled: true
+      variables:
+        SmallInstance:
+          InstanceType: t2.small
+          ImageId: &amazon_linux_ami "${ami owners:amazon name_regex:amzn-ami-hvm-2018.03.*-x86_64-gp2}"
+          AvailabilityZone: ${output vpc::AvailabilityZone0}
+          SubnetId: ${output vpc::PublicSubnet0}
+        LargeInstance:
+          InstanceType: m5.xlarge
+          ImageId: *amazon_linux_ami
+          AvailabilityZone: ${output vpc::AvailabilityZone1}
+          SubnetId: ${output vpc::PublicSubnet1}
+
+
+Referencing External Stacks
+:::::::::::::::::::::::::::
+
+This example creates a security group in VPC from the previous example by
+importing it as an external stack with a custom profile::
+
+  namespace: other-example
+  stacks:
+    - name: vpc
+      fqn: example-test-vpc
+      profile: custom-profile
+      external: yes
+
+    - name: sg
+      class_path: stacker_blueprints.ec2.SecurityGroups
+      variables:
+        SecurityGroups:
+          VpcId: ${output vpc::VpcId}
+          SecurityGroupIngress:
+            - CidrIp: 0.0.0.0/0
+              FromPort: 22
+              ToPort: 22
+              IpProtocol: tcp
+
+
 
 Variables
 =========
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -1,3 +1,5 @@
+.. highlight:: yaml
+
 =============
 Configuration
 =============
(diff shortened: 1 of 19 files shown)


danielkza:
Should't the unit tests also be running on Python 3 now?

stacker/config/__init__.py
```diff
@@ -289,6 +284,29 @@ class Stack(Model):
...
+    def _claim_polymorphic(cls, value):
+        if value.get("external", False) is True:
+            return True
+
+    def __init__(self, *args, **kwargs):
+        super(ExternalStack, self).__init__(*args, **kwargs)
+        self.external = True
+
+    def validate_fqn(self, data, value):
+        if (not value and not data["stack_name"]) or \
```
danielkza: I just noticed this is wrong, since we can infer the CF name from the stack name. Will fix!

danielkza:
That was my first approach, but I noticed more than half of an external
stack's attributes would be completely invalid/useless, and that I needed
to add conditions that checked `stack.external` in dozens of places.
Additionally, being loose with parsing can be an easy source of mistakes,
so much that one of Stacker's own functional tests had a mistyped key that
was never noticed!

While there were a significant amount of changes, adding enough tests to
the Stack models to cover all cases where external was at would be as big
if not bigger.


Em ter, 3 de jul de 2018 20:25, Eric Holmes <notifications@github.com>
escreveu:

stacker/context.py
```diff
@@ -137,15 +137,18 @@ def get_stacks(self):
...
-                stack = Stack(
-                    definition=stack_def,
-                    context=self,
-                    mappings=self.mappings,
-                    force=stack_def.name in self.force_stacks,
-                    locked=stack_def.locked,
-                    enabled=stack_def.enabled,
-                    protected=stack_def.protected,
-                )
+                if isinstance(stack_def, ExternalStackModel):
```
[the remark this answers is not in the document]
danielkza: That was my first approach, but I noticed more than half of an external stack's attributes would be completely invalid/useless, and that I needed to add conditions that checked `stack.external` in dozens of places. Additionally, being loose with parsing can be an easy source of mistakes, so much that one of Stacker's own functional tests had a mistyped key that was never noticed! While there were a significant amount of changes, adding enough tests to the Stack models to cover all cases where external was at would be as big if not bigger.

danielkza:
Interesting note: I found 20+ usages of invalid options in stacks in my company's repo when using this PR (due to enabling strict parsing). Most of them were harmless (using `enable` instead of `enabled`), but a few were a bit more dangerous.

I can make a PR that just enables strict parsing before this one (which is larger and a bit more controversial). What do you think @ejholmes?

docs/config.rst
```diff
@@ -419,8 +401,51 @@ Here's an example from stacker_blueprints_, used to create a VPC::
...
+
+Referencing External Stacks
+:::::::::::::::::::::::::::
+
+This example creates a security group in VPC from the previous example by
+importing it as an external stack with a custom profile::
+
+  namespace: other-example
+  stacks:
+    - name: vpc
```
phobologic: maybe throw a comment here stating that this stack won't be changed/build/etc because of the external flag (it is redundant, but I think it's useful)

docs/stack-config-managed.yml
```diff
@@ -0,0 +1,69 @@
...
+# switch to interactive mode before applying the changes, allowing you to
+# approve/skip the change.
+protected: false
+
+# requires (list of str, optional): A list of other stack names that must be
+# created/updated before this stack. This is not needed on normal usages, as
+# dependencies are automatically created when using an output from a stack in
+# the variables of another stack. Use it when you know that a hidden dependency
+# exists (or one is documented in the CloudFormation reference).
+dependencies:
```
phobologic: This should be `requires`, right?

phobologic:
@ejholmes @danielkza - I'm thinking we should spin a release of stacker this week, it's been quite a while, and we have a lot of features/updates already.  Do you think we should wait for this?  What's left for discussing here?

danielkza:
I'm looking to revive this PR as I update our internal Stacker fork to the latest master. The strict parsing change has been integrated for a while already, any other concerns I should handle to get this merged? @ejholmes @phobologic

docs/config.rst
```diff
@@ -324,81 +326,61 @@ Stacks
...
-Stacks Example
-~~~~~~~~~~~~~~
+stacks that will be part of your environment.
+The top level keyword ``stacks`` is populated with a list of dictionaries, each
+representing a single stack.
+
+Stacks can be classified as:
+
+Managed
+  Stacks that are fully managed by the current Stacker configuration.
```
phobologic: Worth mentioning this is the default?

docs/config.rst
```diff
@@ -324,81 +326,61 @@ Stacks
...
+  They will be created, updated or destroyed as needed.
+External
+  Stacks that will not be modified by the current Stacker configuration.
```
phobologic: Might be worth saying that these stacks are configured with the `external` flag.

docs/stack-config-managed.yml
phobologic: Ping on this one ^^

stacker/stack.py
```diff
@@ -192,3 +193,94 @@ def resolve(self, context, provider):
...
+    def required_parameter_definitions(self):
+        return dict()
+
+    def resolve(self, context, provider):
+        pass
+
+    def set_outputs(self, outputs):
+        self.outputs = outputs
+
+    def should_submit(self):
```
phobologic: What does this end up looking like in the CLI output when we hit an external stack?  Does it say it's submitting/creating?

tests/suite.bats
```diff
@@ -162,7 +162,7 @@ namespace: ${STACKER_NAMESPACE}
 stacks:
   - name: vpc
     class_path: stacker.tests.fixtures.mock_blueprints.VPC
-    stack_policy: tests/fixtures/stack_policies/default.json
+    stack_policy_path: fixtures/stack_policies/default.json
```
phobologic: We broke these up into individual files, so you'll probably want to merge master and get that cleared up here.

stacker/stack.py
danielkza: I believe it prints "skipped". Do you think we should add an additional status to print something different?

docs/config.rst
danielkza: Yeah, I can clarify this section a bit!

docs/stack-config-managed.yml
danielkza: Indeed, I missed it :)

docs/config.rst
danielkza: Sure.

danielkza:
Rebased to latest master. Was a bit harder than what I hoped, and I ended up having to make some changes to the build action tests.

I'll address your review comments in further commits @phobologic :)

stacker/stack.py
phobologic: Hmm - yeah, that might be useful.  It can still be skipped, just with a slightly different message maybe?

danielkza:
I also found some issues running the functional tests, good thing we have those!

danielkza:
I'm circling back to this. Rebased to latest master again, but unfortunately I'm setting failed unit tests due to #709.

danielkza:
@phobologic Things looked up-to-date to me, but I took the opportunity to improve the wording on the docs a little bit and forced-pushed.

phobologic:
Weird!  I wonder why it is showing the additions of stack_name, which is already in the master branch. Ok, once tests pass I'll go ahead and merge :)

danielkza:
I think it's because I moved it around in the file a little bit.

PR #623, opened 2018-07-08
Reject unrecognized options in config objects (but still allow it at the top-level)

danielkza:


stacker/config/__init__.py
```diff
@@ -137,19 +141,12 @@ def parse(raw_config):
...
-    # should consider enabling this in the future.
-    strict = False
-
-    return Config(config_dict, strict=strict)
+    # Top-level excess keys are removed by Config._convert, so enabling strict
+    # mode is fine here.
+    try:
+        return Config(config_dict, strict=True)
+    except SchematicsError as e:
+        raise exceptions.InvalidConfig(e.errors)
```
phobologic: I don't know the schematics Exceptions well enough - but will this provide enough data to figure out what the issue is?  I really wish python2 provided a good way for wrapping exceptions.

stacker/tests/test_config.py
```diff
@@ -537,6 +537,20 @@ def test_raise_construct_error_on_duplicate_stack_name_dict(self):
         with self.assertRaises(ConstructorError):
             parse(yaml_config)
 
+    def test_parse_invalid_inner_keys(self):
```
phobologic: Do we already have a test that has excess keys at the top level that would cover that use case?

stacker/config/__init__.py
danielkza: Yes, the only issue is the message is not very explanatory. `e.errors` is something like this when an invalid key is present: `{"stacks": {"0": {"garbage": "Rogue field"}}}`. But I think the best place to fix it is in Schematics itself.

stacker/tests/test_config.py
danielkza: `test_parse_with_arbitrary_anchors` seems to do the job: https://github.com/Cobliteam/stacker/blob/94cc1e8a4eea1eb7ee8329caa4b25a61cd98c648/stacker/tests/test_config.py#L168

phobologic: Awesome, thanks!

phobologic:
Thanks @danielkza !

PR #624, opened 2018-07-08
Cleanup all but last 6 buckets created by functional tests

### LINKED ISSUE #558: Functional test suite creates too many stacker buckets in CI
Each time the functional test suite is ran in CI, it creates a unique stacker bucket for the CI build. Eventually, we rack up a bunch of S3 buckets (about 100 when I checked today, which I already cleaned up). We should change the test suite so that each test either:

1. It uses an explicit `stacker_bucket`. Or..
2. We set `stacker_bucket: ""` to disable creating a bucket.

phobologic:
6 is a totally arbitrary #, but I figure it might be worth leaving some
around in case you want to troubleshoot something.

Fixes #558

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -133,5 +133,5 @@ jobs:
       - run:
           command: |
             sudo pip install awscli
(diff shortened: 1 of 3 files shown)


russellballestrini:
Great work, could we `Squash and Merge` when this is ready?

phobologic:
Yep!  Always Sqush Merge :)

PR #632, opened 2018-07-20
Add --only flag to build command

ejholmes:
This adds an `--only` flag to `stacker build`, which, when used in combination with the `--stacks` flag, can make it easier for workflows where you want to work on a single stack, without effecting dependencies.

Take the following example. Let's say we have a graph like this:

```
app -> cluster -> vpc
```

If we want to work on the `app` stack, we can specify `--stacks app`, and stacker will filter the graph so that only `app` and it's transitive dependencies are operated in. However, this has some disadvantages:

1. If `cluster` is currently undergoing a change, `stacker build` will fail.
2. stacker works through the entire dependency graph, trying to update all transitive dependencies.

With the addition of the `--only` flag, all stacks other than those explicitly specified with `--stacks` are marked as `locked`. This means that you can operate on only the `app` stack, without worry about what state that transitive dependencies are in.

### DIFF
--- a/stacker/stack.py
+++ b/stacker/stack.py
@@ -84,6 +84,12 @@ class Stack(object):
 
     @property
     def requires(self):
+        # By definition, a locked stack has no dependencies, because we won't
+        # be performing an update operation on the stack. This means, resolving
+        # outputs from dependencies is unnecessary.
+        if self.locked:
+            return []
+
         requires = set(self.definition.requires or [])
 
         # Add any dependencies based on output lookups
--- a/stacker/commands/stacker/build.py
+++ b/stacker/commands/stacker/build.py
@@ -34,6 +34,9 @@ class Build(BaseCommand):
                                  "dependencies. Can be specified more than "
                                  "once. If not specified then stacker will "
                                  "work on all stacks in the config file.")
+        parser.add_argument("--only", action="store_true",
+                            help="Lock all stacks, other than those provided "
+                                 "in the --stacks flag.")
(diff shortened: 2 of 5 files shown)


stacker/stack.py
```diff
@@ -84,6 +84,12 @@ def __repr__(self):
 
     @property
     def requires(self):
+        # By definition, a locked stack has no dependencies, because we won't
+        # be performing an update operation on the stack. This means, resolving
+        # outputs from dependencies is unnecessary.
+        if self.locked:
```
ejholmes: This change is pretty useful by itself, even if we don't add the `--only` flag. There's no reason for locked stacks to depend on anything, because by definition, they don't depend on anything.

ejholmes: Good catch. If the stack is in `--forced`, it still gets updated regardless of if it's locked or not:

https://github.com/cloudtools/stacker/blob/87688453c64ef333c48786000f2f890cbe664633/stacker/actions/build.py#L39-L57

It might be better if we moved that check to where we initialize the `stacker.Stack`, and just unlock the stack if it's in `--force`, so that `locked` always means what you think it means.

phobologic: Hmm, yeah, I kind of like the "cleanness" of the way it is now - that a stack is locked if it's locked, but that we ignore it if we're told to.  That said, I do like the feature you're building in here.

Another question - how would this affect things like the `dag` command?  If something was locked, would it look like it was an unattached node of the graph?

PR #636, opened 2018-08-03
Documentation update: Adding Default Lookup to official docs

GarisonLotus:
This has been implemented since February.

### DIFF
--- a/docs/lookups.rst
+++ b/docs/lookups.rst
@@ -94,19 +94,28 @@ default Lookup
 --------------
 
 The ``default`` lookup type will check if a value exists for the variable
-in the environment, then fall back to a default defined in the stacker
-config if the environment doesn't contain the variable. This allows defaults
-to be set at the config file level, while granting the user the ability
-to override that value per environment.
+in the environment file, then fall back to a default defined in the stacker
+config if the environment file doesn't contain the variable. This allows
+defaults to be set at the config file level, while granting the user the
+ability to override that value per environment.
 
 Format of value::
   <env_var>::<default value>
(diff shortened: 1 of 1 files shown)


GarisonLotus:
@phobologic - Just an FYI, no actual code updates in this PR.

docs/lookups.rst
```diff
@@ -86,6 +87,27 @@ You can specify an output lookup with the following syntax::
 
   ConfVariable: ${output someStack::SomeOutput}
 
+
+.. _`default lookup`:
+
+default Lookup
+--------------
+
+The ``default`` lookup type will check if a value exists for the variable
```
phobologic: This isn't 100% true, right?  It can also be used to give default variables to lookups.

```
${default ${output MyVpc::VpcId}::vpc-123456}
```

GarisonLotus: I... hadn't, really thought of that. It's an interesting idea. I didn't write the lookup module, so let me dig into it to check if that would work. (Or @troyready could chime in, since he wrote it)

GarisonLotus:
@phobologic - Troy and I just spoke about this idea, and we don't believe that this would work.  Although we have not tested it, we think that stacker would throw an exception for the output lookup in your example, stoping the default fall back from completing. 

I'm thinking it might be a good idea to make it more obvious that we mean "environment file" in the documentation, since that is how the feature actually was implemented. 

EDIT: ok, I tested it out, yup - output lookup throws an exception, so I will update the docs to make it more obvious we are referring to environment files only.

PR #637, opened 2018-08-03
allow configs without stacks

troyready:
This relaxes the current stack validation, replacing it instead with
action warning messages.

With this, configurations can be created with only hooks.

### DIFF
--- a/stacker/actions/build.py
+++ b/stacker/actions/build.py
@@ -413,6 +413,8 @@ class Action(BaseAction):
 
         """
         plan = self._generate_plan(tail=tail)
+        if not plan.keys():
+            logger.warn('WARNING: No stacks detected (error in config?)')
         if not outline and not dump:
             plan.outline(logging.DEBUG)
             logger.debug("Launching stacks: %s", ", ".join(plan.keys()))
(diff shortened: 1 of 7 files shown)


troyready:
An alternative to this is just requiring a dummy stack; not sure if that's better to keep the Model more 'pure':

```stacks:
  placeholderstack:
    enabled: false
    class_path: nonexistentmodule.fakeclass```

phobologic:
I think this should be fine - I think once we have the hooks as part of the graph, we likely won't need this :)

phobologic:
@troyready can you merge master into this branch and see if it helps with the functional tests?

troyready:
Sure thing; rebased and pushed.

phobologic:
Ahh - I think I see what it is. The tests that used to stop because there were no stacks are now continuing on, and make calls to cloudformation.  Can you mark those as requiring AWS now?  There's a function for that that we use in the other tests that normally require AWS.

phobologic:
Ok, merged with #640 (cleaned up the tests there, since I don't have push permissions to your repo).  Thanks @troyready

PR #638, opened 2018-08-04
allow raw cfn templates to be loaded from remote package_sources

### LINKED ISSUE #628: Raw CFN templates cannot be used in remote packages
"Remote" blueprints can be used because downloaded packages are added to the python syspath. Raw template cannot be accessed in this way.

troyready:
Fixes #628 

This will be backwards compatible with any existing deployments, and should provide the expected load order for any more advanced use. Also gives a nicer error message now on filename typos.


phobologic:
This looks good - can you provide some tests @troyready ?

PR #648, opened 2018-08-27
Creates a memoized provider builder for AWS.

russellballestrini:
This could speed up stacker by many seconds based on the profiling work
@ejholmes performed in https://github.com/cloudtools/stacker/issues/630

	modified:   stacker/providers/aws/default.py

### DIFF
--- a/stacker/providers/aws/default.py
+++ b/stacker/providers/aws/default.py
@@ -12,6 +12,9 @@ import time
 import urllib.parse
 import sys
 
+# thread safe, memoized, provider builder.
+from threading import Lock
+
 import botocore.exceptions
 from botocore.config import Config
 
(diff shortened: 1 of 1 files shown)


stacker/providers/aws/default.py
```diff
@@ -448,17 +453,29 @@ def generate_stack_policy_args(stack_policy=None):
...
-        session = get_session(region=region, profile=profile)
-        return Provider(session, region=region, **self.kwargs)
+
+        # a thread safe and memoized, provider builder.
+        with self.lock:
+            # memoization lookup key derived by taking md5 of region + profile.
+            key = md5("%s%s" % (region, profile)).hexdigest()
+            
+            # create a new provider on key miss.
+            if key not in self.providers:
```
phobologic: I'd change this - since we'll likely find the key more often than not, it would probably be better as a try/except block:


```
try:
    return self.providers[key]
except KeyError:
    ... build new provider and return it ...
```

Also, self.providers isn't instantiated anywhere that I can tell before this - probably need a `self.providers = {}` in the `__init__` method.

stacker/providers/aws/default.py
```diff
@@ -448,17 +452,34 @@ def generate_stack_policy_args(stack_policy=None):
...
     def build(self, region=None, profile=None):
-        if not region:
-            region = self.region
-        session = get_session(region=region, profile=profile)
-        return Provider(session, region=region, **self.kwargs)
+        """Get or create the provider for the given region and profile."""
+
+        with self.lock:
+            # memoization lookup key derived by taking md5 of region + profile.
+            key = md5("{}{}".format(region, profile)).hexdigest()
```
ejholmes: I think you can just use the [hash built-in](https://docs.python.org/2/library/functions.html#hash) to generate a consistent hash for the dict key:

```python
>>> hash("{}{}".format(region, profile))
-2792180382875253710
```

stacker/providers/aws/default.py
```diff
@@ -448,17 +452,34 @@ def generate_stack_policy_args(stack_policy=None):
...
+            try:
+                # assume provider is in provider dictionary.
+                return self.providers[key]
+            except KeyError:
```
ejholmes: Using an exception for this seems less readable, and introduces two return paths instead of one. Can we just do:

```python
if key not in self.providers:
  self.providers[key] = ...
return self.providers[key]
```

russellballestrini: This is what I originally had but @phobologic suggested using the try/except block because it would be optimized for the typical case.

stacker/providers/aws/default.py
russellballestrini: fixed, couldn't we also just create a string key, like `"{}-{}".format(region, profile)`?

ejholmes: True, just a string key would be fine as well.

PR #655, opened 2018-08-31
Version dependencies format

zipkid:
Specify version dependencies as documented in
https://docs.python.org/3.6/distutils/setupscript.html#relationships-between-distributions-and-packages.
Using the '~=' format causes OS packages built with fpm to have a wrong
dependency.

(note the '=2.0' instead of '~=2.0')
```The following packages have unmet dependencies:
 python3-stacker : Depends: python3-gitpython (= 2.0) but 2.1.11-15 is 
to be installed
                   Depends: python3-schematics (= 2.0.1) but 2.0.1-15 is 
to be installed
                   Depends: python3-dateutil (= 2.0) but 2.7.3-15 is to 
be installed```

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -16,10 +16,10 @@ install_requires = [
     ##
     "PyYAML>=3.12",
     "awacs>=0.6.0",
-    "gitpython~=2.0",
-    "schematics~=2.0.1",
+    "gitpython>=2.0,<3.0",
+    "schematics>=2.0.1,<2.1.0",
     "formic2",
-    "python-dateutil~=2.0",
+    "python-dateutil>=2.0,<3.0",
 ]
 
 tests_require = [

### CODE REVIEW

**Comment by zipkid** [comment]
Hi,

Something seems to be wrong in the test environment. 
The tests seem to fail with

**Comment by phobologic** [comment]
@zipkid how are you running tests? They seem to pass in circleci.  Otherwise, this looks fine to me.

File: setup.py [hunk 1]
```diff
@@ -16,10 +16,10 @@
     ##
     "PyYAML>=3.12",
     "awacs>=0.6.0",
-    "gitpython~=2.0",
-    "schematics~=2.0.1",
+    "gitpython>=2.0,<=3.0",
```
phobologic: I believe you should be using `<` not `<=` - basically `~=2.0`  meants it has to be at least version 2.0, but can't move into 3.0.  Can you update, then I'll merge?  Thanks.

zipkid: Hi,

I have updated the PR.

The first time i pushed the CircleCi tests all failed with hat error. After i rebased on master and pushed again it passed. It may have been a temporary issue :-)

PR #657, opened 2018-09-03
Fixes issue with non-interactive changeset updates & stack policies

### LINKED ISSUE #650: Update of a Raw Blueprint with Transform fails
Hi,

I believe I have discovered a bug in Stacker. The bug seems to be triggered by trying to update a raw template which uses Transforms. I'm using master at the time of this report (d1353b3a942731ada1532f0e1422d64673988c1a) with python 3.6.5 on macOS 10.13.6. Stacker is installed in a virtualenv.

Config to reproduce:
config.yaml:
```yaml
---
namespace: 'whatever'
stacks:
    - name: 'dummy'
      template_path: dummy.json
```
dummy.json:
```json
{
    "Transform": "AWS::Serverless-2016-10-31",
    "Resources": {
        "DummyParameter": {
            "Type": "AWS::SSM::Parameter",
            "Properties": {
                "Type": "String",
                "Value": "foobar"
            }
        }
    }
}
```

Steps to reproduce: run `build` twice:
```
% stacker build config.yml
[2018-08-28T15:56:44] Using default AWS provider mode
[2018-08-28T15:56:50] dummy: submitted (creating new stack)
[2018-08-28T15:57:20] dummy: complete (creating new stack)
```
```
% stacker build config.yml
[2018-08-28T16:00:56] Using default AWS provider mode
[2018-08-28T16:00:57] create_change_set() got an unexpected keyword argument 'stack_policy'
  File "<stripped>/stacker/venv/lib/python3.6/site-packages/stacker/plan.py", line 93, in _run_once
    status = self.fn(self.stack, status=self.status)
  File "<stripped>/stacker/venv/lib/python3.6/site-packages/stacker/actions/build.py", line 353, in _launch_stack
    stack_policy=stack_policy,
  File "<stripped>/stacker/venv/lib/python3.6/site-packages/stacker/providers/aws/default.py", line 805, in update_stack
    stack_policy=stack_policy, **kwargs)
  File "<stripped>/stacker/venv/lib/python3.6/site-packages/stacker/providers/aws/default.py", line 891, in noninteractive_changeset_update
    'UPDATE', service_role=self.service_role, **kwargs
TypeError: create_change_set() got an unexpected keyword argument 'stack_policy'
[2018-08-28T16:00:57] dummy: failed (create_change_set() got an unexpected keyword argument 'stack_policy')
[2018-08-28T16:00:57] The following steps failed: dummy
```

phobologic:
create change set doesn't allow a `stack_policy` argument like
create/update stack calls.  This was handled in the interactive version,
but not in the non-interactive version.

Fixes #650

### DIFF
--- a/stacker/providers/aws/default.py
+++ b/stacker/providers/aws/default.py
@@ -397,6 +397,8 @@ def generate_cloudformation_args(stack_name, parameters, tags, template,
             with create_change_set.
         service_role (str, optional): An optional service role to use when
             interacting with Cloudformation.
+        stack_policy (:class:`stacker.providers.base.Template`): A template
+            object representing a stack policy.
         change_set_name (str, optional): An optional change set name to use
             with create_change_set.
 
@@ -434,6 +436,16 @@ def generate_cloudformation_args(stack_name, parameters, tags, template,
 
 
 def generate_stack_policy_args(stack_policy=None):
+    """ Converts a stack policy object into keyword args.
+
+    Args:
+        stack_policy (:class:`stacker.providers.base.Template`): A template
+            object representing a stack policy.
+
+    Returns:
+        dict: A dictionary of keyword arguments to be used elsewhere.
+    """
+
     args = {}
     if stack_policy:
         logger.debug("Stack has a stack policy")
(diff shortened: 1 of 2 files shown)


stacker/providers/aws/default.py
```diff
@@ -914,6 +947,8 @@ def noninteractive_changeset_update(self, fqn, template, old_parameters,
             'UPDATE', service_role=self.service_role, **kwargs
         )
 
+        self.deal_with_changeset_stack_policy(fqn, stack_policy)
```
russellballestrini: Ok cool, so this is the bug, you refactored the logic so you could call it here.

phobologic:
https://github.com/cloudtools/stacker/blob/master/stacker/tests/providers/aws/test_default.py#L416-L418

looks like the usecase

nielslaukens:
@ejholmes ChangeSets are required when a Transform is present, both on create & on update.

PR #658, opened 2018-09-04
fix interactive_update_stack calls with empty string parameters

troyready:
PR #644 didn't account for empty string parameters


troyready:
Might make sense to break out the old/new params -> dict conversions to a separate method so they can be have unit tests made?

phobologic:
Yeah, definitely seems like at least the new parameter dict generator might be better as a function - it's fairly complicated for a one liner.

PR #659, opened 2018-09-06
Fix KMS lookup for Python2

russellballestrini:
Explicitly get an UTF-8 bytestring but don't try to decode anything afterwards.

Resolves this error:

```
  'ascii' codec can't decode byte 0xcb in position 5: ordinal not in range(128))
```

	modified:   stacker/lookups/handlers/kms.py
	modified:   stacker/tests/lookups/handlers/test_kms.py


russellballestrini:
Verified to fix the issue found at @ Remind

russellballestrini:
I suppose it isn't necessary to be explicit but that's how the this bug crept in.

The trailing `.decode` was the actual issue.

Issue #515, opened 2017-11-29
Tail (-t) breaks if aws spend more than 5 seconds to start the stack.


Evnsan:
Hey folks, I am facing this problem when trying to deploy a stack with asg and codedeploy. Do you think it's a big deal?

```
[2017-11-29T15:10:55] Tailing stack: alb-test-1-app
Process Process-2:
Traceback (most recent call last):
  File "/usr/lib/python2.7/multiprocessing/process.py", line 258, in _bootstrap
    self.run()
  File "/usr/lib/python2.7/multiprocessing/process.py", line 114, in run
    self._target(*self._args, **self._kwargs)
  File "/home/evnsan/git/cobli/deploy/cloudformation/cloudform-venv/local/lib/python2.7/site-packages/stacker/providers/aws/default.py", line 541, in tail_stack
    self.tail_stack(stack, retries=retries + 1, **kwargs)
  File "/home/evnsan/git/cobli/deploy/cloudformation/cloudform-venv/local/lib/python2.7/site-packages/stacker/providers/aws/default.py", line 541, in tail_stack
    self.tail_stack(stack, retries=retries + 1, **kwargs)
  File "/home/evnsan/git/cobli/deploy/cloudformation/cloudform-venv/local/lib/python2.7/site-packages/stacker/providers/aws/default.py", line 541, in tail_stack
    self.tail_stack(stack, retries=retries + 1, **kwargs)
  File "/home/evnsan/git/cobli/deploy/cloudformation/cloudform-venv/local/lib/python2.7/site-packages/stacker/providers/aws/default.py", line 541, in tail_stack
    self.tail_stack(stack, retries=retries + 1, **kwargs)
  File "/home/evnsan/git/cobli/deploy/cloudformation/cloudform-venv/local/lib/python2.7/site-packages/stacker/providers/aws/default.py", line 541, in tail_stack
    self.tail_stack(stack, retries=retries + 1, **kwargs)
  File "/home/evnsan/git/cobli/deploy/cloudformation/cloudform-venv/local/lib/python2.7/site-packages/stacker/providers/aws/default.py", line 535, in tail_stack
    include_initial=False)
  File "/home/evnsan/git/cobli/deploy/cloudformation/cloudform-venv/local/lib/python2.7/site-packages/stacker/providers/aws/default.py", line 577, in tail
    initial_events = self.get_events(stack_name)
[...]

PR #663, opened 2018-09-26
Get rid of recursion for tail retries

phobologic:
Also, extend the way we retry/timeout, that should work around #515.

Not sure of a great way to test this unfortunately.


danielkza:
Looks good to me :+1:

PR #674, opened 2018-10-24
Improve Resilency on Approval Options when making changes to stack

blade2005:
By making it all lowercase the usable options become Y/y, v/V and anything else skips.

This addresses perhaps a small issue but one which may be confusing if you've accidentally turned on caps lock while coding.

I'm also a big believer that in cases like this there shouldn't be a difference between uppercase and lowercase options.


phobologic:
Thanks for this - going to go ahead and merge, the lint issues are due to a flake8 upgrade that we fixed in another PR.  I'll resolve any issues after merge.

PR #678, opened 2018-11-01
Add `in_progress` option to stack config.

ejholmes:
When stacker attempts to update a stack that's already in `UPDATE_IN_PROGRESS`, the existing behavior is to throw an exception:

```
[2018-11-01T22:19:50] acme-inc: failed (Stack: "acme-inc" cannot be updated nor re-created from state UPDATE_IN_PROGRESS: Update already in-progress)
```

In a CI/CD environment where lots of concurrent changes are happening, this can be problematic.

This PR adds a new per-stack `in_progress` option, which allows you to configure the behavior of stacker when the stack is `UPDATE_IN_PROGRESS` or `CREATE_IN_PROGRESS`. The default is the current behavior (error out), but can be set to `wait`, which will cause stacker to wait for the stack to be updateable before proceeding.

### DIFF
--- a/stacker/status.py
+++ b/stacker/status.py
@@ -77,6 +77,7 @@ class StackDoesNotExist(SkippedStatus):
 
 
 PENDING = PendingStatus()
+WAITING = PendingStatus(reason="waiting")
 SUBMITTED = SubmittedStatus()
 COMPLETE = CompleteStatus()
 SKIPPED = SkippedStatus()
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -404,6 +404,12 @@ A stack has the following keys:
   that will be applied when the CloudFormation stack is created and updated.
   You can use stack policies to prevent CloudFormation from making updates to
   protected resources (e.g. databases). See: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/protect-stack-resources.html
+**in_progress**:
+  (optional): If provided, specifies the behavior for when a stack is in
+  `CREATE_IN_PROGRESS` or `UPDATE_IN_PROGRESS`. By default, stacker will raise
+  an exception if the stack is in an `IN_PROGRESS` state. You can set this
+  option to `wait` and stacker will wait for the previous update to complete
+  before attempting to update the stack.
 
 Stacks Example
 ~~~~~~~~~~~~~~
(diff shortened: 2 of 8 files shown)


phobologic:
Some questions:

- anything we should worry about with rollback states or anything like that?
- do we only want it to be per-stack (and not globally)?

russellballestrini:
`concurency_handler: error` default

`concurency_handler: wait`

ejholmes:
> anything we should worry about with rollback states or anything like that?

Probably. I just kept it simple for this version, but in reality it should probably not `return True` immediately after waiting, and instead continue down with the new stack status so that rollbacks get handled.

> do we only want it to be per-stack (and not globally)?

I could go either way. I kinda lean towards per-stack just cause that's more flexible, but we could do both (per-stack overrides global default).

---

Re: the naming, I'm also not a huge fan. It's purpose is to configure the desired behavior when a stack is in an "in progress" state, or otherwise not updatable (`CREATE_IN_PROGRESS`, `UPDATE_IN_PROGRESS`, `DELETE_IN_PROGRESS`, `UPDATE_COMPLETE_CLEANUP_IN_PROGRESS`). All of these have `IN_PROGRESS` in the name, which is why I went with `in_progress`.

Just brainstorming:

```yaml
in_progress: wait
in_progress_behavior: wait
operation_in_progress: wait
not_ready: wait
```

`concurrency_handler` makes me think threading, and not necessarily accurate (it may just be that someone did something manually, and not necessarily a result of concurrent `stacker builds`).

ejholmes:
I've changed the implementation slightly so that it's all implemented within `_launch_stack`, instead of in the aws provider. This should be better because:

1. SIGTERM/SIGINT works properly
2. It let's us update the status of the stack to indicate that stacker is waiting

[asciicast](https://asciinema.org/a/FNX11aV06KBfNnGDyg9cqIsWt)

russellballestrini:
I vote for `in_progress_behavior: `

ejholmes:
Cool, I like `in_progress_behavior` too. Updated to that.

phobologic:
Yeah, `in_progress_behavior` is probably the most descriptive/easy to understand.  It's long, and that's annoying, but I think it's the best.

stacker/status.py
```diff
@@ -77,6 +77,7 @@ class StackDoesNotExist(SkippedStatus):
 
 
 PENDING = PendingStatus()
+WAITING = PendingStatus(reason="waiting")
```
phobologic: It might be worth something like `waiting on previous update` for the reason, just to be more explicit.

russellballestrini:
@ejholmes could we get this PR rebased and merged in? We had people in the chatroom asking for it.

PR #679, opened 2018-11-02
Fix lint warnings

ejholmes:
Newer versions of flake8 have added some new warnings (W503, W504 and W605). They're kinda ridiculous, so I've added them to the list of ignores.


ejholmes:
Python 3.6 test failures are probably unrelated.

PR #682, opened 2018-11-09
Use default ACL for uploaded lambda code

zollman:
The "Authenticated-Read" ACL, currently set on all uploads, allows your code
to be read by all S3 users. Default behavior should be to use the permissions
implied by the bucket policy, i.e. "private".

Organizations that do not grant SetObjectAcl permissions (for fear of
data loss) will block this call.

### DIFF
--- a/stacker/hooks/aws_lambda.py
+++ b/stacker/hooks/aws_lambda.py
@@ -228,8 +228,7 @@ def _upload_code(s3_conn, bucket, prefix, name, contents, content_hash):
     else:
         logger.info('lambda: uploading object %s', key)
         s3_conn.put_object(Bucket=bucket, Key=key, Body=contents,
-                           ContentType='application/zip',
-                           ACL='authenticated-read')
+                           ContentType='application/zip')
 
     return Code(S3Bucket=bucket, S3Key=key)
 
--- a/stacker/hooks/aws_lambda.py
+++ b/stacker/hooks/aws_lambda.py
@@ -194,7 +194,8 @@ def _head_object(s3_conn, bucket, key):
             raise
 
 
-def _upload_code(s3_conn, bucket, prefix, name, contents, content_hash):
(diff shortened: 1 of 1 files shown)


zollman:
... getting flake8: error: no such option: --require-code
make: *** [lint] Error 2 

Running python 3.7, latest macOS

troyready:
Seems great to me.

A strict semver interpretation might call this a 2.0 change. Alternatively, maybe no one ever interacted with the zipfiles outside of stacker and this could be in 1.x?

zollman:
Given where this lives deep in the code, I could alternately make this a config or environment option & not change the default. My problem is that right now I have to monkeypatch to get the code to work, since I don't have SetObjectAcl permissions and therefore I can't use this code at all.

phobologic:
This seems like a good change to me, though we probably should make it configurable (and default to the more secure option) just in case people do use this for some reason outside of stacker.  Can you make those changes, @zollman ?

zollman:
Will take me a few days, but sounds good @phobologic .

zollman:
Added a config option. 
And just in time! AWS' new Block Public S3 buckets feature prevents authenticated-read also, leading to Access Denied errors:
https://aws.amazon.com/blogs/aws/amazon-s3-block-public-access-another-layer-of-protection-for-your-accounts-and-buckets/

 Caveats:

* If you change payload_acl from one run to a second, it does not re-upload the file and fix the ACL - it skips because the file already exists
* I deliberately did not write docs or checks for the set of canned ACLs, because AWS may change them. If you set an invalid payload_acl, you get an error on upload.

Let me know if this works.

Thanks for the guidance @phobologic & team - if you're at re:invent this week, I would love to say thanks in person.

phobologic:
This is great, thanks @zollman - sorry for the delay in taking another look, and sorry for missing you at re:Invent - this is the first year I didn't go though!  I'm working on fixing the circleci issues with moto/yaml, and once that's merged I'll be merging this :)

PR #684, opened 2018-11-19
Pinning PyYAML to 3.13 to deal with cfn-flip pin

phobologic:
https://github.com/awslabs/aws-cfn-template-flip/pull/54

YAML was pinned in the cfn-flip package that troposphere depends on, and
without this we have issues with building.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -8,10 +8,10 @@ lint:
 	flake8 --require-code --min-version=2.7 --ignore FI50,FI51,FI53,FI14,E402,N802,W605 stacker/tests # ignore setUp naming
 
 test-unit: clean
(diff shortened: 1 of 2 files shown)


troyready:
Agreed on adding a comment about it

phobologic:
Looks like the pinning isn't required anymore - this removes it now, but I've gone back to pinning boto/botocore to deal with the current issues in moto.

    https://github.com/spulec/moto/issues/1924
    https://github.com/spulec/moto/issues/1941

@xiaket pinging you since you had removed the pinning of boto/botocore, hopefully this doesn't lead to too many issues.

xiaket:
Thanks @phobologic ! Yeah, this pinning gave us some little trouble a while back but it's all sorted out by now. Cheers.

PR #696, opened 2019-02-04
Add Additional ECS Unit Test

alecrajeev:
added another unit test for code coverage


alecrajeev:
Unit test will verify that trying to create a cluster without a name throws the correct error

PR #700, opened 2019-02-08
Add Keypair Unit Test

alecrajeev:
Uses [moto](https://github.com/spulec/moto) to mock creating and describing the keypair API for unit testing. This increases coverage for hooks/keypair

### DIFF
--- a/stacker/tests/hooks/test_keypair.py
+++ b/stacker/tests/hooks/test_keypair.py
@@ -1,0 +1,162 @@
+from __future__ import print_function
+from __future__ import division
+from __future__ import absolute_import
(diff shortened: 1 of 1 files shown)


danielkza:
This test has hardcoded paths to CircleCI that break running the tests locally: #709 

Having written the hook and now looking at this, I strongly recommend to:

a) Not test the logging messages. That makes the test very fragile and adds little value against regressions.
b) Create a temporary directory and use it as the "home base" for any test that requires simulating/inspect file manipulation. It's the most comprehensive way to observe behavior close to reality, and does not break in different environments.

PR #701, opened 2019-02-11
add jinja2 template parsing

troyready:
Decided to make it completely opt-in by requiring templates to be explicitly named `*.j2`, so this won't affect anyone that doesn't explicitly chose to use it.

### DIFF
--- a/stacker/blueprints/raw.py
+++ b/stacker/blueprints/raw.py
@@ -8,6 +8,8 @@ import json
 import os
 import sys
 
+from jinja2 import Template
+
 from ..util import parse_cloudformation_template
 from ..exceptions import InvalidConfig, UnresolvedVariable
 from .base import Blueprint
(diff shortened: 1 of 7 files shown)


xiaket:
Great feature! This is so cool!

stacker/blueprints/raw.py
```diff
@@ -186,7 +188,22 @@ def rendered(self):
...
+                    if len(os.path.splitext(template_path)) == 2 and (
+                            os.path.splitext(template_path)[1] == '.j2'):
+                        self._rendered = Template(template.read()).render(
+                            context=self.context,
+                            mappings=self.mappings,
+                            name=self.name,
+                            # self.resolved_variables are not available during
+                            # template rendering; config variables are
+                            # provided directly instead
+                            variables=list(
```
phobologic: Hmm - does this mean we can't use lookups effectively here?

stacker/tests/factories.py
```diff
@@ -32,10 +32,13 @@ def mock_context(namespace="default", extra_config_args=None, **kwargs):
     if extra_config_args:
         config_args.update(extra_config_args)
     config = Config(config_args)
-    environment = kwargs.get("environment", {})
+    if kwargs.get("environment"):
```
phobologic: Curious what htis change is about?  Doesn't seem to change much functionally?

troyready: Oh yeah, this was interesting.

The problem with the current code is that you can't include `environment` in the kwargs, because it's being explicitly provided already. Without the change, it errors like:
```
  File "/home/troyready/python/stacker/stacker/tests/blueprints/test_raw.py", line 156, in test_j2_to_json
    environment={'foo': 'bar'}),
  File "/home/troyready/python/stacker/stacker/tests/factories.py", line 39, in mock_context
    **kwargs)
TypeError: type object got multiple values for keyword argument 'environment'
```

stacker/blueprints/raw.py
troyready: Yeah, this was too much of caveat to endure. Fixed now.

troyready:
@phobologic I dug into the variable resolution timing and fixed it. Should work as expected now.

stacker/tests/factories.py
phobologic: Oooh, interesting.  Thanks for the explanation.

phobologic:
Thanks @troyready !

russellballestrini:
Nice job @troyready!

PR #705, opened 2019-03-04
Allow for custom log formats to be defined in stacker.yaml

russellballestrini:
At [Remind](https://www.remind.com) we run stacker on a large portion of our stacks in stage and production concurrently using continuous delivery. This means our logs are interlaced between stage and prod and are hard to use.

Also since we use  continuous delivery we want to put the `deployment_id` or `git commit hash` on every log line so that we can aggregate them all up with our log parsing and viewing tools.

This PR makes it possible to set a custom log format from `stacker.yaml`.

	modified:   .gitignore
	modified:   stacker/commands/stacker/__init__.py
	modified:   stacker/commands/stacker/base.py
	modified:   stacker/config/__init__.py
	modified:   stacker/logger/__init__.py
	new file:   stacker/tests/fixtures/not-basic.env
	new file:   stacker/tests/fixtures/vpc-custom-log-format-info.yaml
	modified:   stacker/tests/test_stacker.py

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -68,3 +68,5 @@ dev.yaml
 dev.env
 
 tests/fixtures/blueprints/*-result
+
+FakeKey.pem
--- a/docs/config.rst
+++ b/docs/config.rst
@@ -469,6 +469,28 @@ Here's an example of a target that will execute all "database" stacks::
       required_by:
         - databases
 
+Custom Log Formats
+------------------
+
+By default, stacker uses the following `log_formats`::
+
+  log_formats:
+    info: "[%(asctime)s] %(message)s"
+    color: "[%(asctime)s] \033[%(color)sm%(message)s\033[39m"
+    debug: "[%(asctime)s] %(levelname)s %(threadName)s %(name)s:%(lineno)d(%(funcName)s): %(message)s"
+
+You may optionally provide custom `log_formats`.
+
+You may use any of the standard Python
+[logging module format attributes](https://docs.python.org/2.7/library/logging.html#logrecord-attributes)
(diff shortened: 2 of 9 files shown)


stacker/tests/fixtures/vpc-custom-log-format-info.yaml
```diff
@@ -0,0 +1,24 @@
+log_format_info: "[%(asctime)s] ${environment} custom log format - %(message)s"
```
ejholmes: In `stacker.logger.setup_logging` we also support `color` and `debug` keys. Should we just make this a nested structure.

**Example**

```yaml
log_format:
  info: "[%(asctime)s] ${environment} custom log format - %(message)s"
  color: # Some color format
  debug: # Some debug format
```

.gitignore
```diff
@@ -68,3 +68,5 @@ dev.yaml
 dev.env
 
 tests/fixtures/blueprints/*-result
+
+FakeKey.pem
```
phobologic: What's this about?

docs/config.rst
```diff
@@ -469,6 +469,24 @@ Here's an example of a target that will execute all "database" stacks::
       required_by:
         - databases
 
+
+Custom Log Formats
+------------------
+
+You may optionally provide a custom log format for stacker to use::
+
+  # optional log format overrides.
```
phobologic: Might be worth putting the defaults here, or at least explaining how we use the python logging module standard format variables (link to here? https://docs.python.org/2.7/library/logging.html#logrecord-attributes)

PR #710, opened 2019-03-12
add cf notification arns parameter

kukhmaz:
Added new parameter to add sns notifications to cloudformation stacks.

### DIFF
--- a/stacker/providers/aws/default.py
+++ b/stacker/providers/aws/default.py
@@ -698,7 +698,7 @@ class Provider(BaseProvider):
 
     def create_stack(self, fqn, template, parameters, tags,
                      force_change_set=False, stack_policy=None,
-                     notification_arns=[],**kwargs):
+                     notification_arns=[], **kwargs):
         """Create a new Cloudformation stack.
 
         Args:
@@ -987,7 +987,8 @@ class Provider(BaseProvider):
         )
 
     def default_update_stack(self, fqn, template, old_parameters, parameters,
-                             tags, stack_policy=None, notification_arns=[], **kwargs):
+                             tags, stack_policy=None,
+                             notification_arns=[], **kwargs):
         """Update a Cloudformation stack in default mode.
 
         Args:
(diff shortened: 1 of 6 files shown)


stacker/providers/aws/default.py
```diff
@@ -301,15 +301,16 @@ def wait_till_change_set_complete(cfn_client, change_set_id, try_count=25,
 
 def create_change_set(cfn_client, fqn, template, parameters, tags,
                       change_set_type='UPDATE', replacements_only=False,
-                      service_role=None):
+                      service_role=None, notification_arns=[]):
```
phobologic: You shouldn't use a mutable type as an argument to a method because that is instantiated when that method is parsed, not when it's executed.  Here's an example of why not:

```
>>> def blah(a, x=[]):
...     x.append(a)
...     return x
...
>>>
>>> x = blah("apple")
>>> y = blah("banana")
>>>
>>> print x
['apple', 'banana']
>>> print y
['apple', 'banana']
```

Instead go ahead and make the default `None`, and then have something like this inside the function:

```
notification_arns = notification_arns or []
```

phobologic: And actually - I'd just leave it as None here, and then handle the converting it to a list at the `generate_cloudformation_args` method, that way you can just use None everywhere that calls that method.

stacker/providers/aws/default.py
```diff
@@ -418,6 +420,7 @@ def generate_cloudformation_args(stack_name, parameters, tags, template,
         "Parameters": parameters,
         "Tags": tags,
         "Capabilities": capabilities,
+        "NotificationARNs": notification_arns
```
phobologic: Since this could be blank, you might want to follow the pattern below for optional arguments.  Not sure how boto handles a blank list for the argument here.

phobologic:
Ping @kukhmaz

phobologic:
Closing since I haven't heard back in a while - please re-open if you have a chance to look at it again!

russellballestrini:
I'm going to take this over.

russellballestrini:
Closing in favor of https://github.com/cloudtools/stacker/pull/756

PR #715, opened 2019-03-14
 hooks: keypair: add some features and rewrite tests 

danielkza:
- Add support for importing a local public key file
- Add support for storing generated private keys in SSM parameter store
- Refactor code to be more streamlined and separate interactive input
from other work

This is currently based on #714 since the tests make heavy use of pytest fixtures.


stacker/hooks/keypair.py
```diff
@@ -2,21 +2,183 @@
...
+    }
+
+
+def interactive_prompt(keypair_name, ):
+    if not sys.stdin.isatty():
+        return None, None
+
+    try:
+        while True:
+            action = input(
```
phobologic: worth using this input method since we use it most other places? https://github.com/cloudtools/stacker/blob/1cb24aa1a596261d533a5dfc4536757a3743b6ab/stacker/ui.py#L14

danielkza: Done!

PR #716, opened 2019-03-14
commands: info: add configurable output formats

danielkza:


phobologic:
Any idea what is up with the failing tests?

PR #727, opened 2019-04-15
Update index.rst

murribu:
Change `.html` to `.rst`


murribu:
Oops. Never mind. I see that this is being implemented on the website; and it works there. It only breaks if you're navigating on github.

PR #732, opened 2019-06-04
test diff before the stack exists

kddejong:
Currently a diff will fail if the stack doesn't exist.  Most of the logic is there but it doesn't fully cover it. 

I'm still not sure how the outputs should be fully handled so not changing it much but only appending it if the stack currently exists.


kddejong:
The impact of the change isn't fully noted in the tests.  The first commit should fail and the second commit should pass.

PR #739, opened 2019-07-17
Doc Update: Sturdy -> Onica

fbattistella:
Sturdy joined onica a good while ago. Stumbled upon the doc page today and saw we never requested an update on it.


troyready:
Thanks @fbattistella ! It looks like the line `We are an AWS Premier Consulting & Managed Service Partner` is redundant to the rest of the messaging and can be removed?

fbattistella:
@troyready updated 👍

PR #740, opened 2019-07-24
Add YAML environment file support

mromaszewicz:
Environment files ending in .yaml or .yml will be rendered using
a different strategy. The env file is treated as a dict, with
the value for each key being a parsed YAML object. After parsing
the config, we look for things that look like variable references,
and replace the references with objects.

### DIFF
--- a/docs/environments.rst
+++ b/docs/environments.rst
@@ -3,7 +3,15 @@ Environments
 ============
 
 When running stacker, you can optionally provide an "environment" file. The
-stacker config file will be interpolated as a `string.Template
+environment file defines values, which can then be referred to by name from
+your stack config file. The environment file is interpreted as YAML if it
+ends in `.yaml` or `.yml`, otherwise it's interpreted as simple key/value
+pairs.
+
+Key/Value environments
+----------------------
+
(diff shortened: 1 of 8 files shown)


sd0408:
When this is merged will we be able to define list object and lists of objects in these yaml files for parsing later on?

docs/environments.rst
```diff
@@ -43,6 +51,58 @@ files in your config. For example::
...
+    - name: webservers
+      class_path: stacker_blueprints.asg.AutoscalingGroup
+      variables:
+        InstanceType: ${web_instance_type}
+        IngressCIDRsByPort: ${ingress_cidrs_by_port}
+
+We've defined a stack which expects a list of ingress CIDR's allowed access to
+each port. Our environment files would look like this::
+
+  # in the file: stage.env
```
sd0408: shouldn't this doc state stage.yml and not stage.env?

deepmap-marcinr: Yup, will update shortly!

deepmap-marcinr:
So, guys, what's the verdict on this PR?

mromaszewicz:
Any thoughts on merging this?

PR #746, opened 2019-11-13
Locked stacks still have requirements

### LINKED ISSUE #745: Regression: Locked stacks do not properly construct dependency list.
Consider this scenario:
Stack A has output Foo.
Stack B depends on A::Foo

This is expected to work:
stacks:
```
- name: A
   enabled: true
   class_path: <something>
- name: B
   enabled: true
   locked: true
   class_path: <something>
   variables:
     SomeInput: ${output A::Foo}
  ```

It is expected that stack A must be in the dependency DAG before stack B, but this doesn't happen as of stacker 1.5.0. It works properly in stacker 1.4.1

In stacker 1.5.0, dependencies are built correctly if stack B is not locked. However, at creation time, it should not matter whether a stack is locked or not.

This is easiest to reproduce by having a stacker file with 2 stacks in it, and swap their order if it works initially. It will fail in 1.5.0+

mromaszewicz:
This fixes #745 . Locked stacks still have dependencies, since there
is no difference between a locked stack and an unlocked stack at
creation time.

### DIFF
--- a/stacker/stack.py
+++ b/stacker/stack.py
@@ -85,12 +85,6 @@ class Stack(object):
 
     @property
     def requires(self):
-        # By definition, a locked stack has no dependencies, because we won't
-        # be performing an update operation on the stack. This means, resolving
-        # outputs from dependencies is unnecessary.
-        if self.locked and not self.force:
-            return []
-
         requires = set(self.definition.requires or [])
 
         # Add any dependencies based on output lookups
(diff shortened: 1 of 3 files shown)


mromaszewicz:
Hmm, the lint check failed because the CircleCI system has the wrong version of boto. How can I fix this?

phobologic:
I'll dig into the boto errors outside of this PR, thanks @mromaszewicz

PR #762, opened 2020-09-08
KMS lookups give values $VAL wrapped as "b'$VAL'" when using python 3

baryal1299:
Decrypted string is returned as b'text' which should be decoded to utf-8.


troyready (changes requested):
Appreciate the contribution here. I think the decoding needs to be gated on python 3, e.g. after `if sys.version_info[0] > 2:` condition

hauntingEcho:
This may have been superceded by #765

PR #768, opened 2021-02-03
Fix crash when diffing a stack that has unquoted date strings

aarongorka:
Same as https://github.com/cloudtools/stacker/pull/591


aarongorka:
Tested by creating a stack with the current version from pip, crashing stacker with a diff, applying the change and then running diff again to confirm it no longer crashes.

aarongorka:
The failed test step is running python 2.7, I have to ask: what's up with that?

russellballestrini (approved):
looks good to me. As for the test failures, they seem unrelated to this PR.

PR #772, opened 2021-11-19
Updated `set_template_description()`

scrthq:
Troposphere updated the underlying method name to be `set_description()` instead of `add_description()`. Without patching this behavior, setting a description on the top-level config file is unusable with the latest version of Troposphere, returning the following: `AttributeError: 'Template' object has no attribute 'add_description'`


kukhmaz:
Any update on this fix?

